PR #9, opened 2016-11-24
Include more region details

kring:
`/regions/search` now includes a bounding box (it already had the other properties):
image

And `/datasets/search` now includes both the name and bounding box:
image

If the region is not found, the name in the query part of the dataset search response will be `[Unknown]`, the `boundingBox` will be missing, and there will be an error just as before.

The region type and ID in the query are case insensitive, but the response will always have the correct case.

This currently does an extra Elasticsearch for each region included in the query, which is not ideal.  But since the UI will only ever include one region in the query, I think we can live with this for now.


src/main/scala/au/csiro/data61/magda/search/elasticsearch/ElasticSearchQueryer.scala
```diff
@@ -319,7 +323,19 @@ class ElasticSearchQueryer(implicit val system: ActorSystem, implicit val ec: Ex
...
-            case _ => Future(RegionSearchResult(query, response.totalHits, response.as[MatchingRegion].toList))
+            case _ => Future(RegionSearchResult(query, response.totalHits, response.as[Region].toList))
+          }
+        }
+    }
+  }
+
+  def findRegion(regionType: String, regionId: String): Future[Region] = {
+    clientFuture.flatMap { client =>
+      client.execute(ElasticDsl.search in "regions" / "regions" query { idsQuery((regionType + "/" + regionId).toLowerCase) } start 0 limit 1 sourceExclude ("geometry"))
```
AlexGilleran: Can we use ElasticDsl.get if we already know the id?

kring:
I fixed the merge conflicts.

PR #10, opened 2016-11-24
Gmdcsw

AlexGilleran:


src/main/resources/env-specific-config/local.conf
```diff
@@ -20,5 +20,5 @@ indexedServices {
 }
 
 indexer {
-	alwaysReindex = false
+	alwaysReindex = true
```
kring: Did you mean to leave this in?  Maybe #7 would be better?

kring:
👍 
You can change alwaysReindex in master if you want, or not.

PR #23, opened 2017-01-20
Rename section to aspect and change where snapshots are stored

kring:


kring:
@AlexGilleran I'm PRing this so you can confirm the new snapshot location is ok on your system.

PR #43, opened 2017-04-20
Make indexer accept webhooks from registry.

AlexGilleran:
@kring - I pulled some domain classes out of the registry project into common so I could use them as input for the webhooks, you might want to look over what I did :).


CkanConnector/package.json
```diff
@@ -4,9 +4,11 @@
   "description": "MAGDA CKAN Connector",
   "main": "index.js",
   "scripts": {
-    "build": "tsc",
+    "build": "export REGISTRY_URL=http://localhost:6100/;npm run generateSwaggerApis && tsc",
```
kring: nice, though this won't work on Windows because npm runs scripts under CMD even when it's invoked from bash.  That's why the `bash -c` stuff below.  But I'll fix it next time I run into it.

kring: Oh also we might not want to `npm run generateSwaggerApis` because it requires the registry already be running.  In master the generated code is checked in now, avoiding the problem you ran into.

common/src/main/scala/au/csiro/data61/magda/model/Registry.scala
```diff
@@ -0,0 +1,81 @@
...
+object Registry {
+  @ApiModel(description = "A type of aspect in the registry.")
+  case class AspectDefinition(
+    @(ApiModelProperty @field)(value = "The unique identifier for the aspect type.", required = true) id: String,
+
+    @(ApiModelProperty @field)(value = "The name of the aspect.", required = true) name: String,
+
+    @(ApiModelProperty @field)(value = "The JSON Schema of this aspect.", required = false, dataType = "object") jsonSchema: Option[JsObject])
+
+  case class WebHookPayload(
```
kring: As you may have gathered, I'm kind of fond of the one-file-per-type thing.  But is that crazy cause of Scala compilation times or something?

kring: ehh I guess this is one file per type, because it's all in a Registry object.  nevermind then, nothing to see here. :)

registry-api/src/main/scala/au/csiro/data61/magda/registry/Protocols.scala
```diff
@@ -1,17 +1,15 @@
...
-  implicit object EventTypeFormat extends RootJsonFormat[EventType] {
-    def write(e: EventType) = JsNumber(e.value)
-    def read(value: JsValue) = EventType.withValue(value.asInstanceOf[JsNumber].value.toInt)
-  }
+trait Protocols extends DefaultJsonProtocol with DiffsonProtocol with au.csiro.data61.magda.model.Temporal.Protocols with CommonRegistryProtocols {
 
   implicit val badRequestFormat = jsonFormat1(BadRequest.apply)
-  implicit val aspectFormat = jsonFormat3(AspectDefinition.apply)
-  implicit val recordFormat = jsonFormat3(Record.apply)
   implicit val recordSummaryFormat = jsonFormat3(RecordSummary.apply)
```
kring: Should we move some more of these into Common?  Even if they're not needed there yet, it seems a bit odd not to have `BadRequest` and `RecordSummary` given the others that have been moved there.

kring:
Registry changes look fine to me.  Not sure why Travis doesn't like the PR (but the push is ok). I forced it and I'll merge this if the green tick shows up.

kring:
Still no love from Travis. :(

PR #44, opened 2017-04-27
Moved registry typescript files and generation script into registry

AlexGilleran:
Now it can be referenced by multiple connectors.


registry-api/generate-typescript.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+CURRENT_DIR=$(pwd)
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+CODEGEN_JAR=~/.ivy2/cache/io.swagger/swagger-codegen-cli/jars/
```
kring: I had to build a custom version of swagger-codegen to fix some problems.  Are you sure that isn't necessary anymore?
https://github.com/TerriaJS/magda/blob/master/tools/swagger-codegen-cli.md

AlexGilleran: :grimacing: oh ok I'll put it back to the old one.

kring: Maybe they fixed it, I'll check.

kring: Looks like it's still needed.  The change just avoids a "no implicit any" error from the typescript compiler.  Here's the change:
https://github.com/swagger-api/swagger-codegen/compare/master...TerriaJS:magda

I'm not 100% sure, though, because `generate-typescript.sh` is just spewing errors on my system at the moment. ☹️

PR #52, opened 2017-05-09
Use Lerna to manage multi-package repo

kring:
Some fun stuff to try (`npm install -g lerna` first):

* `lerna bootstrap` - Installs all dependencies of all components, hoisting most of them up to the root `node_modules` directory so we don't need a million copies.
* `lerna publish --skip-npm --skip-git` - Interactively update the version of all packages (only if anything has changed).
* `lerna run docker-build-local` - Run the `docker-build-local` npm script in all packages that have it.

To add a new package, just create a directory with a package.json as normal.  Then add it as a dependency in other package.jsons, and run `lerna bootstrap` to link things up.

Even the Scala projects can pull in data-ish dependencies by adding them to package.json and then expecting to find them in `node_modules/whatever`.

I'm sure I broke all sorts of things, and no idea if it works on Mac, so let me know.


kring:
Also @AlexGilleran you'll notice I got rid of `npm run startMinikube`.  Instead, you can override the registry URL either by setting `REGISTRY_URL` (as before), or with:
`npm config set @magda/ckan-connector:registryUrl $(minikube ip):30010`

AlexGilleran:
Looks good but there's a conflict :).

PR #57, opened 2017-05-31
Broken link sleuther

steve9164:
An initial attempt at a sleuther. It works, but it's not excellent.


magda-typescript-common/src/Registry.ts
```diff
@@ -48,10 +50,31 @@ export default class Registry {
...
+    async forEachRecord(callback: (record: Record) => Promise<any>, aspect?: Array<string>, optionalAspect?: Array<string>, dereference?: boolean) {
+        console.log(aspect);
+        const operation = (pageToken: string) => () => this.recordsApi.getAll(aspect, optionalAspect, pageToken, undefined, undefined, dereference);
+        let records;
+        let pageToken = undefined;
+        do {
+            records = await retry(operation(pageToken), this.secondsBetweenRetries, this.maxRetries, (e, retriesLeft) => console.log(formatServiceError('Failed to GET records.', e, retriesLeft)))
+                .then(result => result.body)
+                .catch(createServiceError);
+            await Promise.all((<any>records).records.map(callback));
```
steve9164: Why do the typescript definitions expect the type `records: Array<RecordSummary>`?

AFAICT this is because of a problem with the generated typescript here: https://github.com/TerriaJS/magda/blob/60af342/magda-typescript-common/src/generated/registry/api.ts#L1124

magda-typescript-common/src/Registry.ts
```diff
@@ -48,10 +50,31 @@ export default class Registry {
             .catch(createServiceError);
     }
 
+    async forEachRecord(callback: (record: Record) => Promise<any>, aspect?: Array<string>, optionalAspect?: Array<string>, dereference?: boolean) {
```
steve9164: I'd have preferred to use a generator/iterator style approach, but it seems it can't be done since there's promises involved (see [for-await-of proposal here](https://github.com/tc39/proposal-async-iteration)).

magda-broken-link-sleuther/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@magda/broken-link-sleuther",
+  "description": "MAGDA Broken Link Sleuther",
+  "version": "0.0.1",
```
kring: Make the version match the one in the other components.

magda-broken-link-sleuther/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "ftp": "^0.3.10",
+    "lru-cache": "^4.0.2",
+    "request": "^2.79.0",
+    "ts-node": "^2.1.0",
+    "urijs": "^1.18.7"
+  },
+  "config": {
+    "registryUrl": "http://localhost:6100/v0",
+    "docker": {
+      "name": "data61/magda-project-open-data-connector",
```
kring: This needs to be updated.

magda-broken-link-sleuther/src/BrokenLinkSleuther.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+    async run(): Promise<BrokenLinkSleutherResult> {
+        let checkedDistributions = 0;
+        let brokenLinks = 0;
+        let tooManyRequests = 0;
+
+        await this.registry.putAspectDefinition(aspectDefinition);
+
+
+        const registryPage = AsyncPage.create<{records: Array<{id: string, aspects: {'dcat-distribution-strings': {downloadURL: string}}}>, nextPageToken: string}>((previous) => {
```
kring: Might be a little cleaner to define an interface (or even a `type` on the line before) for this rather than putting the type inline.

kring: although really you would want to use the type generated from swagger.json, except that it's missing.. haha

steve9164: Yeah. Type inference would be fine here, but since the generated code has the wrong type I had to explicitly type this, so I just stuck it inline.

kring: Right, makes sense!

PR #142, opened 2017-07-10
Improvement

chloeleichen:
1. show error message for featured dataset (https://github.com/TerriaJS/magda/issues/139). This happens because the featured dataset uses promises.all, and we need to make the datasets show error independently
2. Fix an error in record parsing

### DIFF
--- a/magda-web-client/src/config.js
+++ b/magda-web-client/src/config.js
@@ -4,7 +4,7 @@ import Format from "./SearchFacets/Format";
 import Region from "./SearchFacets/Region";
 import Temporal from "./SearchFacets/Temporal";
 
-const fallbackApiHost = "http://search.data.gov.au/";
+const fallbackApiHost = "http://magda-api-dev.terria.io/";
 // const apiHost = 'http://minikube.data.gov.au:30016/';
 // const apiHost = 'http://localhost:3016/';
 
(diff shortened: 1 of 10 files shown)


steve9164:
I added another commit fixing another couple of places (that I originally modified) that were wrong.

magda-web-client/src/config.js
```diff
@@ -39,7 +39,7 @@ export const config = {
   appTitle: 'Australian open data search',
   featuredDatasets: [
     "19432f89-dc3a-4ef3-b943-5326ef1dbecc",
-    "bdcf5b09-89bc-47ec-9281-6b8e9ee147aa"
+    "0ff854cc-a621-4929-89d3-f0370c3dd344"
```
kring: Did you mean to commit this?

magda-web-client/src/helpers/publisher.js
```diff
@@ -17,11 +26,16 @@ type PublisherRaw = {
 
 
 export function parsePublisher(publisherRaw: PublisherRaw) : Publisher{
+    let error = null;
+    if(publisherRaw && !publisherRaw.id){
+      error = publisherRaw.message || 'an error occuered';
```
kring: typo, occuered -> occurred

magda-web-client/src/config.js
chloeleichen: nope

kring:
looks good other than those two small things.

PR #148, opened 2017-07-17
table preview 

chloeleichen:
Table view is available for xml, csv, json, xsl and excel
pdf view available for pdf files
text view for txt files


kring:
I'm getting an error in this branch when I compile:
```
./src/helpers/xmlToTabular.js
Module not found: Can't resolve 'fast-xml-parser' in 'C:\Users\rin035\github\magda\magda-web-client\src\helpers'
```

chloeleichen:
@kring fixed can you have another look

kring:
@steve9164 I know we talked about you reviewing this, but I think I will actually, because I'm interested in it from a feature perspective mostly.  I'm not so worried about the code.  Sorry for the delay @chloeleichen I'll merge this really soon.

kring:
This is looking great!  A few things I noticed, of varying priority:

* The PDF preview jumps up to the top of the page each time you click next page.  Also do we need a way to zoom in or change the size or something?
* Line breaks are not honored in text files.  e.g. this:
image
turns into this:
image
* Maybe preview DOCX and PPTX too if it's not hard? (it might be hard, in which case don't worry about it)
* I'm not sure that previewing GeoJSON as a table makes much sense in general.  It would be better to preview it using a hierarchy.  Something like this? https://codebeautify.org/jsonviewer/cbaebbad or http://jsonviewer.stack.hu/
* Over a meg of JavaScript already!  We'll need a good strategy for this (i.e. only load the code for a preview when we go to use that previewer).

kring:
I'll write separate issues for the things above.

PR #149, opened 2017-07-18
Sleuther Framework + Linked Data Rating Sleuther

AlexGilleran:
New sleuther framework takes care of: 
- Registering its aspect(s)
- Crawling the registry on startup
- Registering its webhook
- Listening at that webhook's address for events.

Linked data sleuther:
- Takes in datasets
- Looks up all the dataset's distributions and crudely examines their licenses / formats to figure out a star rating for each
- Dataset rating = best dataset

Potentially controversial:
The LD sleuther also writes to a generic dataset quality aspect. This aspect is a map of other sleuther's view of the quality of the data, with that sleuther's aspect name as the key - each entry has a percentage of the quality and a weighting determining how much that sleuther's rating should be taken into account when determining an overall rating. This is updated by PATCHing its `dataset-linked-data-quality` field so that it only overwrites its own rating. This required me to change the PATCH to create a new record if one didn't already exist.

Ideally this means that anyone can write a sleuther that affects data quality as interpreted by the UI and search without having to write plugins for either of those.


kring:
Well reviewing this turned into a bit of a yak shave; I ended up reworking the typescript configuration, including a `create-all-tsconfigs.js` in `scripts` to generate the now _two_ tsconfig files we need per component (one for use in ts-node, the other for a normal compile) 🙄.   But the sleuther framework and linked data quality sleuther both look good!

PR #150, opened 2017-07-19
Made search use quality.

AlexGilleran:


lerna.json
```diff
@@ -30,7 +30,8 @@
   ],
   "version": "0.0.28-SNAPSHOT",
   "hoist": [
-    "tsmonad"
+    "tsmonad",
+    "@types/express"
```
kring: Are you sure this is necessary?  I can't think of why it would be.  😟

kring:
This looks great to me, once the tests are behaving.

lerna.json
AlexGilleran: oh no that was just to make it work before 😬

AlexGilleran:
Success!

PR #151, opened 2017-07-19
Database migration support

kring:
`magda-postgres` builds a docker image that layers on top of the the normal `postgres:9.6` image to add support for database migration using Flyway.  Any Flyway versioned SQL scripts in `/flyway/sql/<dbname>` are used at container startup to migrate the named database.  The database is also created automatically if necessary.

I'm not really sure `magda-postgres` belongs in magda, it could just be a separate repo since it's not a magda service at all.  But it was convenient to put it here for now.


kring:
The tests are passing now. :shipit:

PR #152, opened 2017-07-20
fixes #147 add go to source btn

### LINKED ISSUE #147: Add "Go to Source" link in accordion view
In non-accordion view would it be possible to put a "Go to Source" view that links to the same link as source link on data set page? Like 

image

chloeleichen:
I didn't make it green as Allan suggested. I think it's better to be a bit more subtle than the view source button

### DIFF
--- a/magda-web-client/src/Dataset/DatasetSummary.js
+++ b/magda-web-client/src/Dataset/DatasetSummary.js
@@ -76,7 +76,7 @@ export default class DatasetSummary extends Component {
                       <div>
                         <Link className='btn dataset-summary__btn-view-dataset'
                                    to={`/dataset/${encodeURIComponent(dataset.identifier)}`}>View dataset</Link>
-                        <a className='btn dataset-summary__btn-go-to-source' href={dataset.landingPage}>Go to Source</a>
+                        {dataset.landingPage && <a className='btn dataset-summary__btn-go-to-source' href={dataset.landingPage}>Go to Source</a>}
                       </div>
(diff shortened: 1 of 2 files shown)


magda-web-client/src/Dataset/DatasetSummary.js
```diff
@@ -73,8 +73,11 @@ export default class DatasetSummary extends Component {
                     </span>
                   </div>
                   {this.props.isExpanded && <div className='dataset-summary__middle clearfix'>
-                      <div><Link className='btn dataset-summary__btn-view-dataset'
-                                 to={`/dataset/${encodeURIComponent(dataset.identifier)}`}>View dataset</Link></div>
+                      <div>
+                        <Link className='btn dataset-summary__btn-view-dataset'
+                                   to={`/dataset/${encodeURIComponent(dataset.identifier)}`}>View dataset</Link>
+                        <a className='btn dataset-summary__btn-go-to-source' href={dataset.landingPage}>Go to Source</a>
```
kring: What if the dataset doesn't have a landingPage property?

chloeleichen: Fixed

kring: Are you sure?  I don't see it.

chloeleichen: Sorry looks like i didn't commit that file. Fixed now

PR #158, opened 2017-07-21
Map preview with MagdaCatalogItem

steve9164:
~~Not sure if the merge with master is good or not.~~ Reverted that merge (it was with an old version of master anyway)


magda-web-client/src/helpers/previewData.js
```diff
@@ -23,6 +23,14 @@ export function getPreviewDataUrl(distributions: Array<ParsedDistribution>){
     const html = viewableDistribution.filter(d=> d.format.toLowerCase() === 'html');
     const rss = viewableDistribution.filter(d=> d.format.toLowerCase() === 'rss');
 
+    const geoFormat = ["csvGeoAu" , "wfs" , "wms" , "czml" , "kml"];
```
kring: Should be `csv-geo-au` I think.

kring: Actually you can make it `csv-geo-` so that it'll work if csv-geo-au is generalized for other countries as well.

PR #161, opened 2017-07-24
use markdown for text preview

chloeleichen:
Fixes https://github.com/TerriaJS/magda/issues/154

Use markdown viewer to display text file rather than just dumping text, seems to work ok


kring:
Better, but still not quite right.  The problem is that markdown usually ignores single newlines.  So we get this:
image
Instead of this:
image

Some markdown renderers have an option to treat single newlines as actual newlines instead of ignoring them.  If we can turn on an option like that, it could work.

But still, rendering a text file as markdown could be a bit dodgy.  Something as simple as a `<pre>` tag with `word-wrap: break-word` might be sufficient (I say that without having tried it).

chloeleichen:
@kring you are right it looks/works much better using `pre`
see updated

PR #162, opened 2017-07-24
Broken link sleuther framework

AlexGilleran:


magda-sleuther-broken-link/src/index.ts
```diff
@@ -0,0 +1,370 @@
...
+    > = distributions.map((distribution: Record) =>
+      checkDistributionLink(
+        distribution,
+        distribution.aspects["dcat-distribution-strings"]
+      )
+    );
+
+    // Group the checks against their host so that we're only making one request per site simultaneously.
+    const brokenLinkChecksByHost: Promise<BrokenLinkSleuthingResult[]>[] = _(
+      linkChecks
```
steve9164: This will only limit per dataset, unless datasets are processed serially by the sleuther framework, in which case the retry parameters may need to tweaked more (each broken link will take ~20 seconds of retries).

steve9164:
It looks like you've accidentally committed a large GeoJSON file `magda-indexer/~/magda-regions/STE.json` in your last commit. It might be worth reverting and redoing that commit without the file so that the repo doesn't grow by ~60mb.

AlexGilleran:
OK I think I've finally got tests for the sleuther framework, broken list sleuther and have travis running them all!

magda-sleuther-broken-link/src/sleuther.ts
```diff
@@ -0,0 +1,441 @@
+import Registry from "@magda/typescript-common/dist/Registry";
+import { Record } from "@magda/typescript-common/dist/generated/registry/api";
+import unionToThrowable from "@magda/typescript-common/dist/util/union-to-throwable";
```
kring: At the risk of bike shedding, can I suggest that we always make the export name, file name, and import name all the same?  It's nice to not have to map names unnecessarily, e.g. press ctrl-p, type the name of the a symbol in one source file, and quickly end up at the definition of that symbol.

AlexGilleran: as in `import api from "@magda/typescript-common/dist/generated/registry/api"` and use it as `api.Record` ?

kring: no, as in rename `union-to-throwable.ts` to `unionToThrowable.ts`. :)

kring: But also I'd generally suggest exporting one thing per file, for the same reason.  That api.ts is an exception because it's generated.

kring:
I'm going to merge this.  The sleuther framework itself looks great.  The tests are mind boggling to me at the moment (and like 2x as many lines as the implementation itself!) but that's probably just because I don't totally get property based testing yet.  There's also a test failure on Windows, but that's a known problem that we can live with for now.

kring:
Aaand just as a I write that I see that the Travis build is now failing.  Better look into that, first.

AlexGilleran:
All good again

.gitignore
```diff
@@ -31,3 +31,4 @@ snapshots
 node_modules/
 dist/
 package-lock.json
+/magda-registry-datastore/
```
kring: We don't want to exclude this directory from git, do we?

AlexGilleran: :/ I have no idea why I did this.

kring:
What's the deal with all the magda-web-client changes?  Other than that question, and the `.gitignore` thing, I'm happy to merge this.

AlexGilleran:
The web client changes are because this makes the web-client get built in prod mode on travis, and it doesn't allow warnings in prod mode - all those slightly bad regexs created build errors and red builds :(.

Could've tried to make it allow the warnings but I figured I might as well just fix the warnings seeing how they are legit after all.

PR #164, opened 2017-07-25
use Google doc viewer

chloeleichen:
use google doc viewer to handle pdf, doc, ppt, xls, xlsx,  and image files preview, and everything else that doesn't belong to known format will be thrown to google doc viewer for a go.

Addressing https://github.com/TerriaJS/magda/issues/155 and https://github.com/TerriaJS/magda/issues/153


steve9164:
We should check the terms and conditions of Google docs before we deploy this. I couldn't find any relevant information about embedding with a quick search.

kring:
I can't see any reason this would be disallowed.  Because:

1. We don't need to sign up for an account to do this sort of embedding, so we don't need to agree to any terms of use.
2. I don't see anything in the terms of use that would forbid it anyway.
3. Google has embed support for Google Docs, and documentation on how to use it.
4. Web servers can return an HTTP header that makes browsers disallow use of the page in an iframe.  Google is not including that header.

So IMO that all adds up to Google being ok with this use case.  And why not?  It's pointing more users toward Google services.

PR #168, opened 2017-07-27
dynamically load modules that are used to parse preview data

chloeleichen:
Re: https://github.com/TerriaJS/magda/issues/157

Dynamically load parsers for the preview data to make the front end smaller 
screen shot 2017-07-27 at 3 17 46 pm

### DIFF
--- a/magda-web-client/src/Components/LazyComponent.js
+++ b/magda-web-client/src/Components/LazyComponent.js
@@ -1,0 +1,28 @@
+import React, { Component } from 'react';
+
+class LazyComponent extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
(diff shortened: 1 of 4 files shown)


steve9164:
This reduces the bundle size, but is it possible to bundle `vega`, `draft-js` or `react-json-tree`? Asynchronously loading any/all of these would reduce the bundle size much more dramatically (as would removing `react-pdf-js`)

chloeleichen:
I believe react-json-tree has already been bundled in the latest commit. I'll look into draft js and vega. In general, @steve9164 do you think this is the way to go?

chloeleichen:
close as it's included in the other PR

PR #170, opened 2017-07-28
Refactor

chloeleichen:
Using https://github.com/reactjs/react-codemod#pure-component to convert classes that only have a render method, only have safe properties (statics and props), and do not have refs to Stateless Functional Components, because it has some benefits: https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc


chloeleichen:
Updated @steve9164 can you have a look again

chloeleichen:
actually going to close this as it will be in another PR

PR #172, opened 2017-07-28
Don't expode if webhooks experience an error.

kring:
This doesn't actually fix anything, but at least will give us a stack trace when webhook processing fails, rather than silently breaking all webhook processing from then on.


kring:
eh let me clean this up to be less nully before you merge it.

kring:
ok all good now

kring:
I'm merging this to avoid conflicts with new stuff I'm working on.

PR #178, opened 2017-08-08
Deferred web hooks

kring:
This is an improvement to the registry's web hook system that allows processing of web hook notifications by a receiver to be completed asynchronously with the web hook POST itself.  Queuing of notifications is done on the registry side, where it is practically free due to the registry's event stream.  The way it works is this:

* The registry POSTs to a webhook receiver with a payload like this:

```json
{
  "action": "records.changed",
  "lastEventId": 23662,
  "records": [
    {
      "id": "c848345d-cc9c-475a-a56f-47161c9d0b83",
      "name": "Depositional Path Length - modified2",
      "aspects": {
        "dataset-distributions": {
          "distributions": []
        }
      }
    }
  ],
  "deferredResponseUrl": "http://localhost:6101/v0/hooks/24e204b7-fce2-4002-90d9-e0bbbc0407dc/ack"
}
```

* If the receiver returns anything other than a 2xx code, these events are considered undelivered and we'll retry later.  Currently we retry the next time something changes and triggers web hook processing again, no matter how soon (or long) that is.  We can do smarter things here, but that's outside the scope of this PR.

* If the receiver returns 2xx, we check to see if the response is JSON that looks like this:

```json
{
  "deferResponse": true
}
```

If it _doesn't_ look like that (it can look like anything else), we consider this payload to be sychronously delivered and we'll deliver the next set of events, if any, immediately.

But if it _does_ look like that, it means the receiver has requested asynchronous handling of this payload.  The registry will not deliver any more notifications to this particular web hook until the step below is completed.

* When the receiver has finished handling this payload, it posts to the `deferredResponseUrl` included in the payload.  The body of the request should look like this:

```json
{
  "succeeded": true,
  "lastEventIdReceived": 23662
}
```

The `lastEventIdReceived` should be the `lastEventId` from the original payload that is being acknowledged.  If `succeeded` is true, the next set of events (after the `lastEventIdReceived`), if any, will soon be posted to the webhook end point.

If `succeeded` is false, the receiver had some kind of problem and the last payload was not processed successfully.  The next payload POSTed to the receiver will include the same events as last time, possibly including any that have occurred since.

A webhook receiver can POST to the `ack` service with `"succeeded": false` at startup in order to make sure that A) notifications are not paused waiting on a response from the receiver, and B) no notifications were missed even if the receiver crashed in the middle of processing the last payload.


magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/WebHookActor.scala
```diff
@@ -31,25 +105,38 @@ object WebHookActor {
           this.processAgain = true
         } else {
           this.isProcessing = true
-          println("WebHook Processing: STARTING")
-          processor.sendNotifications().map {
+
+          println(s"WebHook ${this.id} Processing: STARTING")
```
AlexGilleran: Can we use the akka logger for this so it has log levels and is susceptible to setting the log level?

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/WebHookActor.scala
```diff
@@ -31,25 +105,38 @@ object WebHookActor {
...
-      case GetStatus => sender() ! Status(this.isProcessing)
-      case DoneProcessing(_, exception) => {
-        if (exception.isEmpty) {
-          println("WebHook Processing: DONE")
-        } else {
-          println("WebHook Processing: FAILED")
+      case GetStatus => sender() ! Status(Some(this.isProcessing))
+      case DoneProcessing(result, exception) => {
+        if (exception.nonEmpty) {
+          println(s"WebHook ${this.id} Processing: FAILED")
```
AlexGilleran: ditto re logging

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/WebHookActor.scala
```diff
@@ -31,25 +105,38 @@ object WebHookActor {
...
           exception.get.printStackTrace()
         }
 
+        val runAgain = result match {
+          case None => false
+          case Some(WebHookProcessingResult(_, _, true, _)) => {
+            // response deferred
+            println(s"WebHook ${this.id} Processing: DEFERRED BY RECEIVER")
```
AlexGilleran: ditto re logging

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/WebHookActor.scala
```diff
@@ -31,25 +105,38 @@ object WebHookActor {
...
+            false
+          }
+          case Some(WebHookProcessingResult(previousLastEvent, newLastEvent, false, _)) => {
+            // POST succeeded, is there more to do?
+            println(s"WebHook ${this.id} Processing: SUCCEEDED")
```
AlexGilleran: ditto re logging

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/WebHookActor.scala
kring: Yep, good idea.  Though consider merging it first since the version in master does the same thing and this PR also fixes problems running connectors and things.

PR #180, opened 2017-08-10
show preview by default

chloeleichen:
remove preview tab on dataset page, rather we give user the option to view preview for each distribution on dataset page, with a toggle


chloeleichen:
going to close this too as this change be part of a bigger feature

PR #186, opened 2017-08-28
Helm

kring:
This adds a Helm chart for Magda.  There's still more to do, but since this only adds new files in the Helm directory, there's not much harm in PRing it now for an early look.

You can deploy to our different environments with:
* Production: `helm install -f deploy/helm/search-data-gov-au.yml deploy/helm/magda`
* Dev: `helm install -f deploy/helm/magda-dev.yml deploy/helm/magda`
* Minikube: `helm install -f deploy/helm/minikube.yml deploy/helm/magda`

Or specify configuration values on the command-line or in a separate config file.


AlexGilleran:
If you cherry pick https://github.com/TerriaJS/magda/pull/183/commits/241006a452dda07ade9cf87cc36f3ef730bc9551 it should fix the build.

kring:
Ok, thanks.  There was a breaking change in lru-cache or something?  It'd be good to mention it in the commit message so that later when we're wondering why that version is pinned, it's easy to find an explanation and see if it still holds.

deploy/helm/magda/charts/web-server/values.yaml
```diff
@@ -0,0 +1,21 @@
+# Default values for gateway.
```
AlexGilleran: It's web-server not gateway now right?

AlexGilleran:
> There was a breaking change in lru-cache
Yeah in `@types/lru-cache` :(

kring:
This is ready.

PR #195, opened 2017-09-01
Admin API, Helm Updates

AlexGilleran:
- Adds a first run admin API that can create new jobs, start them, stop them and delete them
Helm:
- Increases maxUnavailable to 1 in dev so the images swap faster
- Changes image pullPolicy to IfNotPresent in prod so that pods restart quicker
- Adds default nodeports to pods in minikube so you can access them directly - these match their localhost ports, `-6000 + 30000`.
- Updates the dev doco to use helm
- Makes it possible to select what pods you want to run in helm


AlexGilleran:
- Renamed everything with `auth` to `authorization`
- Renamed `ld` to `linked-data`
- Moved `authorization-api` client into `typescript-common`

kring:
Thanks @AlexGilleran!

PR #198, opened 2017-09-12
Admin api tests

AlexGilleran:
There's a lot of indendation fixing in this - the important parts that actually changed are in magda-admin-api/src.

Also hoisted express and @types/express which seems to have appeased the typescript kraken.


magda-gateway/src/createApiRouter.ts
```diff
@@ -36,7 +36,7 @@ export default function createApiRouter(options: ApiRouterOptions): Router {
     ) {
         if (jwtSecret && req.user) {
             const token = jwt.sign(
-                { userId: req.user.id, isAdmin: req.user.isAdmin },
+                { userId: req.user.id },
```
kring: What's the reason for this change?  Unless you have some other scheme going, I think we do need that `isAdmin` flag in the JWT for now.  Obviously it's not the final solution.

AlexGilleran: I took it out because we weren't using it anywhere - the admin api looks up whether or not the user is an admin on a per request basis and the discussions api doesn't have any admin-restricted features yet.

The reason I didn't use it is because it's simpler not to and I'm not sure it's a great idea in general - it doesn't scale to a more fine-grained permissions system (presumably at some point everything the user has access to do will be too big to practically fit in a JWT), plus doing it this way means that when you edit a user's permissions you'll need to either invalidate their session or edit its contents in the session store.

kring: I agree with everything you said, except that we don't use it.  This is coming from the JWT right?
https://github.com/TerriaJS/magda/blob/master/magda-web-client/src/Project/ProjectDetails.js#L25

kring: Oh wait that'd come from the cookie, not the JWT.  Nevermind me stupid.

PR #205, opened 2017-09-26
Admin

chloeleichen:
a basic connector admin interface: 
http://localhost:6108/admin/connectors


magda-web-client/src/Components/Admin/Connectors.js
```diff
@@ -1,45 +1,143 @@
...
 
   renderByUser(user){
     if(!user){
-      return <Login
-        signInError={
-          this.props.location.state && this.props.location.state.signInError
-        }
-        providers={this.props.providers}
-      />
+      return <div> <Link to="/account">Sign in</Link> as admin to veiw a list of connectors available</div>
```
AlexGilleran: view

magda-web-client/src/Components/Admin/Connectors.js
```diff
@@ -1,45 +1,143 @@
...
+            </form>
+          </div>
+  }
+
+  submitNewConnector(){
+    if(this.state.newConnectorName.length > 0 && this.state.newConnectorType.length > 0){
+      this.props.createConnector({
+        name: this.state.newConnectorName,
+        type: this.state.newConnectorType,
+        id: uuidV1()
```
AlexGilleran: Ideally the `id` should be something identifiable that makes sense when you're looking at jobs/pods in kubectl. Maybe even url-encode the name?

magda-web-client/src/Components/Admin/Connectors.js
```diff
@@ -1,45 +1,155 @@
...
+  deleteConnector(connector){
+    this.props.deleteConnector(connector.id)
+  }
 
+  renderConnector(connector){
+    return (<tr key={connector.id} className={`${connector.name === this.state.newConnectorName ? 'success': ''}`}>
+    <td>{connector.name}</td>
+    <td>{connector.type}</td>
+    <td>{connector.sourceUrl}</td>
+    <td><button className={`btn ${!connector.job ? 'btn-success': 'btn-warning'}`} type='button' onClick={this.toggleConnector.bind(this, connector)}>{!connector.job ? 'Start' : 'Stop'}</button></td>
```
AlexGilleran: this needs to be something like `connector.job && connector.job.status === 'active'`. Connectors can also have `job.status === 'succeeded'` and `job.status === failed` :).

magda-web-client/src/Components/Admin/Connectors.js
```diff
@@ -1,45 +1,157 @@
...
+        type: this.state.newConnectorType,
+        id: encodeURI(this.state.newConnectorName)
+      });
+      this.setState({
+        newConnectorFormIsOpen: false
+      })
+  }
+
+  toggleConnector(connector){
+    const action = connector.job ? 'stop' : 'start';
```
AlexGilleran: Needs to be changed here too right?

chloeleichen: done

PR #206, opened 2017-09-26
Made postgres automatically back up

AlexGilleran:
Our postgres image now has WAL-E baked into it. When the container env `$BACKUP==WAL`, it will make a daily full backup (keeping the last 30) to some kind of storage backend (Google Storage, S3, SWIFT or Azure Block Store), and stream incremental Write-Ahead-Log backups. If the container starts and finds its $PGDATA volume is brand new, it'll pull the latest full backup and then replay the write-ahead-log onto it before Flyway runs. If it starts but finds that it's already run before with this volume, it just starts up with the existing data as normal.

With `$BACKUP=true` it'll simply make the daily backup without WAL streaming or automatic recovery.

There's also some really basic Postgres tuning too - if a memory limit is set in helm it'll use it to set shared_buffers and effective_cache_size to their conservative recommended values.


deploy/helm/minikube-dev.yml
```diff
@@ -4,9 +4,19 @@ global:
...
     googleClientId: "275237095477-f7ej2gsvbl2alb8bcqcn7r5jk0ur719p.apps.googleusercontent.com"
+combined-db:
+  image:
+    tag: "latest"
+  waleBackup:
+    method: WAL
+    gsPrefix: "gs://magda-postgres-backups/test"
+    executionTime: 03:00
+    googleApplicationCreds:
+      secretName: storage-account-credentials
```
kring: Will this prevent the minikube deployment from working if the `storage-account-credentials` secret doesn't exist?

AlexGilleran: Oops - yeah I forgot to remove the google storage config from minikube, ideally it shouldn't be using that at all unless you opt-in to it.

PR #210, opened 2017-10-05
Restrict registry to admins

AlexGilleran:
This turned out to be pretty big.

- PUT/POST/PATCH/GET `/hooks` in the registry now requires an `X-Magda-Session` header containing a userId that when checked in the admin API returns an `isAdmin` flag of `true`.
    - Created new auth directives to make this work - these take care of getting the header, decrypting it and querying the auth api.
- Adjusted unit tests to make these requests as admin, added tests to make sure that requests to these routes without admin access fail
- Made all the services that write to the registry identify themselves as an admin
- Connectors, Sleuthers, the Admin API and the Indexer now all get the JWT secret and a userId to use passed into them
- The auth database now has a default admin user created by flyway
- I made a lot of the node services that were accepting config via the command line do so, falling back to env vars
- Tore the remaining crawling logic out of the indexer - now it only knows to crawl the registry rather than having extra code to support CKAN, CSW etc.
    - Removed the `source` field from the search schema seeing how it's always registry now
- Made `/admin` work from `magda-web-server` instead of returning 404 - still doesn't work because it appears the UI is hard coded to dev :(.
- Made the minikube docker-registry use a PV so it doesn't occasionally lose all your images.
- Mandated double-quotes in javascript. Could make it single-quotes too, but right now we've got a mix.


.editorconfig
```diff
@@ -13,3 +13,6 @@ indent_size = 2
 
 [*.md]
 trim_trailing_whitespace = false
+
+[*.js]
+quote_type = double
```
kring: This doesn't seem to do anything in VS Code.  Plus most of our code is `*.ts`.

PR #222, opened 2017-10-12
Database passwords

### LINKED ISSUE #218: Lock down database with passwords
Right now they're running without passwords which is fine as long as no one gets into the network, but we should improve it.

AlexGilleran:
Fixes #218 

- By default all postgres containers are passed a POSTGRES_PASSWORD env var, which means that the `postgres` user gets a password and the container has access from outside 127.0.0.1 locked except for users. With combined-db this is the same for all databases (naturally), when the databases are split they can be configured all be different
- Every database gets a `client` user created it for it that only has privileges to read/write data but not do terrible things like drop tables
- Now all the containers that connect to the database use the `client` user rather than `postgres`
- All this can be disabled by setting `global.noDbAuth` to true

On the way I ended up completely ruining my minikube VM and having to recreate it from scratch, so I made some other improvements:
- Made the connector screen not hard coded to magda-dev anymore so you can use it for local crawling.
- Renamed the `config` configmap to `web-app-config` and made it created by helm.


AlexGilleran:
Oop, this is actually slightly broken for combined-db, fixing...

AlexGilleran:
OK fixed :)

PR #225, opened 2017-10-17
Stopped printing jwtSecret to command line on fail

### LINKED ISSUE #219: Make sure we don't print secrets on command line
Right now we're allowing the user to override particularly JWT_SECRET via the command line, and defaulting it to the JWT_SECRET env variable. This means that if the right args don't get passed, the secret will get printed to the command line 😬

AlexGilleran:
Fixes #219

### DIFF
--- a/magda-admin-api/src/index.ts
+++ b/magda-admin-api/src/index.ts
@@ -1,63 +1,64 @@
 import * as express from "express";
-import buildApiRouter from "./buildApiRouter";
 import * as yargs from "yargs";
 
-const argv = yargs
-    .config()
-    .help()
-    .option("listenPort", {
-        describe: "The TCP/IP port on which the admin api should listen.",
-        type: "number",
-        default: 6112
-    })
-    .option("dockerRepo", {
-        describe:
-            "The docker repo to look for docker images for creating connectors etc.",
-        type: "string",
-        default: "localhost:5000/data61"
-    })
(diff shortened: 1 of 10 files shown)


magda-admin-api/src/index.ts
```diff
@@ -50,13 +50,14 @@ const argv = yargs
         default: process.env.USER_ID || process.env.npm_package_config_userId
     })
     .option("jwtSecret", {
-        describe:
-            "Secret for decoding JWTs to determine if the caller is an admin",
+        // describe:
```
kring: Did you mean to comment out these lines?

kring:
Looks like you missed `magda-ckan-connector`.

kring:
After these changes, it will still print the secrets to the console if you run with `--help`.  I'm not totally sure what the threat model is here so I can't really say if that's "as bad" a problem, but it seems less than ideal, at least.

I think it'd be better to `Object.assign`-in the secrets after yargs gets done with it rather than disabling help on failure.  That would prevent the secrets from being shown even when you run with `--help` and it would give us back the (useful) help when you run it incorrectly.

AlexGilleran:
Assuming these changes don't end up 🐙 -ing...

Changed the whole approach so that there's a `addJwtSecretFromEnvVar` function that accepts an `argv` object, checks whether it's got a `jwtSecret` and if not adds one from env or package.json.

PR #230, opened 2017-10-19
markdown description

chloeleichen:
Fixing https://github.com/TerriaJS/magda/issues/229, ~probably~ definitely in an overly complicated way, happy to hear what you think.


kring:
@chloeleichen you should use the same approach we use in TerriaJS.  The one here is potentially vulnerable  to XSS attacks and that sort of thing.  Remember that markdown can contain HTML, and markdown-it doesn't sanitize it, so using `dangerouslySetInnerHTML` is ... dangerous!

magda-web-client/src/UI/MarkdownViewer.js
```diff
@@ -1,36 +1,29 @@
...
+  constructor(props) {
+   super(props);
+   this.state = {isCollapsed: true};
+ }
 
-function MarkdownViewer(props) {
-    let markdown = {__html: marked(props.markdown)};
-    return <div className='markdown' dangerouslySetInnerHTML={markdown}/>
+ componentDidMount(){
+   const numberOfChildrenFromMarkdownAsHtml = this.markdown.children.length;
```
kring: This should probably happen more often than just the initial mount, right?  Like if the markdown changes and the content length increases, a user of `MarkdownViewer` would probably expect `updateContentLength` to be called again.

chloeleichen:
updated, @kring can you have a look again.

kring:
@chloe do you remember why we ended up using `html-to-react` rather than `dangerouslySetInnerHTML` in TerriaJS?  I'm just curious.  This should be ok from an XSS perspective now, at least.

kring:
Please merge in master (there are conflicts) and then I'll merge this.

chloeleichen:
I'm going to wait for https://github.com/TerriaJS/magda/pull/235 to merge first as there are some directory changes in there

PR #233, opened 2017-10-20
Web connector react router v4

chloeleichen:
Using reacr-router v4, fixes the issue we've been having on serving the app at sub directory(`/admin` in this case), at least on my machine. 

It's also smaller. 

might want to merge https://github.com/TerriaJS/magda/pull/235 first as it updates react-router in `magda-web-client`


chloeleichen:
don't merge this yet

PR #238, opened 2017-11-01
Generate a sitemap that includes all the dataset files

### LINKED ISSUE #179: Make web interface search engine friendly
So that Google et al can find and index all the datasets.

AlexGilleran:
- index.html now links to `/sitemap`
- `/sitemap` generates a sitemap index which links to a standard sitemap that just has the index, and also links to sitemaps for every 100 datasets by linking to urls that have the pageToken as a parameter
- To get these page tokens, the web pod calls a new endpoint in the registry at `/records/pagetokens`, which queries the database to get tokens for each page given a certain `limit`.
- Each datasets sitemap gives the url of 100 datasets after the provided pagetoken.

Fixes #179

### DIFF
--- a/magda-web-server/package.json
+++ b/magda-web-server/package.json
@@ -17,7 +17,7 @@
     "@magda/scripts": "^0.0.30-0",
     "@types/chai": "^4.0.1",
     "@types/config": "0.0.32",
-    "@types/express": "^4.0.35",
+    "@types/express": "^4.0.39",
     "@types/lodash": "^4.14.68",
     "@types/mocha": "^2.2.41",
     "@types/nock": "^8.2.1",
--- a/magda-web-server/package.json
+++ b/magda-web-server/package.json
@@ -22,7 +22,9 @@
   },
   "dependencies": {
     "@magda/web-client": "^0.0.30-0",
+    "@magda/typescript-common": "^0.0.30-0",
     "express": "^4.15.3",
+    "sitemap": "^1.13.0",
     "urijs": "^1.18.12",
     "yargs": "^8.0.2"
   },
--- a/magda-web-server/package.json
+++ b/magda-web-server/package.json
@@ -10,15 +10,30 @@
     "start": "node dist/index.js",
     "dev": "run-typescript-in-nodemon src/index.ts",
     "docker-build-local": "create-docker-context-for-node-component --build --push --tag auto --local",
-    "docker-build-prod": "create-docker-context-for-node-component --build --push --tag auto"
+    "docker-build-prod": "create-docker-context-for-node-component --build --push --tag auto",
+    "test": "mocha --compilers ts:@magda/scripts/node_modules/ts-node/register --require @magda/scripts/node_modules/tsconfig-paths/register \"src/test/**/*.spec.ts\""
   },
   "devDependencies": {
     "@magda/scripts": "^0.0.30-0",
+    "@types/chai": "^4.0.1",
     "@types/config": "0.0.32",
     "@types/express": "^4.0.35",
+    "@types/lodash": "^4.14.68",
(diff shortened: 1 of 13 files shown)


magda-web-server/package.json
```diff
@@ -10,19 +10,36 @@
...
+    "supertest": "^3.0.0",
+    "typed-promisify": "^0.4.0",
+    "typescript": "~2.5.0",
+    "xml2js": "^0.4.19"
   },
   "dependencies": {
     "@magda/web-client": "^0.0.30-0",
+    "@magda/typescript-common": "^0.0.30-0",
     "express": "^4.15.3",
+    "sitemap": "^1.13.0",
```
kring: This should be a devDependency.  It's needed a build time, but not at runtime.

magda-web-server/src/index.ts
```diff
@@ -16,8 +20,22 @@ const argv = yargs
...
+        describe:
+            "The absolute base URL of the Magda site, when accessed externally. Used for building sitemap URLs which must be absolute.",
+        type: "string",
+        default: "http://localhost:6100/",
+        required: true
+    })
+    .option("registryApiBaseUrlInternal", {
+        describe: "The url of the registry api for use within the cluster",
+        type: "string",
+        default: "http://localhost:6101/",
```
kring: this should be `http://localhost:6101/v0/` right?  I had to change it to that to make it work outside the cluster, at least.

magda-web-server/src/buildSitemapRouter.ts
```diff
@@ -0,0 +1,123 @@
...
+        catchError(
+            res,
+            registry
+                .getRecordsPageTokens(DATASET_REQUIRED_ASPECTS)
+                .then(handleError)
+                .then(async result => {
+                    const datasetsPages = result.map(token => {
+                        return (
+                            baseExternalUrl +
+                            "/sitemap/dataset/afterToken/" +
```
kring: Use urijs to build the URL so that it's not sensitive to whether or not `baseExternalUrl` has a trailing slash.

magda-web-server/src/buildSitemapRouter.ts
```diff
@@ -0,0 +1,123 @@
...
+                            token
+                        );
+                    });
+
+                    const smi = sm.buildSitemapIndex({
+                        urls: [baseExternalUrl + "/sitemap/main"].concat(
```
kring: Same here.

kring:
Looks good other than those minor comments.

magda-web-server/src/buildSitemapRouter.ts
```diff
@@ -0,0 +1,123 @@
...
+            cacheTime: 600000,
+            urls: [
+                {
+                    url: ``,
+                    changefreq: "weekly"
+                }
+            ]
+        });
+
+        sitemap.toXML(function(err: Error, xml: string) {
```
AlexGilleran: Here's an example if where it's used at runtime.

magda-web-server/package.json
AlexGilleran: Are you sure? The sitemap generation is all done on-the-fly, not as a build thing, right... or did I mess it up? 😬

magda-web-server/src/index.ts
AlexGilleran: Oops, sorry!

magda-web-server/package.json
kring: Oops sorry you're right.  For some reason I thought I was looking at client-side code when I saw the bit that required-in sitemap.  But it's node.js code, so yeah it needs to be in dependencies.

AlexGilleran:
@kring fixed those things :)

PR #244, opened 2017-11-03
Bumped version to 0.0.30-0 for further dev.

AlexGilleran:


kring:
This is OBE (all packages use version 0.0.31-0 now).

PR #246, opened 2017-11-10
Update lerna version in package.json

steve9164:
Update lerna version in package.json to match version in lerna.json.
Currently `npm run build` fails due to incompatible lerna version. Changing the version here should fix that problem.


steve9164:
Note: `npm run build` will only fail currently if you have run `npm install` in the root directory (because that install an old `lerna` which gets called by `npm` scripts)

PR #247, opened 2017-11-17
Visualisation Sleuther

steve9164:
Currently marks csv formats with a visualisation-info aspect that contains information like:
- validity of csv
- fields in csv, and whether each is a numeric or time column
- whether a csv has both a numeric column and a time column, in which case it is considered a timeseries csv

I think I've updated and added the helm scripts that are needed, but I'm not sure how to test them all.


kring:
Looks good!  I'm just going to rename `visualisation` to `visualization` because mixing spelling is confusing (e.g. we already have `magda-authorization-api`) and then I'll merge this.

kring:
Since I was questioning my assertion that we should be using American spelling, I went looking for some other opinions and found this:
https://stackoverflow.com/questions/157807/gb-english-or-us-english
Which I think makes a pretty strong case for using American spelling in all our identifiers even if we don't like it.

PR #261, opened 2017-12-01
add timestamp to when connectors retrieve data externally. 

### LINKED ISSUE #101: Connectors should capture the "last retrieval date" in the source aspect of the registry
This way we can tell how fresh the registry's knowledge of the dataset is.

tristochief:
fix #101


magda-ckan-connector/src/Ckan.ts
```diff
@@ -46,6 +46,7 @@ export interface CkanOptions {
     maxRetries?: number;
     secondsBetweenRetries?: number;
     ignoreHarvestSources?: string[];
+    retrievedAt: number;
```
kring: Use a `moment` instead of a `number` / `Date`.

magda-ckan-connector/src/createTransformer.ts
```diff
@@ -28,7 +28,8 @@ export default function createTransformer({
             URI: URI,
             ckan: new CkanUrlBuilder({
                 name: name,
-                baseUrl: sourceUrl
+                baseUrl: sourceUrl,
+                retrievedAt: -1
```
kring: The `dataset-source` aspect builder is going to access this property, so if you set -1 here I'm pretty sure all the datasets will have an invalid `retrievedAt` date.

magda-ckan-connector/src/index.ts
```diff
@@ -22,6 +22,11 @@ const argv = addJwtSecretFromEnvVar(
             type: "string",
             demandOption: true
         })
+        .option("retrievedAt", {
```
kring: Harvesting can take a long time.  Ideally the `retrievedAt` date would indicate when the request for that particular dataset was initiated, not when the overall process began.

magda-csw-connector/src/CswUrlBuilder.ts
```diff
@@ -34,6 +38,8 @@ export default class CswUrlBuilder {
     constructor(options: CswUrlBuilderOptions) {
         this.name = options.name || 'CSW';
         this.baseUrl = new URI(options.baseUrl);
+        this.retrievedAt = moment(options.retrievedAt).toISOString() || moment(Date.now()).toISOString();
+        console.log("built CSW Url with retrievedAt:"+this.retrievedAt);
```
kring: Debug code?  You should remove this.

kring:
I think we really want a per-record `retrievedAt` field, but it's going to be a bit tricky.  The best approach I can think of off the top of my head is to change methods like `getJsonDataset` in `ConnectorSource` in `JsonConnector.ts` to return `Promise<Dataset>` instead of `Promise<any>`.  `Dataset` is an interface that has a `retrievedAt` field in addition to a `dataset` field which holds what the Promise previously held directly (the dataset JSON).  Then it's the responsibility of the source to fill in the `retrievedAt` field based on when it actually did the retrieval.

kring:
@tristochief did you see my comments above?

kring:
@tristochief I'm reverting the merge because this still needs more work.  Please don't merge your own PRs into master.

tristochief:
hi kevin sorry about that, I was trying to merge master into
trial-retrievedBy

PR #269, opened 2017-12-07
catch error when stats not available

chloeleichen:
as title suggested


kring:
@chloeleichen there's storybook stuff in here too.  Is that intentional?

chloeleichen:
nope i'll get rid of it

PR #277, opened 2017-12-12
Storybook

chloeleichen:
Added storybook as a styleguide + ui component testing tool, live here: http://magda-storybook.surge.sh/


AlexGilleran:
Are you still going with this @chloeleichen or do you want to merge it now?

PR #278, opened 2017-12-13
added retrievedAt to dataset schemas, and required function changes t…

tristochief:
#fix add retrieveAt to records.


kring:
@tristochief please merge in master.

PR #279, opened 2017-12-15
Add script to build / dockerize changed components

kring:
Added `npm run build-changed` script at the root. When run, it compiles any changed components and builds/pushes new Docker images for changed components to `localhost:5000`.

One limitation is it doesn't take `node_modules` changes into account, so if you change dependencies you'll need to build manually.  But it _does_ know that if you change `magda-typescript-common`, all the things that depend on it need to be built.

Another potential problem is if you kill the docker build before it finishes (or it fails because you forgot to run `eval $(minikube docker-env)` or something), it won't retry the docker build the next time you run unless the code also changed again.


project/DeployPlugin.scala
```diff
@@ -18,8 +18,8 @@ object DeployPlugin extends AutoPlugin {
       println(rmi)
 
       // Delete pods associated with this deployment, so they spin back up with the new version
-      val delete = Process("kubectl delete pods -l service=" + baseDirectory.value.getName().replace("magda-", "")).!!
-      println(delete)
+      // val delete = Process("kubectl delete pods -l service=" + baseDirectory.value.getName().replace("magda-", "")).!!
```
AlexGilleran: Did you mean to leave this commented? Do we not want to do this any more?

kring: Yeah, I meant to comment it.  I was making the behavior of the Scala docker build consistent with the TypeScript.  I could have gone the other way (make the TypeScript restart the pods, too), but I haven't found the behavior to be particularly useful.  Have you?

AlexGilleran: Ok cool

PR #280, opened 2017-12-18
Testing yarn (for discussion)

steve9164:
This switches `npm` out for `yarn` as our package manager. This shrinks the time to set up a clean install of magda from hours to minutes, reduces our `package-lock` hell and saves disk space through deduplication of dependencies.

Unfortunately it also removes the ability to install deps for only one submodule (although there is a yarn RFC regarding this) so it required some pretty serious Gitlab surgery. Travis is much faster though 👍 .

It also required the logic around how we build docker images to change to support the way that yarn arranges its packages and particularly the differing output of the `yarn list --json` command.

- [x] No unit tests
- [x] Changes
- [x] Linked to an issue.

---
Old explanation:

Changes lerna's role from being a package manager (which it's often not the best at) to being a versioner, script runner etc. and uses `yarn` for package management. Changes clean package install time from ~5 minutes to less than 1 minute.

Not complete yet (having troubles with `eslint` plugins within terriamap, and I might have missed some other problems)

Biggest change from `lerna -> yarn` is that local packages symlinks are hoisted:
Old:
- `magda/`
    + `magda-ckan-connector/`
        * `node_modules/`
            - `@magda/typescript-common`
    + `magda-other-package/`
        * `node_modules/`
            - `@magda/typescript-common`

New:
- `magda/`
    + `magda-ckan-connector/`
    + `magda-other-package/`
    + `node_modules/`
        * `@magda/typescript-common`

This creates some problems with build configs that assumed `@magda` packages would be  in `node_modules` (most of which I think I have fixed, I might have missed a few).

No doc changes yet, just testing the benefits (and any possible disadvantages) of using `yarn`.


kring:
One concern with the intra-repo dependencies beind hoisted is that we can forget to add add a dependency to package.json and stuff will sorta still work. May not be a big problem, though.  Are dependencies from outside the repo hoisted as well?   That will be more concerning because it will make each microservice _much_ bigger (and therefore take much longer to dockerize).

steve9164:
Testing yarn vs old lerna on TerriaJS/magda@b28f0d329

Old lerna:
- `lerna bootstrap`: 10m:00s
- Magda dir size: 2.7GB
- Sum of `node_modules` using `find . -name node_modules -maxdepth 2 -exec du -hcd0 {} +`: 2.5G
- Adding a `@magda` package: have to edit package.json with correct version etc.
- Adding external package: can use `npm install package@v --save[-dev]`
- `lerna clean`: 54s

Lerna with yarn:
- `yarn install` (with no lockfile): 2m03s
- `lerna bootstrap`: 5s (is this even neccessary? I think you really just use `lerna` commands for running stuff and versioning in yarn approach)
- Magda dir size: 1.0GB
- Sum of `node_modules`: 856MB
- Adding a `@magda` or external package: `yarn workspace '@magda/package-that-needs-new-dep' add new-dep [-- --dev]`
  + For some reason this command takes ~25 seconds to complete (even with an existing lockfile), but at least it should be exactly the same structure at that point as if it was just bootstrapped

Unfortunately I couldn't get docker builds working with `yarn` structure, since the builds assume that magda components are in component/node_modules, instead of hoisted at ../node_modules

Benefits of yarn:
- `lerna bootstrap` -> `yarn install && lerna bootstrap` takes 30 seconds instead of >7 minutes
- A single lockfile
- No need to edit all the package.json
- Smaller magda directory on disk
- No `lerna` needed on CI (I think)

Disadvantages of yarn:
- `yarn workspace` command still has no documentation, isn't entirely consistent with other yarn commands and doesn't have great error messages (I had problems with trailing slashes on package directories listed in workspaces in package.json)
- Need more changes to the way docker builds work and some of the config (like tsconfig)
- Possibly larger docker builds?


(Feel free to add to the above lists)

AlexGilleran:
Putting this on hold until `yarn list` works well enough for us to build small docker images.

steve9164:
What exactly is needed (from `yarn` and/or us) to progress this further?

PR #281, opened 2017-12-20
Duplicate sleuther4

tristochief:
@AlexGilleran  edited as per your comments!


magda-sleuther-duplicate-records/src/onRecordFound.ts
```diff
@@ -0,0 +1,147 @@
...
+                    type: "accessURL" as "accessURL"
+                } || { type: "none" };
+
+                distContainer[1].url = {
+                    url: distribution.aspects["dcat-distribution-strings"]
+                        .downloadURL as string,
+                    type: "downloadURL" as "downloadURL"
+                } || { type: "none" };
+
+                distContainer[0].id = distContainer[1].id = distribution.id;
```
AlexGilleran: Might be easier to write that whole thing in one line as an array literal... something like

```javascript
            const distContainer: DistContainer[] = [
                 {
                     url: distribution.aspects["dcat-distribution-strings"].accessURL ? {
                         url: distribution.aspects["dcat-distribution-strings"].accessURL as string,
                         type: "accessURL" as "accessURL"
                     } : { type: "none"},
                     id: distribution.id
                 },
                 {
                     url: distribution.aspects["dcat-distribution-strings"].downloadURL ? {
                         url: distribution.aspects["dcat-distribution-strings"].downloadURL as string,
                         type: "downloadURL" as "downloadURL"
                     } : { type: "none"},
                     id: distribution.id
                 })
            ];
```

magda-sleuther-duplicate-records/src/onRecordFound.ts
```diff
@@ -0,0 +1,147 @@
...
+                    },
+                    id: null
+                });
+
+                
+                distContainer[0].url = {
+                    url: distribution.aspects["dcat-distribution-strings"]
+                        .accessURL as string,
+                    type: "accessURL" as "accessURL"
+                } || { type: "none" };
```
AlexGilleran: I'm not sure exactly what this line is supposed to be doing - the `|| { type : "none" }` bit can never execute because the object before `||` is always going to be truthy. I'm guessing this is supposed to check whether there's an `accessURL` and set `{type: none}` if not?

magda-sleuther-duplicate-records/src/onRecordFound.ts
```diff
@@ -0,0 +1,147 @@
...
+                    (distribution.aspects["dcat-distribution-strings"]
+                        .downloadURL as string)
+                ) {
+                    distContainer.pop();
+                }
+
+                return distContainer;
+            }
+
+            throw new Error(
```
AlexGilleran: Even though it's not strictly necessary, it'd be good to put this in an `else` block so it's obvious that it's not supposed to be on the main path of execution.

magda-sleuther-duplicate-records/src/onRecordFound.ts
```diff
@@ -0,0 +1,147 @@
...
+    )
+        .groupBy(distContainer => {
+            return distContainer.url.url;
+        })
+        .value();
+
+    // Construct a datatype that represents the format of either an aspect or record (its an aspect representation atm) so that we can easily put it
+    //into the database.
+    let keys = getKeys(groupedDistContainers);
+    var groups: DuplicateRecordsAspect[] = [];
```
AlexGilleran: These declarations can both be `const` :)

magda-sleuther-duplicate-records/src/onRecordFound.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (mochaObject.isRunning) testGroups = groups;
+    return Promise.resolve([]);
+}
+
+export function testGetKeys(dictionary: any): Array<string> {
+    if (mochaObject.isRunning) return getKeys(dictionary);
+    throw new Error("This function is called only when testing with mocha");
+}
+
+function getKeys(dictionary: any): Array<string> {
```
AlexGilleran: You can use `Object.keys` for this I think :)

magda-sleuther-duplicate-records/src/test/urlsFromDataSet.ts
```diff
@@ -0,0 +1,13 @@
+import { Record } from "@magda/typescript-common/dist/generated/registry/api";
```
AlexGilleran: Once again, move this into typescript-common and refer to it from both here and the broken link sleuther rather than copying it :)

magda-sleuther-duplicate-records/src/onRecordFound.ts
```diff
@@ -93,55 +90,31 @@ export default async function onRecordFound(
...
-                url: null,
-                ids: []
-            };
-
-            duplicateRecordsAspect.url = key;
-
-            groupedDistContainers[key].forEach(function(groupedDistContainers) {
-                duplicateRecordsAspect.ids.push(groupedDistContainers.id);
-            });
+            const dduplicateRecordsAspect: DuplicateRecordsAspect = {
```
AlexGilleran: `dduplicate` -> `duplicate` :)

magda-sleuther-duplicate-records/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+import onRecordFound from "./onRecordFound";
+import duplicateRecordsAspectDef from "./duplicateRecordsAspectDef";
+import datasetQualityAspectDef from "@magda/sleuther-framework/dist/common-aspect-defs/datasetQualityAspectDef";
+import commonYargs from "@magda/sleuther-framework/dist/commonYargs";
+
+const ID = "sleuther-duplicate-records";
+
+const argv = commonYargs(ID, 6112, "http://localhost:6112");
+
+function sleuthduplicateRecordss() {
```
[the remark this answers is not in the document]
AlexGilleran: This still needs changing

PR #283, opened 2018-01-03
Improve Speed and Reliability of WebHooks

### LINKED ISSUE #240: Webhook resuming doesn't seem to work on prod

### LINKED ISSUE #241: Indexing via webhooks is reaaaaally slow
Why?

AlexGilleran:
Fixes #240, Fixes #241.

**Reliability**
I'm still not entirely sure of all the cases that were causing webhooks to stop running... and this still won't completely resolve it because sometimes there's nothing you can do (what happens if the sleuther finds data that it just can't deal with?) except alert an administrator somehow. However, I've made super sure that sleuthers always resume their webhook on startup, and that the registry always starts sending hooks to all its subscribers (even async ones that are marked as waiting) on startup, so if a sleuther or the registry goes down, business-as-usual resumes as soon as it comes back up.

**Speed**
Paging through events was taking a really long time before - there's still a lot of little improvements we can make here but a big deal was that the webhooks were retrieved from the database page-by-page without filtering, then filtered outside the database. This meant that a webhook actor could go through hundreds of pages (each taking a second or two) without ever actually finding a relevant event to raise a hook for. In particular the broken link sleuther would take a single record created event then generate several events (one for each distribution + a quality one), which its own webhook actor would then have to retrieve from the database and process, even though it obviously has no interest in processing its own output.

So now filtering of events is done in SQL wherever - this means that if, say, the indexer wakes up and resumes its hook, but the next 5000 events are all for the broken link aspect (which it doesn't listen to), all those events will be skipped in one DB call.

This got particularly challenging when it comes to the case of a webhook that listens to an aspect that links to another aspect which is turn modified - but I'm *pretty sure* I've solved this with an SQL subquery - at least enough to pass unit tests.

**Other Stuff**
- Rather than refresh the webhook actors every time the webhooks are processed, I've made it so that it builds it at the start and rebuilds it when it gets posted an invalidation message from the hooks service, which happens when the hooks are modified. Should save a few db queries.
- Stopped the sleuthers from doing everything in parallel, as this was causing them to take all the registry's open ports and ironically end up going much slower.

### DIFF
--- a/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
+++ b/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
@@ -39,15 +39,15 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
   def create = post {
     pathEnd {
       entity(as[WebHook]) { hook =>
-        DB localTx { session =>
-          val result = HookPersistence.create(session, hook) match {
+        val result = DB localTx { session =>
+          HookPersistence.create(session, hook) match {
             case Success(result) =>
               complete(result)
             case Failure(exception) => complete(StatusCodes.BadRequest, BadRequest(exception.getMessage))
           }
-          webHookActor ! WebHookActor.InvalidateWebhookCache
-          result
         }
+        webHookActor ! WebHookActor.InvalidateWebhookCache
+        result
       }
     }
   }
@@ -81,15 +81,15 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
     path(Segment) { (id: String) =>
       {
         entity(as[WebHook]) { hook =>
-          DB localTx { session =>
-            val result = HookPersistence.putById(session, id, hook) match {
+          val result = DB localTx { session =>
+            HookPersistence.putById(session, id, hook) match {
               case Success(result) =>
                 complete(result)
               case Failure(exception) => complete(StatusCodes.BadRequest, BadRequest(exception.getMessage))
             }
-            webHookActor ! WebHookActor.InvalidateWebhookCache
-            result
           }
+          webHookActor ! WebHookActor.InvalidateWebhookCache
+          result
         }
       }
     }
@@ -108,11 +108,11 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
         val result = DB localTx { session =>
           HookPersistence.delete(session, hookId) match {
             case Success(result) =>
-              webHookActor ! WebHookActor.InvalidateWebhookCache
               complete(DeleteResult(result))
             case Failure(exception) => complete(StatusCodes.BadRequest, BadRequest(exception.getMessage))
           }
         }
+        webHookActor ! WebHookActor.InvalidateWebhookCache
         result
       }
     }
--- a/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
+++ b/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
@@ -40,12 +40,13 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
     pathEnd {
       entity(as[WebHook]) { hook =>
         DB localTx { session =>
-          HookPersistence.create(session, hook) match {
+          val result = HookPersistence.create(session, hook) match {
             case Success(result) =>
-              webHookActor ! WebHookActor.InvalidateWebhookCache
               complete(result)
             case Failure(exception) => complete(StatusCodes.BadRequest, BadRequest(exception.getMessage))
           }
+          webHookActor ! WebHookActor.InvalidateWebhookCache
+          result
         }
       }
     }
@@ -81,12 +82,13 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
       {
         entity(as[WebHook]) { hook =>
           DB localTx { session =>
-            HookPersistence.putById(session, id, hook) match {
+            val result = HookPersistence.putById(session, id, hook) match {
               case Success(result) =>
-                webHookActor ! WebHookActor.InvalidateWebhookCache
                 complete(result)
               case Failure(exception) => complete(StatusCodes.BadRequest, BadRequest(exception.getMessage))
             }
+            webHookActor ! WebHookActor.InvalidateWebhookCache
+            result
           }
         }
       }
--- a/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
+++ b/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
@@ -41,7 +41,9 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
       entity(as[WebHook]) { hook =>
         DB localTx { session =>
           HookPersistence.create(session, hook) match {
-            case Success(result)    => complete(result)
+            case Success(result) =>
+              webHookActor ! WebHookActor.InvalidateWebhookCache
+              complete(result)
             case Failure(exception) => complete(StatusCodes.BadRequest, BadRequest(exception.getMessage))
           }
         }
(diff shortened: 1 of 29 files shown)


magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
```diff
@@ -41,7 +41,9 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
       entity(as[WebHook]) { hook =>
         DB localTx { session =>
           HookPersistence.create(session, hook) match {
-            case Success(result)    => complete(result)
+            case Success(result) =>
+              webHookActor ! WebHookActor.InvalidateWebhookCache
```
kring: Isn't this a race condition?  The Invalidate message is sent before the transaction commits.  If the actor wakes up and receives the message before the commits, too, it won'ot see the new hook that was added.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/WebHookProcessor.scala
```diff
@@ -4,31 +4,26 @@ import akka.actor.ActorSystem
...
-    val (webHook, eventPage) = DB readOnly { implicit session =>
-      HookPersistence.getById(session, id) match {
-        case None => throw new RuntimeException(s"No WebHook with ID ${id} was found.")
-        case Some(webHook) => (webHook, EventPersistence.getEvents(session, webHook.lastEvent, None, Some(maxEvents), None, None, None))
-      }
-    }
+  def sendSomeNotificationsForOneWebHook(id: String, webHook: WebHook, eventPage: EventsPage): Future[WebHookProcessor.SendResult] = {
 
-    val events = if (webHook.isWaitingForResponse.getOrElse(false)) List() else eventPage.events
+    //    val events = if (!startup && webHook.isWaitingForResponse.getOrElse(false)) List() else eventPage.events
```
kring: nit: remove this commented-out code.

kring:
Other than those two comments (one trivial, the other serious but only if I'm not mistaken), this looks great to me!

There's a difference from previous behavior that is worth calling attention to, though. Previously, `aspects` and `optionalAspects` in the webhook definition would _only_ affect the records included in the payload. It would never change which events you received (assuming your webhook was set up to receive events, not just records).  But now, you won't receive, say, a record aspect changed event unless that aspect is mentioned in `aspects` or `optionalAspects`. I guess we can call that a feature. :)

magda-registry-api/src/test/scala/au/csiro/data61/magda/registry/WebHookProcessingSpec.scala
```diff
@@ -81,27 +87,39 @@ class WebHookProcessorSpec extends ApiSpec {
...
       }
 
-      val result2 = Await.result(processor.sendSomeNotificationsForOneWebHook("test"), 5 seconds)
-      result2.statusCode should be(Some(StatusCodes.OK))
+      Util.waitUntilDone(actor, "test")
 
       payloads.length shouldBe 1
       payloads(0).events.get.length shouldBe 1
-      payloads(0).records.get.length shouldBe 0
+      payloads(0).records.get.length shouldBe 1
```
kring: I'm not sure this test really makes sense anymore.  Or its description  is wrong or something.  Previously it did this:
1. Set up a webhook that only cares about aspect A.
2. Modify a record's aspect B.
3. Assert that we get an event but _not_ a record in the webhook payload.

In other words, it checks that the payload `includes records only for events modifying aspects that were requested`.

But with the changes in this PR, events are filtered by aspects, too.  So we won't get any events, and so we won't get any payload at all.  It could be useful to have a test for that, but that's not what this test is anymore, because the webhook is now subscribed to aspect B, too.

So maybe just delete the word "only" from this spec's description, i.e. "includes records for events modifying aspects that were requested."  Adding that "no payload generated" test mentioned above would be great, too, if it's not hard.

magda-registry-api/src/test/scala/au/csiro/data61/magda/registry/WebHookProcessingSpec.scala
```diff
@@ -81,27 +87,39 @@ class WebHookProcessorSpec extends ApiSpec {
...
       payloads(0).aspectDefinitions.get.length shouldBe 0
     }
   }
 
-  it("does not duplicate records or aspect definitions") { param =>
-    testWebHook(param, None) { payloads =>
-      val a = AspectDefinition("A", "A", Some(JsObject()))
+  it("does not duplicate records") { param =>
```
kring: The meaning of this test changed, too.  Previously it:
1. Did some things that generated multiple events impacting a single record.
2. Ran the webhook processing logic.
3. Asserted that the single payload contained the record only once, rather than (for example) once per event.

But now it expects multiple payloads and combines them all together and just asserts that there aren't any extraneous records listed in any of the payloads (which is pretty unlikely).

I believe I added this test based on an actual bug, so it's worth preserving. Maybe that can be done by setting the webhook inactive while generating the multiple events, and then setting it active again afterward?

magda-registry-api/src/test/scala/au/csiro/data61/magda/registry/WebHookProcessingSpec.scala
```diff
@@ -464,21 +678,22 @@ class WebHookProcessorSpec extends ApiSpec {
...
-      testAsyncWebHook(param, None) { payloads =>
+      testAsyncWebHook(param, None) { (payloads, actor) =>
         val aspectDefinition = AspectDefinition("testId", "testName", Some(JsObject()))
         param.asAdmin(Post("/v0/aspects", aspectDefinition)) ~> param.api.routes ~> check {
           status shouldEqual StatusCodes.OK
         }
 
-        val result1 = Await.result(processor.sendSomeNotificationsForOneWebHook("test"), 5 seconds)
-        result1.deferredResponse should be(true)
+        //        val result1 = Await.result(processor.sendSomeNotificationsForOneWebHook("test", false), 5 seconds)
```
kring: Remove commented-out code.

magda-sleuther-broken-link/src/index.ts
```diff
@@ -17,7 +17,7 @@ function sleuthBrokenLinks() {
         async: true,
         writeAspectDefs: [brokenLinkAspectDef, datasetQualityAspectDef],
         onRecordFound: (record, registry) =>
-            onRecordFound(record, registry, argv.retries)
+            onRecordFound(record, registry, /*argv.retries*/ 0)
```
kring: Did you mean to comment this out?

magda-sleuther-framework/src/index.ts
```diff
@@ -148,7 +148,7 @@ export default async function sleuther(
             }
         }).map((page: RecordsPage<Record>) => page.records);
 
-        await forEachAsync(registryPage, 20, (record: Record) =>
+        await forEachAsync(registryPage, 1, (record: Record) =>
```
kring: It can be a separate PR, but this seems like it should be a parameter to the `sleuther` function.  Some sleuthers will benefit from being able to process more than one record at a time.

magda-sleuther-framework/src/setupWebhookEndpoint.ts
```diff
@@ -13,20 +13,21 @@ export default function setupWebhookEndpoint(
...
 
-            const promises: Promise<
-                void
-            >[] = payload.records.map((record: Record) =>
-                options.onRecordFound(record, registry)
+            const promises: (() => Promise<void>)[] = payload.records.map(
+                (record: Record) => () =>
+                    options.onRecordFound(record, registry)
+            );
+            const megaPromise = promises.reduce(
```
kring: Consider putting the records in an `AsyncPage` (e.g. `AsyncPage.single(payload.records)`) and then use `forEachAsync` to allow a configurable number of simultaneous records here too.

kring:
Sorry, had a few more things.  I'm really done now. :)

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
AlexGilleran: Yeah that makes sense... I think it's being shielded by the big webhook actor only processing one thing at a time, but it should be after.

magda-sleuther-framework/src/index.ts
AlexGilleran: Agreed.

AlexGilleran:
@kring updated :)

PR #286, opened 2018-01-09
Summarize Sleuther

tristochief:
provides a machine readable, and human readable format field for easy manipulation and interpretation of the file formats of each distribution. Also provides a confidenceLevel to allow user to make code based on probabilities of correctness.

@AlexGilleran


magda-sleuther-format-layerer/src/format-engine/measures.ts
```diff
@@ -0,0 +1,196 @@
...
+
+        let processedFormat: string;
+
+        const transforms = [
+            this.identity,
+            (str: string) => this.lowerCase(this.removePeriod(this.replaceComma(this.replaceSpaceThenComma(this.replaceCommaThenSpace(str)))))
+        ];
+
+        for (const tf of transforms) {
+            processedFormat = this.applyTransform(format, tf);
```
AlexGilleran: Isn't this the same as `processedFormat = tf(format)` ?

magda-sleuther-format-layerer/src/format-engine/measures.ts
```diff
@@ -0,0 +1,196 @@
...
+        if(format === null || format === "") {
+            this.canDeduceFormat = false;
+            return null;
+        }
+
+        let processedFormat: string;
+
+        const transforms = [
+            this.identity,
+            (str: string) => this.lowerCase(this.removePeriod(this.replaceComma(this.replaceSpaceThenComma(this.replaceCommaThenSpace(str)))))
```
AlexGilleran: This looks like it's equivalent to `str.toLowerCase().replace(/\s*[.,]\s*/g, ' ')` ?

magda-sleuther-format-layerer/src/format-engine/measures.ts
AlexGilleran: Also this is stylistic, but maybe look into using `array.reduce` for this.

magda-sleuther-format-layerer/src/format-engine/measures.ts
```diff
@@ -0,0 +1,196 @@
...
+            processedFormat = this.applyTransform(format, tf);
+        }
+
+        // hard coded rules for separating out multiple formats when provided
+        let splitFormat: Array<string>;
+        let finalFormat: Array<string>;
+
+        // e.g: application/exe
+        splitFormat = processedFormat.split("/");
+        if (splitFormat.length > 2)
```
AlexGilleran: It's a bit easier to read if we put in the `{}`s around these if clauses :)

magda-sleuther-format-layerer/src/format-engine/measures.ts
```diff
@@ -0,0 +1,196 @@
...
+            throw new Error(
+                "a MIME type has more than 1 slash: " + processedFormat
+            );
+        if (splitFormat.length > 1) finalFormat[0] = splitFormat[1];
+        else {
+            // E.g. pdf & xlsx & doc & docx
+            splitFormat = processedFormat.split(" & ");
```
AlexGilleran: This might work better as something like `split(\/s*(&)/s*\)` in case the number of spaces varies or the separator is different.

Also isn't this pretty similar to what you're doing above with the commas, full stops etc? Might be easier to do it all in one pass.

magda-sleuther-format-layerer/src/format-engine/measures.ts
```diff
@@ -0,0 +1,196 @@
...
+            if(splitFormat.length > 1) finalFormat = splitFormat;
+            else {
+                splitFormat = processedFormat.split(" ");
+                if(splitFormat.length > 1) {
+                    // E.g. zip (xlsx)
+                    if(processedFormat.indexOf('(') > -1) {
+                        //TODO make this more efficient or elegant
+                        let length: number = (processedFormat.substr(processedFormat.indexOf(')')).length - processedFormat.substr(processedFormat.indexOf('(')).length);
+                        finalFormat[0] = processedFormat.substr(processedFormat.indexOf('('), length);
```
AlexGilleran: This is another thing that might be neater with a regex :)

magda-sleuther-format-layerer/src/onRecordFound.ts
```diff
@@ -0,0 +1,78 @@
...
+            extensionMeasure: new DownloadExtensionMeasure(distribution),
+            dcatMeasure: new DcatFormatMeasure(distribution)
+        };
+    });
+
+    const judgeResults: FormatSnapshotJudge[] = measureBundles.map(function(
+        measureBundle
+    ) {
+        return new FormatSnapshotJudge({
+            snapshots: [].concat(
```
AlexGilleran: Rather than `[].concat( ... )` you can just put all 3 inside of `[ ... ]` right?

magda-typescript-common/src/format/Measure.ts
```diff
@@ -0,0 +1,21 @@
...
+    SelectedFormats
+} from "./formats"
+import { Record } from "../../dist/generated/registry/api";
+
+//TODO turn this from a functional/OOP implementation to an OOP implementation (Make it a 'pumping' object)
+export abstract class Measure {
+    /**
+     * set to False if Measure isn't able to determine filetype with the data given in constructor; dcat field is null for example
+     */
+    public canDeduceFormat: boolean = false;
```
AlexGilleran: Could this be a function instead?

The way it is now, if someone looks at this before calling `getSelectedFormats()` first, it'll be wrong.

Another possibility would be to return it as part of the output from `getSelectedFormats()`

magda-sleuther-format-layerer/src/format-engine/FormatSnapshotEvaluator.ts
```diff
@@ -0,0 +1,58 @@
...
+    public getBestSnapshot(): SnapshotJudgeResult {
+        // turn array into an object:
+        // TODO change architecture to make this solution better
+        let snapshotDownload: Snapshot;
+        let snapshotExtension: Snapshot;
+        let snapshotDcat: Snapshot;
+
+        this.snapshots.snapshots.forEach(function(snapshot) {
+            if (snapshot.measure.canDeduceFormat) {
+                if (snapshot.measure instanceof DownloadMeasure)
```
AlexGilleran: Instead of using `instanceof` could we let each `Measure` implementation contain its own confidence level and return it from a `confidenceLevel()`function defined on the `Measure` interface?

AlexGilleran:
Might want to rename this PR too :)

magda-sleuther-format-layerer/src/format-engine/FormatSnapshotEvaluator.ts
tristochief: You stated inside the SnapshotJudge getBestSnapshot() function, that I should get the snapshot with the best confidence level, the problem is, the confidenceLevel inside all the snapshots were determined by each measure. The measure cannot determine accurately how confident it is within itself. The reason is hard to explain, but its because each measure is relying on an assumption that the method its using to determine the format is 100% correct, and it changes its confidence level depending on how much it could use its method. An example is the mimeType npm package. The Measure is assuming that, the package will return a 100% correct format, and will only decrease the accuracy if there is an error with the npm package. The Measure isn't self aware, and can figure out how accurate the npm package is, its assuming its 100% correct. Long story short, the confidenceLevels for each snapshot are relative (think of physics class, where you have different frames of references. If you go inside a rocket in space without windows, you cannot tell how fast its moving if it isn't accelerating). The SnapshotJudge's job is to go meta and determine the absolute confidenceLevel of each Measure. So its wrong to use confidenceLevel instead of instanceof to determine which format to use at this current stage of affairs.

magda-sleuther-format-layerer/src/onRecordFound.ts
tristochief: 4. Lastly you wrote to do [...] instead of [].concat. The problem is, I have 3 arrays and I want to concatenate them all into 1 array (flattened), if I put all 3 arrays inside [...], wont it return an array of arrays?

magda-typescript-common/src/format/Measure.ts
tristochief: 1. You suggested that the Measure class should just be a function. I agree with your suggestion, because Measures are meant to be immutable processors, rather than mutable objects. Right now, if I put all of the Measure functions into 1 file, 2 problems exist: 

---All private functions, that the measure function would use (like toLowerCase for the DcatMeasure) wouldn't be contained within anything, meaning it may be messy. Is this ok? 

---2ndly my SnapshotJudge Object's role is to work out which Measure is the most reliable. To work this out I'm storing the measure class inside a Snapshot interface. If I use a function, I wouldn't be able to differentiate       between different measures, because there's no instanceof for functions. I would have to use a more dodgy solution, like say store a string of the function name and use that. Is this good design practice?

2. Really, not only should the Measure be a function instead of a class, but the SnapshotJudge (Snapshot Evaluator) shouldn't be either, because they are just functional processors of information, like an AND gate is, not objects. Should I make the SnapshotJudge a function too?

magda-sleuther-format-layerer/src/onRecordFound.ts
AlexGilleran: oh right they're all arrays, no worries then :)

magda-typescript-common/src/format/Measure.ts
AlexGilleran: I actually meant just making `canDeduceFormat` a function!

-but-

It makes even more sense to make these immutable processors functions, that's a good idea.

In general (at least with my background, which is... java 🤦‍♂️ ) using `instanceof` to define behaviour isn't great OOP, because when you implement an interface then that implementation should be able to work everywhere the interface does. Wherever you use `instanceof`, your code has to have knowledge of the implementation rather than the interface. So I wouldn't be too attached to doing it that way :).

magda-sleuther-format-layerer/src/format-engine/formats.ts
```diff
@@ -0,0 +1,64 @@
...
+
+            throw new Error("Couldn't find an equivelant synonym for: " + rawFormat.toLowerCase());
+        } catch(message) {
+            throw new Error(message);
+        }
+    }
+}
+
+function getSynonymObject(): Object {
+    return fs.readFileSync(__dirname + "/synonyms.json").toJSON;
```
AlexGilleran: I've never heard of `toJSON` - is this a placeholder?

In any case you might consider using `require` instead so it resolves the filename easier and parses the JSON for you.

magda-sleuther-format-layerer/src/format-engine/measures/dcatFormatMeasure.ts
```diff
@@ -0,0 +1,116 @@
...
+    for (const tf of transforms) {
+        processedFormat = applyTransform(format, tf);
+    }
+
+    // hard coded rules for separating out multiple formats when provided
+    let splitFormat: Array<string> = [];
+    let finalFormat: Array<string> = [];
+
+    // e.g: application/exe
+    splitFormat = processedFormat.split("/");
```
AlexGilleran: dcat-distribution-strings actually has a `mediaType` field which the mime type is supposed to go in, might be worth looking in that too.

magda-sleuther-format-layerer/src/format-engine/measures/downloadExtensionMeasure.ts
```diff
@@ -0,0 +1,66 @@
...
+        [".*\\.(xlsx)(\\.zip)?$", "XLSX"],
+        [".*\\.(xls)(\\.zip)?$", "XLS"],
+        [".*\\.(tif)(\\.zip)?$", "TIFF"],
+        [".*\\.(zip)$", "ZIP"],
+        [".*\\.(html|xhtml|php|asp|aspx|jsp|htm)(\\?.*)?", "HTML"],
+        [".*\\/$", "HTML"]
+    ];
+
+    let formatsFromURL: Array<string> = [];
+    urlRegexes.some(function(regex) {
```
AlexGilleran: This looks like it could just use `urlRegexes.find` ?

magda-sleuther-format-layerer/src/onRecordFound.ts
```diff
@@ -0,0 +1,113 @@
...
+                format: formatResult.format.format,
+                confidenceLevel: formatResult.absConfidenceLevel
+            }
+        )
+    });
+
+    //TODO delete this piece of code
+    console.log("the best results gathered were: " + bestFormatResults.toString());
+
+    Promise.resolve();
```
AlexGilleran: Is there supposed to be a `return` here? In any case doesn't look like you need it.

magda-sleuther-format-layerer/src/format-engine/formats.ts
tristochief: hey alex do you mean fs.readFileSync(__dirname + "/synonyms.json").require()

magda-sleuther-format-layerer/src/format-engine/formats.ts
```diff
@@ -21,44 +21,44 @@ export enum Formats {
...
-            } else {
-                json = getSynonymObject();
-            }
-
-            for(var label in json) {
-                for(var i = 0; i < json[label].length; i++) {
-                    if(json[label][i].toLowerCase() === rawFormat.toLowerCase()) {
-                        return (<any>Formats)[label.toUpperCase()] || new Error("There is no " + label + " format in the Formats enum");
-                    }
+        let synonymObject: any;
```
AlexGilleran: Why redeclare this?

magda-sleuther-format-layerer/src/format-engine/measures/dcatFormatMeasure.ts
```diff
@@ -7,106 +7,103 @@ import { Record } from "@magda/typescript-common/dist/generated/registry/api";
...
-            } else {
-                console.log("f:" + finalFormat);
-                // can only deduce 1 format in this scenario
-                finalFormat.push(processedFormat);
-            }
-        }
-    }
+    const processedFormat: Array<string> = cleanUpAssemblyChain.reduce(function(
+        accumliation,
+        currentObject
```
AlexGilleran: This is a function not an object right? Can we name it `currentFunction`?

Also it's spelled `accumulation` ;).

magda-sleuther-format-layerer/src/format-engine/measures/dcatFormatMeasure.ts
```diff
@@ -7,106 +7,103 @@ import { Record } from "@magda/typescript-common/dist/generated/registry/api";
...
-                    finalFormat.push(splitFormat[0]);
-                }
-            } else {
-                console.log("f:" + finalFormat);
-                // can only deduce 1 format in this scenario
-                finalFormat.push(processedFormat);
-            }
-        }
-    }
+    const processedFormat: Array<string> = cleanUpAssemblyChain.reduce(function(
```
AlexGilleran: I really like this assembly chain idea in general but I think it'd be a bit clearer if you made it so that `cleanUpAssemblyChain` was all functions, and`format` was the last argument of reduce.

Like

```
cleanUpAssemblyChain.reduce(function(
    accumliation,
    currentObject
) {
    return currentObject(accumliation);
}, format)
```

Should still work the same but there won't be any mixed types in cleanUpAssemblyChain.

AlexGilleran:
@tristochief What happened here - weren't we going to remove the _summarize_ sleuther from this one so it just has the format sleuther stuff?

But looks like this is the other way around? 😕 

Are you going to put the format sleuther in another PR?

PR #337, opened 2018-01-11
Don't use the name of the SA2 that contains it as the name of SA1s

kring:
To reduce clutter in region search results.

To find SA1s, users will need to search for the region's ID, because SA1s don't have friendly names.


kring:
🦏

PR #378, opened 2018-01-16
Reduce CPU Usage When Processing Hooks

### LINKED ISSUE #370: Processing Webhooks Takes Too Much CPU In the Database
Need to optimise the query that selects events to process.

AlexGilleran:
Fixes #370 

### What this PR does
- Optimises the query used for getting new events for each webhook
- Stops the sleuther framework posting `success: false` as the result of an uncaught error - this has the effect of just replaying that set of events over and over so they generally fail, over and over.
- Makes the broken link sleuther handle nonsense URLs like "active" better.

### Checklist

### DIFF
--- a/magda-sleuther-framework/src/test/webhooks.spec.ts
+++ b/magda-sleuther-framework/src/test/webhooks.spec.ts
@@ -165,8 +165,10 @@ baseSpec(
                                                             options.id
                                                         }/ack`,
                                                         {
-                                                            succeeded:
-                                                                batch.overallSuccess,
+                                                            // We post succeeded: true even if we failed because in the event of
+                                                            // an uncaught error from the sleuther we want to move on, not get
+                                                            // stuck in a loop. "succeeded" would be better called "repeatWebhook"
+                                                            succeeded: true,
                                                             lastEventIdReceived: lastHookId
                                                         }
                                                     )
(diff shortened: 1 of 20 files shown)


magda-sleuther-framework/src/test/webhooks.spec.ts
```diff
@@ -165,8 +165,10 @@ baseSpec(
                                                             options.id
                                                         }/ack`,
                                                         {
-                                                            succeeded:
-                                                                batch.overallSuccess,
+                                                            // We post succeeded: true even if we failed because in the event of
+                                                            // an uncaught error from the sleuther we want to move on, not get
+                                                            // stuck in a loop. "succeeded" would be better called "repeatWebhook"
+                                                            succeeded: true,
```
kring: Hmm I'm not so sure about this. Sure, getting stuck in a crash loop is bad, but silently skipping this set of events is even worse. This gets back to what we were talking about earlier: we should be able reliably infer from a webhook's "last event" that a particular sleuther did everything it's supposed to do all the way up to a particular event. In the case of catastrophic and unexpected failure like this, it would be better (IMO) to set the webhook to inactive to stop the retries and let the user sort it out. We could potentially even stick details of the failure in the webhook table (though we should be careful that the webhook table doesn't become catch-all persistence for sleuthers).

kring:
Looks solid to me other than the one reservation.

magda-sleuther-framework/src/test/webhooks.spec.ts
AlexGilleran: Yeah good point, setting the hook to inactive if it fails is a good idea

PR #385, opened 2018-01-19
Issue/285: Private Auth API isn't authenticated

### LINKED ISSUE #285: Private Auth API isn't authenticated
Ironically.

Right now we're relying on the firewall, we should check user credentials too.

t83714:
Fixes #285 

### What this PR does

This PR mainly solves the issue described by issue #285.
Also, include some minor fixes on client side which may lead to blank screen when Auth API not return desired response. 

### Checklist


magda-typescript-common/src/authorization-api/ApiClient.ts
```diff
@@ -2,56 +2,81 @@ require("isomorphic-fetch");
...
+        }
+        this.requestInitOption={
+            headers: {
+                "X-Magda-Session" : this.jwt
+            }
+        }
+    }
+
+    getMergeRequestInitOption(extraOptions: RequestInit = null): RequestInit{
+        let options: RequestInit = this.requestInitOption ? {...this.requestInitOption} : {};
```
AlexGilleran: Could you use [lodash merge](https://lodash.com/docs/4.17.4#merge) for this?

magda-typescript-common/src/authorization-api/ApiClient.ts
```diff
@@ -2,56 +2,81 @@ require("isomorphic-fetch");
...
+                    "Content-Type": "application/json"
+                },
+                body: JSON.stringify(user)
+            }));
+            if(res.status >= 400) {
+                throw new Error(
+                    `Encountered error ${res.status} when POSTing new user to ${this.baseUrl}/private/users`
+                );
+            }
+            const resText = await res.text();
```
AlexGilleran: you can use `res.json()` to automatically parse the json :)

magda-web-client/src/actions/regionMappingActions.js
```diff
@@ -38,11 +38,13 @@ export function fetchRegionMapping() {
...
-        return dispatch(receiveRegionMapping(json));
-      } else{
-          return dispatch(requestRegionMappingError(json.error))
-      }
+        if(!json.error){
+          return dispatch(receiveRegionMapping(json));
+        } else{
+            return dispatch(requestRegionMappingError(json.error))
+        }
+      },(error)=>{
```
AlexGilleran: Nitpicking - I think it's a bit clearer to use `.catch` for errors on promise chains rather than the second argument to `then`

magda-authorization-api/src/createApiRouter.ts
```diff
@@ -33,6 +33,20 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+    router.all("/private/*", function(req, res, next) { //--- private API requires admin level access
+        
+        getUserIdHandling(req, res, options.jwtSecret, async (userId: string) => {
+            try{
+                const user = (await database.getUser(userId)).valueOrThrow(new Error(`Cannot locate user record by id: ${userId}`));
+                if(!user.isAdmin) throw new Error("Only admin users are authorised to access this API");
+                next();
+            }catch(e){
+                console.warn(e);
+                res.status(401).send("Not authorized");
```
AlexGilleran: Right now this will return `401` for both non-existent user and existent user who isn't an admin - it'd be great if we could make it `403 Forbidden` for non-admins!

t83714: Good idea~ Will make changes accordingly. I also noticed my changes impact `magda-admin-api` module. Will also make sure error codes are consistent in the test case tests of that modules as well.

magda-typescript-common/src/authorization-api/ApiClient.ts
t83714: No problem. Will use lodash for this one.

magda-typescript-common/src/authorization-api/ApiClient.ts
t83714: no problem. was for debugging purpose so that I can set a break point to check the actual body text. By the way, should we make AuthApi always return JSON response? At this moment, it return plain text upon error? If so, what should be the response structure? maybe a simple `{errorMessage:'xxxxxx'}`?

magda-web-client/src/actions/regionMappingActions.js
t83714: No problem. Will move the code to a seperate .catch.

t83714:
Realised the previous changes would impact `magda-admin-api` as well. Thus, pushed a few commits for the adjustments. Mainly, alter the logic of `mustBeAdmin` middleware and adjusted the unit test code as well.

magda-authorization-api/src/createApiRouter.ts
t83714: This has been done in commit 1e89cee6d87cc66b1f12c0a624e5215d822cad46

magda-typescript-common/src/authorization-api/ApiClient.ts
t83714: done in commit b21a0874c3ec4037e46f3b24dab6931a83f1979e

magda-typescript-common/src/authorization-api/ApiClient.ts
t83714: done in commit a903ef188d2ce9e411d1df3794c8ec6cecb73f59
Will gradually modify relevant code in future to make sure Auth API always returns JSON response

magda-web-client/src/actions/regionMappingActions.js
t83714: done in commit 2ea4414aa536606b9fa7630b2a2f83a4f95da850

t83714:
Added test cases for ApiClient.ts

t83714:
incorrect value in package.json of newer version `vega-canvas` will break webpack's build process:

https://github.com/vega/vega-canvas/blob/8acacccb82fd22460cf7b5c26e27b7c785ae5d0e/package.json#L38-L39

The previous commit was to solve this issue. Otherwise, travis will always fail to build `magda-web-client` module.

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+import * as request from "supertest";
+import * as express from "express";
+import * as yargs from "yargs";
+import addJwtSecretFromEnvVar from "@magda/typescript-common/dist/session/addJwtSecretFromEnvVar";
+import buildJwt from "@magda/typescript-common/dist/session/buildJwt";
+import createApiRouter from "../createApiRouter";
+import { expect } from "chai";
+import jsc from "@magda/typescript-common/dist/test/jsverify";
+import mockDatabase from "./mockDatabase";
+import mockUserDataStore from "@magda/typescript-common/dist/test/mockUserDataStore";
```
AlexGilleran: Because this has state and is a singleton, it's possible that one test will leave it a state that breaks the next one. Can you make it so that this is recreated or reset between individual tests?

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+            });
+        });
+    });
+
+    describe("GET /private/users/lookup", () => {
+        silenceErrorLogs(() => {
+            it("should return 401 status code if requested without sepecifying user ID", async () => {
+                await jsc.assert(
+                    jsc.forall(
+                        jsc.elements(newlyCreatedUsers),
```
AlexGilleran: So if this test got moved above `should create a new user if called as an admin user` this'd fail right?

I'd really strongly prefer if the unit tests didn't depend on each other. Could you use `userDataGen` (or `jsc.array(userDataGen)` when you need multiple users) to generate users for each spec and insert them into `mockUserDataStore` at the start?

magda-authorization-api/src/test/mockDatabase.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Maybe } from "tsmonad";
+import arrayToMaybe from "@magda/typescript-common/dist/util/arrayToMaybe";
+
+export default class MockDatabase {
+    getUser(id: string): Promise<Maybe<User>> {
+        return new Promise(function(resolve, reject) {
+            resolve(
+                arrayToMaybe(
+                    mockUserDataStore.getRecordByUserId(id).map(item => ({
+                        id: item.id,
```
AlexGilleran: Are you trying to clone the item here? You could use `item => ({...item})` I think?

magda-typescript-common/src/authorization-api/authMiddleware.ts
```diff
@@ -16,27 +16,30 @@ export const mustBeAdmin = (baseAuthUrl: string, jwtSecret: string) => (
...
 
     getUserIdHandling(req, res, jwtSecret, (userId: string) => {
-        apiClient.getUser(userId).then(maybeUser => {
-            maybeUser.caseOf({
-                just: user => {
-                    (req as any).user = user;
-                    if (user.isAdmin) {
+        const apiClient = new ApiClient(baseAuthUrl, jwtSecret, userId);
+
+        apiClient
```
AlexGilleran: I'd really prefer if we kept the explicit `isAdmin` check in here - I see how this works now, but having permission to access `/private/users/{id}` isn't necessarily the same as being an admin. If we changed it at some point in the future so that some users who aren't admins can still access `/private/users/{id}`, or if someone changed `getUser()` to use the `/public` API, then non-admin users would automatically get admin access to every other service that uses this middleware.

magda-typescript-common/src/test/mockAuthApiHost.ts
```diff
@@ -0,0 +1,180 @@
+import * as nock from "nock";
```
AlexGilleran: Appreciate the effort here, but I would've already been more than happy if the tests for `ApiClient` just checked that each method made the right HTTP calls!

What you've done here is great but a bit excessive :)

magda-authorization-api/src/test/arbitaries.ts
```diff
@@ -0,0 +1,55 @@
+import jsc from "@magda/typescript-common/dist/test/jsverify";
```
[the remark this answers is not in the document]
t83714: Thank you~ embarrassing typo indeed T_T Will correct it shortly.

magda-authorization-api/src/test/arbitaries.ts
```diff
@@ -0,0 +1,55 @@
...
+import {
+    peopleNameArb,
+    lcAlphaNumStringArbNe
+} from "@magda/typescript-common/dist/test/arbitraries";
+import { User } from "@magda/typescript-common/dist/authorization-api/model";
+
+export const emailArb: jsc.Arbitrary<string> = jsc
+    .tuple([
+        lcAlphaNumStringArbNe,
+        jsc.elements(["@"]),
```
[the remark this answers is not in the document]
t83714: I am still not very familiar with JSVerify. What I want to achieve is to generate some random email addresses (in valid format). Any good suggestion on this?

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
t83714: I thought it might be a more complete test if we test the whole loop --- i.e. first test create a user and the next test can still pull it out later. Thus, kept the state. But I think one thing you've mentioned could be right --- we should not make tests depends on each other. have to admit I still want to save some effort here on unit tests ;-)  Will re-structure the test later.

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+    let app: express.Express;
+    let argv: any;
+    let newlyCreatedUsers: User[] = [];
+
+    before(function() {
+        argv = retrieveArgv();
+        app = buildExpressApp();
+    });
+
+    function retrieveArgv() {
```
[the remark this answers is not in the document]
t83714: I think it might be a good idea as I also felt the similar code typed twice ;-)
Do you prefer it (and commonYargs.ts) to `typescript-common` or move to `magda-authorization-api` as different modules might requires different parameters?

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
t83714: Yes. As per the other commit, I think I will re-structure the tests.

magda-authorization-api/src/test/mockDatabase.ts
t83714: Notice your Database.ts return different set of fields from two different method (one method will skip field `sourceId`) --- thought you want to make sure different methods have different access and trying to simulate the same behaviour --- although I think I forgot to remove the `sourceId` from the second mocked method.

magda-typescript-common/src/authorization-api/authMiddleware.ts
t83714: I think you are right. When I was working on it, I also thought it might be necessary to consolidate / re-structure the authorization related function. The session validation & permission checking should be move into a bigger version of ApiClients.ts (the API might needs to be extends a little bit too).
Here I just want to keep minimal impact & changes to existing code, so that we can have this issue closed up and move further from there. 
How about I create another issue regarding this and work on this after this issue?

magda-typescript-common/src/test/mockAuthApiHost.ts
t83714: ;-) I have to say I regretted in the half way... but thought better to close it up.

magda-authorization-api/src/test/arbitaries.ts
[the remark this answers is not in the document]
AlexGilleran: The way you're doing emails is fine IMO, I was talking about this line only - `jsc.elements(["@"])` can also be written as `jsc.constant("@")` :).

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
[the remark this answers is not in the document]
AlexGilleran: I'm happy for `commonYargs` to stay where it is - just letting you know about `fakeArgv.ts` because it allows you to make an argv type from an object, so you don't have to include all the other stuff you don't need for tests like descriptions and so forth - like:

```typescript
const args: argv.Arguments = fakeArgv({
    listenPort: 6104,
    dbHost: "localhost"
    //etc
})
```

magda-typescript-common/src/authorization-api/authMiddleware.ts
AlexGilleran: Yeah that sounds like a good idea! But can you leave the `user.isAdmin` check in here just in case we don't get around to doing the bigger story for a while? :)

magda-typescript-common/src/test/mockAuthApiHost.ts
AlexGilleran: I'm very impressed with the thoroughness though 😄

magda-authorization-api/src/test/arbitaries.ts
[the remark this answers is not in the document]
t83714: O I see~ Now I got it! Thanks a lot! Will correct it~

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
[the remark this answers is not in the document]
t83714: Do think it's good to make `fakeArgv.ts` more generic like:
```
import { Arguments } from "yargs";

export default function fakeArgv(options: {
    internalUrl?: string;
    registryUrl?: string;
    jwtSecret?: string;
    userId?: string;
    listenPort?: number;
    dbHost?: string;
    dbPort?: string;
}): Arguments {
    return {
        ...options,
        $0: "",
        _: []
    };
}
``` 
And then put it to `magda-typescript-common`?
We might be able to re-use it from there in future.

magda-typescript-common/src/authorization-api/authMiddleware.ts
t83714: Have put the `isAdmin` checking back --- I think you are right --- It will be safer

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
t83714: this one has been done~

t83714:
@AlexGilleran  I've done the adjustment. Could you please have another look when you got a chance (No rush~)? Thank you~

magda-authorization-api/src/test/buildCreateApiRouter.spec.ts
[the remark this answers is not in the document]
AlexGilleran: Yep - I'm pretty sure you could just make it `fakeArgv(options: any)`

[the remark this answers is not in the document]
t83714: That's a good idea~ so we don't have to keep updating the type definition of `options`.
I've added new commit for this.

PR #387, opened 2018-01-23
Made connectors delete datasets no longer in their source

### LINKED ISSUE #383: Make connectors remove datasets that are no longer present in the source
Right now once a dataset gets created by a connector it's in magda forever.

We need to make it so that once a connector is finished doing a crawl of a source portal, it deletes everything in magda that wasn't there anymore.

We've been putting this off for a while but it's bitten us pretty seriously recently because we changed the format of dataset ids, meaning that now everything in dev is there twice. Better do this now or we might get bitten again somewhere else down the line.

AlexGilleran:
Fixes #383 

### What this PR does
- `JsonConnector` now accepts (or by default generates a UUID for) a `sourceTag`, which is passed as part of every record created when it's crawling.
- The database now stores this `sourcetag` in the `records` table as a nullable varchar
- The registry allows the sourcetag to be updated without creating a new event
- The registry now has a new `DELETE /records` endpoint that allows the specification of `sourceTag` and `sourceId` as query parameters. This deletes everything with a matching source id (as contained in the `source` aspect) that doesn't match the provided `sourceTag`.
- This endpoint returns the number of records deleted
- `JSONConnector` calls this endpoint at the end of every crawl.
- Yak Shave: I've put everything that returns a record summary behind `/summary`, because it really confused me that the same endpoint returns different data types depending on whether you request aspects or not - this also means that it doesn't match its swagger def.

#### Not included in this PR
- Right now the existing records have no `aspects.source.id` field so they won't get trimmed - I'll have to manually update those records with SQL for all the sources, matching on URL or name. 

### Checklist


magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsService.scala
```diff
@@ -97,6 +130,33 @@ class RecordsService(config: Config, webHookActor: ActorRef, authClient: AuthApi
     }
   }
 
+  @Path("/")
+  @ApiOperation(value = "Trim by source tag", notes = "Trims records with the provided source that DON'T have the supplied source tag",
+    nickname = "trimBySourceTag", httpMethod = "DELETE", response = classOf[MultipleDeleteResult])
+  @ApiImplicitParams(Array(
+    new ApiImplicitParam(name = "sourceTag", required = true, dataType = "string", paramType = "query", value = "Source tag of the records to PRESERVE."),
```
kring: Calling this `sourceTagToPreserve` would greatly reduce the chance of error, I think.

kring:
Looks good to me.  re: the new `/summary`, I wonder if we can change the summary format slightly so it's more similar to the normal record type.  e.g. instead of an array of aspect types, it could be an object with the value of each aspect set to null or something.  Or maybe we don't even need access to summaries at all?

AlexGilleran:
Yeah I actually went halfway to doing it that way but I figure that creates more confusion around the contents of the record - like if they get a record with a map of aspect ids set to null and `PUT` it back, should that actually set the aspect ids to null or should we be ignoring them?

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsService.scala
AlexGilleran: Yeah good call 👍

kring:
> if they get a record with a map of aspect ids set to null and PUT it back, should that actually set the aspect ids to null or should we be ignoring them?

Ignore them, just like we do when aspects are missing at the moment.  This can all be a separate PR, though, so I'm merging this.

PR #388, opened 2018-01-25
284 filter records by aspect contents

### LINKED ISSUE #284: Create a way to filter records by their contents from the HTTP API
- Parse in a path and value
- Look up duplicate URLs in a record

AlexGilleran:
Fixes #284 

### What this PR does
Introduces a _super-basic_ way to query for records based on their aspect contents. There's now an `aspectQuery` query param on `GET /records` in the registry, which you can pass a path and a value to in the format `aspectId.json.path:desiredValue` - this will URL decode and parse the query, then change it to a jsonb query in PostGres.

### Checklist


kring:
What no SPARQL? 😛

PR #393, opened 2018-01-31
Fix Travis Build issue

t83714:
### What this PR does

This PR solves recent travis build issue:
- Module `magda-admin-api` failed to build as incompatible version of `kubernetes-client`. The The issue was solved by fixing its version to `3.17.2`
- Module `magda-web-client` failed to build as recent release of `vega-canvas` introduced a [bug](https://github.com/vega/vega-canvas/blob/8acacccb82fd22460cf7b5c26e27b7c785ae5d0e/package.json#L38-L39) that will cause `webpack` fail to build.  The author has released a new version to fix this. Also, reset travis cache to eliminate issues on live travis system.

### Checklist


AlexGilleran:
Yay!

Issue #264, opened 2017-12-04
Some routes aren't rewritten to load react


soyarsauce:
Express isn't configured correctly for some routes?
e.g.
http://magda-dev.terria.io/page/data-sources
http://magda-dev.terria.io/page/search-syntax

Will result in:
Cannot GET /page/data-sources
Cannot GET /page/search-syntax

But are existing pages linked in the footer.

t83714:
fixed by PR 397

PR #397, opened 2018-01-31
Issue 368/264: Direct access to urls `pages/*` will lead to `Cannot GET pages/*` error

### LINKED ISSUE #368: About page and data sources doesn't open with new tab keyboard shortcut in chrome
To replicate, click on:
http://search.data.gov.au/page/about
http://search.data.gov.au/page/data-sources
In Version 63.0.3239.108 of Chrome, using Command + click. 

Would be good if someone replicated this. Hard to test on another computer at this time.

screen shot 2018-01-14 at 9 12 02 pm

t83714:
Fixes #368 #264 

### What this PR does

Fix issue that direct access to urls `pages/*` (e.g. http://search.data.gov.au/page/data-sources) will lead to `Cannot GET pages/*` error.

This issue may not be easy to replicate as:
1> Once user lands on home page, all content (including static pages) has been loaded via main js bundle. Any user clicks on static pages links (on screen) will not trigger HTTP requests.

2> Local Dev environment cannot replicate this issue as gateway will route all request to `http://localhost:6108` --- which is port of `magda-web-client` --- this will alway load index.html.
Production environment will router all requests to `http://web` --- which is the service name of `magda-web-server`. Thus, add route for `/pages/*` on `magda-web-server` will solve this issue.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-## 0.0.32
+## 0.0.33
 
 * Connectors now create organizations, datasets, and distributions with IDs prefixed by the type of record and by the ID of the connector. For example, `org-bom-Australian Bureau of Meteorology` is the organization with the ID `Australian Bureau of Meteorology` from the connector with ID `bom`. Other type prefixes are `ds` for dataset and `dist` for distribution. This change avoids conflicting IDs from different sources.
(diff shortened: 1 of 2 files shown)


t83714:
Oops! missed that~ Thanks for the pick-up & have corrected it.

PR #401, opened 2018-02-01
Format sleuther

### LINKED ISSUE #330: Make formats of distributions from database human readable, and more reliable
Formats in the DB right now are hit and miss, because the dcat-string's format property currently have conflicting, normalized, abbreviated and in some cases, completely incorrect format values. Allow magda developers to access non abbreviated, non normalized formats that are more reliable (more likely to be correct) than the dcat-string's format field.

-- PART I
- [ ] Create the synonym functionality so that we can group words that mean the same file format
- [ ] Create test cases and fix up measures for basic operation
- [ ] Create test cases and fix up measure evaluator (using a hierarchy for now)
- [ ] Run an integration test, to make sure a basic MVP runs without problems
-- PART II
- [ ] Create database infrastructure to store the measure's performance.
- [ ] label manually 100 records with their true file format
- [ ] Modify Measure Evaluator to use Bayesian Statistics to accurately determine the correctness probability of each of the different measures. (with the database infrastructure) and therefore choose the most correct format from all of the labels
- [ ] Modify Measure Evaluator to use Bayesian Statistics to accurately report the probability of the format being correct.
- [ ] OPTIONAL Modify format sleuther architecture to allow synonym functionality to be learnt rather than hard coded in.

tristochief:
Fixes #330

### What this PR does
Returns a distribution's format based on multiple different metadata and statistical methods. If more than 1 format was deduced, it assigns a probability to each format, telling it how likely it correctly identifies the format of the distribution.
@AlexGilleran


magda-sleuther-format/src/format-engine/measures/dcatFormatMeasure.ts
```diff
@@ -82,26 +90,24 @@ export default function getMeasureResult(
...
         filterBracketedFormats
     ];
 
-    const processedFormat: Array<string> = cleanUpAssemblyChain.reduce(function(
-        accumliation,
-        currentObject
-    ) {
-        return currentObject(accumliation);
-    });
+    //TODO ask gilleran the compiler doesn't let me compile this
```
AlexGilleran: Doesn't it?

magda-sleuther-format/src/onRecordFound.ts
```diff
@@ -35,7 +36,7 @@ export default async function onRecordFound(
     }
 
     //TODO delete this piece of code
```
AlexGilleran: Should this be deleted now it's getting merged?

magda-sleuther-format/src/onRecordFound.ts
```diff
@@ -19,9 +19,10 @@ import MeasureEvalResult from "./format-engine/MeasureEvalResult";
 let synonymObject = require("./format-engine/synonyms.json");
 
 //import { Snapshot } from "../../magda-typescript-common/src/format/MeasureSnapShot";
-let mochaObject = {
+export let mochaObject = {
```
AlexGilleran: If you use `nock` to mock the return to the registry instead of calling `recordFormatAspectTest` then you can get rid of this completely :).

AlexGilleran:
Oh also `d.ts` and `st.ts` have some compile issues.

PR #433, opened 2018-02-05
Issue/prettier commit hook

### LINKED ISSUE #399: Put prettier in a pre-commit hook
... so our typescript matches formatting.

t83714:
Fixes #399 

### What this PR does

Added prettier `pre-commit` hook to make sure consistent code style.

Used [pretty-quick](https://prettier.io/docs/en/precommit.html#option-2-pretty-quick-https-githubcom-azz-pretty-quick) to set up the pre-commit hook. `pretty-quick` will re-format the entire file. 

Also tested [precise-commits](https://prettier.io/docs/en/precommit.html#option-4-precise-commits-https-githubcom-jameshenry-precise-commits). It provides partial file formatting capability (Based on my tests, it seems it re-format a block code near changed lines). However, it could lead to partially formatted files based on my tests.

Added `prettier.config.js` to force tab length to 4. It seems default values are good for all other configuration options.

All available options are here:

https://prettier.io/docs/en/options.html

### Checklist


AlexGilleran:
Awesome!

Although is it possible to use prettier to just rip the band-aid off and format all our typescript in one go?

Otherwise we're gonna be dealing with diffs that change formatting for some time :(

t83714:
@AlexGilleran 
Do you mean `prettier` all source code files and include all changes in this PR?
Yes. I can do it.

AlexGilleran:
Yeah I think it'd be better to do that - rip the band aid off quick instead of slow ;).

t83714:
Used prettier to formatted all typescript files. Picked file in the following way (somehow glob parttern doens't not work very well):
```
find . -type f -name '*.ts' | grep -v "node_modules/" | grep -v "/dist/" | grep -v "/build/" | xargs prettier --write
```
To @AlexGilleran: Do we want to format all *.js source files as well?
We probably can pick source files use:
```
find . -type f -name '*.js' | grep -v "node_modules/" | grep -v "/dist/" | grep -v "/build/" | grep -v "/target/" | grep -v "wwwroot/" | grep -v "flow-typed/" | xargs prettier --write
```

AlexGilleran:
Nah good call not doing *.js yet, don't want to step on chloe's toes, or (especially) create a hell-merge when `mui-css` gets merged into master.

PR #438, opened 2018-02-06
Changes to deploy format + viz sleuther.

AlexGilleran:
### What this PR does
Makes a few changes to the format sleuther:
- Helm files to make it deployable on minikube/dev
- Stopped formats from being statically compiled... it works very similarly but a bit more optimistically, using the `synonyms.json` file as the record of what a good format is and allowing formats that aren't in it as long as they don't contain `"WWW:"`. This leaves it open for configuration from a dynamic config map later.
- Removed a few WWW: links from the synonyms, as only some of them are actually HTML - a lot of them are links to pages that allow the download of a non-HTML file.
- Made it so unknown WWW: link types aren't classified as ones that aren't obviously HTML are usually nonsense like FTP urls where we have no idea what the actual file type is

Also makes a few tweaks to the viz sleuther helm config so it can be deployed again.

### Checklist


kring:
Looks ok to me.  There are lots of conflicts that need resolving, though.

AlexGilleran:
@kring merged.

PR #450, opened 2018-02-07
Make indexer listen to delete record events and trim those records from the index

AlexGilleran:
### What this PR does
- Indexer now registers its webhook to include events and listen to DeleteRecord events
- Indexer now checks the events in webhook payloads for DeleteRecord events and deletes them from the index
- Indexer now updates its webhook on every start-up, even if already registered
- Minikube helm config default has registry authentication off

### Checklist


magda-indexer/src/main/scala/au/csiro/data61/magda/indexer/external/registry/RegisterWebhook.scala
```diff
@@ -58,18 +61,20 @@ object RegisterWebhook {
         EventType.DeleteRecord,
         EventType.CreateRecordAspect,
         EventType.PatchRecord,
-        EventType.DeleteRecordAspect),
+        EventType.DeleteRecordAspect,
+        EventType.DeleteRecord),
```
kring: `DeleteRecord` already exists, third from the top.

magda-indexer/src/main/scala/au/csiro/data61/magda/indexer/external/registry/WebhookApi.scala
```diff
@@ -32,23 +34,38 @@ class WebhookApi(indexer: SearchIndexer)(implicit system: ActorSystem, config: C
     magdaRoute {
       post {
         entity(as[WebHookPayload]) { payload =>
-          payload.records match {
-            // TODO: Handle registry config not found
-            case Some(records) =>
-              val dataSets = records.map(record => try {
+          val events = payload.events match {
```
kring: `val events = payload.events.getOrElse(List())`?

magda-indexer/src/main/scala/au/csiro/data61/magda/indexer/external/registry/WebhookApi.scala
```diff
@@ -32,23 +34,38 @@ class WebhookApi(indexer: SearchIndexer)(implicit system: ActorSystem, config: C
...
+            case None       => List()
+            case Some(list) => list
+          }
+
+          val idsToDelete = events.filter(_.eventType == EventType.DeleteRecord)
```
kring: What if the record is deleted and quickly recreated (i.e. by events in the same webhook payload)?

magda-indexer/src/main/scala/au/csiro/data61/magda/indexer/external/registry/RegisterWebhook.scala
AlexGilleran: Good point...

magda-indexer/src/main/scala/au/csiro/data61/magda/indexer/external/registry/WebhookApi.scala
AlexGilleran: I get paid per LOC right? Fixed 😳

magda-indexer/src/main/scala/au/csiro/data61/magda/indexer/external/registry/WebhookApi.scala
AlexGilleran: Good point. Wrote a test for this and somehow it works but I think that's just luck - have made the execution order explicit (delete everything then create everything).

AlexGilleran:
@kring addressed the review.

PR #480, opened 2018-02-15
Fixed some issues with deployment

AlexGilleran:
### What this PR does
- Due to a change that landed a while ago, the gateway expects a user id now so I've added that
- The prettier pre-commit hook doesn't necessarily work unless `prettier` and `pretty-fast` are in the root `node_modules`, so modified lerna config to do that
- Included both jacky's change and my change to make preview-map build properly - this is a really hard problem to confirm a fix of but this should do it.

### Checklist
- [x] CHANGES.md isn't relevant as this is just making stuff that we already recorded in CHANGES.md work more consistently
- [x] There's no issue in Zenhub

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,3 @@ snapshots
 .cache-tests
 node_modules/
 dist/
-package-lock.json
--- a/deploy/helm/magda/charts/gateway/templates/deployment.yaml
+++ b/deploy/helm/magda/charts/gateway/templates/deployment.yaml
@@ -52,6 +52,8 @@ spec:
             secretKeyRef:
               name: auth-secrets
               key: session-secret
+        - name: USER_ID
+          value: {{ .Values.global.defaultAdminUserId }}
 {{- if .Values.auth.facebookClientId }}
         - name: FACEBOOK_CLIENT_SECRET
           valueFrom:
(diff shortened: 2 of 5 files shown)


t83714:
Tested `magda-preview-module` build process in a local travis docker container (image: `travisci/ci-garnet:packer-1512502276-986baf0`), no errors found.

Verified preview module's basic functionality via container's 6110 port --- no visual issue (as what happened before the fix) sighted on UI.

Deployed `magda-gateway` in minikube. Pod running no issue.
`kubectl descrbe pod gateway-5c8f545bc9-xd4zj` showed `USER_ID` enviroment variable had been set:
```
Environment:
      NODE_ENV:                production
      JWT_SECRET:              <set to the key 'jwt-secret' in secret 'auth-secrets'>      Optional: false
      SESSION_SECRET:          <set to the key 'session-secret' in secret 'auth-secrets'>  Optional: false
      USER_ID:                 00000000-0000-4000-8000-000000000000
      FACEBOOK_CLIENT_SECRET:  <set to the key 'facebook-client-secret' in secret 'oauth-secrets'>  Optional: true
      GOOGLE_CLIENT_SECRET:    <set to the key 'google-client-secret' in secret 'oauth-secrets'>    Optional: true
```
Merged the pull request.

Issue #483, opened 2018-02-19
New Map Preview Component for New Dataset Page


t83714:
Requirements:
1> This component should be able to determine the best data source to use by itself and only shows one map preview for user
2> Parameter will be passed to TerriaJS (in the iframe) by `postMessage` rather than URL
3> Change the Map (style) to `Positron (Light)` 
4> Also required to hide the top menu bar
5> Add an `Open in National Map` button
6> Use broken link sleuther result to determine which format to be used as well. Onlly non broken link format will be considered.

t83714:
Done~ merged into master via pr https://github.com/TerriaJS/magda/pull/523

PR #484, opened 2018-02-20
Map previewer Component

t83714:
### What this PR does

Map Previewer Component for new dataset page.

Requirements see issue #483 

### Checklist


magda-web-client/src/UI/DataPreviewMap.js
```diff
@@ -1,18 +1,88 @@
...
+            No map preview available
+        </div>
+    );
+
+    const catalogItem = createCatalogItemFromDistribution(selectedDistribution);
+    const previewMapUrl = config.previewMapUrl + "#start=" + encodeURIComponent(JSON.stringify(catalogItem));
+    return (
+        <div className="data-preview-map">
+            <h3>Map Preview</h3>
+            <div className='mui-row'>
```
chloeleichen: Why do we need the `mui-row` and `mui-col-sm-12` div here?

magda-web-client/src/UI/DataPreviewMap.js
```diff
@@ -1,18 +1,88 @@
...
 import PropTypes from 'prop-types';
 import './DataPreviewMap.css';
+import { config } from '../config'
+
+const defaultDataSourcePreference = ['GeoJSON','WFS','WMS','csv-geo-au','KML'];
+
+function determineDistribution(props)
+{
+    const distributions = props.dataset.distributions;
+    let dataSourcePreference = props.dataset.dataSourcePreference;
```
chloeleichen: We can also use the broken link sleuther here to check if the link if broken first? Otherwise even if the distribution meets the requirement we still can't visualise it on map. The field is called `linkActive(boolean)` and
`linkStatusAvailable(boolean)`

magda-web-client/src/UI/DataPreviewMap.js
```diff
@@ -0,0 +1,153 @@
...
+
+        return (
+            <div className="data-preview-map">
+                <h3>Map Preview</h3>
+                <div className='mui-row'>
+                <div className='mui-col-sm-12'>
+                    <iframe 
+                        title={this.state.selectedDistribution.title}
+                        width="100%"
+                        height="600px"
```
chloeleichen: I think height should be 420px

t83714: @chloeleichen No problem. Will correct that~

magda-web-client/src/UI/DataPreviewMap.js
t83714: @chloeleichen Saw somewhere in your chart component --- copied here and thought it might help to keep looking consistent with yours. Do you think it's better to put it as:
```
<div className="data-preview-map">
                <h3>Map Preview</h3>
                    <iframe 
                        title={this.state.selectedDistribution.title}
                        width="100%"
                        height="420px"
                        frameBorder="0"
                        src={(config.previewMapUrl + "#mode=preview&hideExplorerPanel=1")}
                        ref={f=>this.iframeRef=f}
                    />
            </div>
```

chloeleichen:
Hi @t83714 , two more things:

We need to get rid of the following buttons which don't make sense in this context:
screen shot 2018-02-20 at 4 53 45 pm
We need to "go to National map" button over the map(can just use absolute positioning):
screen shot 2018-02-20 at 4 54 19 pm

magda-web-client/src/UI/DataPreviewMap.js
t83714: @chloeleichen 
I think it's a good idea.
Will add a new commit for this. Thanks for letting me know.

t83714:
@chloeleichen 
Yes. Probably added `Go to National Map` button outside the iframe. 
For removing those 3 buttons, do you know any magic `options` / `parameters` to turn them off before I look into TerriaJs's code?

chloeleichen:
@t83714 Not that I know, but maybe double check with @kring.

Possible alternatives:
1. use some javascript to modify the css in the iframe to hide those buttons (a bit dodgy)
2. Modify Terria to add params for controlling those buttons

t83714:
This has been included by https://github.com/TerriaJS/magda/pull/523
closed

PR #490, opened 2018-02-20
Testing prep

chloeleichen:
### What this PR does
Feedback form, banner
For https://github.com/TerriaJS/magda/issues/475


magda-web-client/src/Components/FeedbackForm.js
```diff
@@ -0,0 +1,120 @@
...
+    this.changeValue = this.changeValue.bind(this);
+    this.onSubmit = this.onSubmit.bind(this);
+    this.onDismissNotification = this.onDismissNotification.bind(this);
+  }
+
+  onSubmit(){
+    this.props.fetchFeedback(JSON.stringify({
+          name: this.state.name,
+          email: this.state.name,
+          feedback: this.state.feedback
```
t83714: It seems feedback API expects the data comes in with key `comment ` rather than ``feedback
See:
https://github.com/TerriaJS/magda/blob/35e83571d99f58f86d62ae3469914a9edc8130ec/magda-feedback-api/src/index.ts#L91

chloeleichen: Good spotting!

PR #495, opened 2018-02-22
ie 9 style bug

chloeleichen:
### What this PR does
Fixing
- [x] the data61 logo looked squashed in IE9
- [x] the search box background is grey is IE


magda-web-client/src/Components/Search/SearchBox.scss
```diff
@@ -43,6 +44,7 @@
     }
     img{
       width: 40px;
+      height: auto;
```
t83714: Tested on win7 / IE9
Search icon looked not in the correct w/h ratio:
image
Change to `height:40px` will make it look better:
image

t83714:
Considered we had a lot of relevant changes on https://github.com/TerriaJS/magda/pull/523
Will re-look at this PR once we merged that one

Issue #336, opened 2018-01-11
Tags: Merge sentence and title case tags


tkeuneman:
- Merge existing and future duplicate tags regardless of capitalisation
- Have all tags start with a capital letter
- Party like it's 1999

Generated from user input:
"a tag (Education and education)"
"Poor metadata entry: I understand that metadata entry and curation is the responsibility of the organisation or individual entering the data however poor practice is resulting in incomplete entries, unhelpful titling, spelling errors, etc.  This is making data.gov.au a dumping ground from which data cannot easily be extracted."

AlexGilleran:
I dunno if it's a bug per se, this is how the data comes into magda.

Fixing it would be a good application of a sleuther :)

tkeuneman:
Yeah, did you want me to relabel?

AlexGilleran:
I assume that means you're cool with it so I did it :)

t83714:
Will have a look at this one in sprint 3 to get better knowledge `sleuthers`

tkeuneman:
Pairing with @chloeleichen

AlexGilleran:
Update: We've decided that this would be much simpler to simply do on the front-end rather than using a sleuther.

t83714:
Had a chat with Tash and there three issues need to be addressed:
1. Duplicated tags with different cases should be merged:
e.g. tag `Oceans` & `oceans` should be merged into one tag `Oceans`
2. Tags contain possible separators (i.e. `,` `;` `|` or `/`)  will be split into shorter tags
e.g. tag `Oceans | Marine Biology | Fish` should be split into 3 tags `Oceans`, `Marine Biology` & `Fish`
3. Two tags should be merged into one if they contain same string after removing all possible `space`.
The shorter tag should be merged into longer tag.
example: dataset `ds-dga-13d68066-3e86-40f1-89f5-c75d9763f4b8`
tag `opencouncildata` should be merged with tag `open council data`.

The first two items will be addressed at front-end within the current sprint. 
Don't have a good solution for item 3. Will leave it to later.

PR #497, opened 2018-02-23
Issue/336 tag case

t83714:
### What this PR does

Fixes for the following problems (as described by my note of issue #336 ):

1> Duplicated tags with different cases should be merged:
e.g. tag Oceans & oceans should be merged into one tag Oceans

2> Tags contain possible separators (i.e. , ; | or /) will be split into shorter tags
e.g. tag Oceans | Marine Biology | Fish should be split into 3 tags Oceans, Marine Biology & Fish

### Checklist


magda-web-client/src/Components/Dataset/DatasetDetails.js
```diff
@@ -63,21 +74,14 @@ class DatasetDetails extends Component {
                         </div>
                     </div>
 
-                    <div className="record-details__sidebar mui-col-sm-4">
-                        <div className="tags">
+                    <div className='record-details__sidebar mui-col-sm-4'>
```
AlexGilleran: Can you run prettier over this file?

t83714: @AlexGilleran it seems prettier won't pick up the inconsistent use of  " or ' used for className.
Has manually changed ' to "

PR #498, opened 2018-02-23
Temporal filter wo temporal

chloeleichen:
### What this PR does

- [ ] When there's no temporal info, we hide the temporal filter

- [ ] Give the body a min height so the footer doesn't look so jumping when content is being loaded

### DIFF
--- a/magda-web-client/src/reducers/datasetSearchReducer.js
+++ b/magda-web-client/src/reducers/datasetSearchReducer.js
@@ -92,7 +92,7 @@ const datasetSearchReducer = (
                 data && data.dataSets && data.dataSets;
             let hitCount: number = data && data.hitCount && data.hitCount;
             let temporalRange: Array<Object> =
-                data && (data.temporal.start && data.temporal.end)
+                (data && data.temporal && data.temporal.start && data.temporal.end)
                     ? [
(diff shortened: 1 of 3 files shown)


magda-web-client/src/reducers/datasetSearchReducer.js
```diff
@@ -92,7 +92,7 @@ const datasetSearchReducer = (
                 data && data.dataSets && data.dataSets;
             let hitCount: number = data && data.hitCount && data.hitCount;
             let temporalRange: Array<Object> =
-                data && data.temporal
+                data && (data.temporal.start && data.temporal.end)
```
t83714: Although itl is promised that data.tempora must has a value, change 
```
data && (data.temporal.start && data.temporal.end)
```
to 
```
(data && data.temporal && data.temporal.start && data.temporal.end)
```
would be more robust and cover any future changes --- just in case

chloeleichen: yes you are right! thanks!

t83714:
Test with [my local URL](http://localhost:6108/search?publisher=Department%20of%20Environment%2C%20Land%2C%20Water%20%26%20Planning&q=*):
Works well:
image

PR #530, opened 2018-03-06
New dataset chart

chloeleichen:
Can test on this: http://localhost:6108/dataset/ds-dga-ae3386a4-2463-4aca-8fd0-ed067b586212/details horrible data anyway


t83714:
@chloeleichen 
Do you have another test link ( or Dataset ID)?
It seems this dataset has no preview data?

t83714:
Thank you @chloeleichen ~ it actually works well --- Just realized it requires the proxy inside the map preview module for loading data file ~

PR #540, opened 2018-03-12
Mobile header

chloeleichen:
### What this PR does
all itms on https://github.com/TerriaJS/magda/issues/501


t83714:
@chloeleichen Looks great! only can find a few minors things:
- DATA.GOV.AU heading, the `DATA` part should not be in black color
- Search Button Icon is not visible 
- Placeholder `Search for open data`; Letter S should be uppercase
image

chloeleichen:
@t83714 regarding the color of the word `data`, Tash is ok to have it all black since the title comes from the config file

t83714:
Tested on my local~ Works well~

PR #548, opened 2018-03-12
Fix double uri encoding

### LINKED ISSUE #430: Dataset URL double URL encoded for registry API fetch
[Dataset](http://magda.surge.sh/dataset/ds-launceston-http%3A%2F%2Fopendata.launceston.tas.gov.au%2Fdatasets%2Fe1c80176d3d44078bae02d6a9f27cc8b_0/details) is double URL encoded when calling registry api
* When you search for Speed Limits by City of Launceston you can’t open the dataset
* Registry fetch URL has double-encoded dataset id

steve9164:
### What this PR does

Fixes #430, and a bug with MagdaCatalogItem on the preview map that was hidden by bug #430.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## 0.0.36
+
+* Fixed a bug that stopped datasets with URL reserved characters in their id from being viewed.
+
 ## 0.0.35
 
 ## 0.0.34
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,6 @@
-## 0.0.36
+## 0.0.35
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+## 0.0.36
```
AlexGilleran: We're only up to 0.0.35!!

steve9164: Whoops. For some reason I thought 0.0.35 was skipped with the double release you did to prod. I think because I'm used to the update-version-number-on-release we use in TerriaJS and there are v0.0.35 commits everywhere.

PR #555, opened 2018-03-13
feedback tooltip error

chloeleichen:
Fixing https://github.com/TerriaJS/magda/issues/553


t83714:
@chloeleichen Thank you~

PR #564, opened 2018-03-13
issue 545 persist search text

chloeleichen:
### What this PR does
We use params `q= 'xxx'` to persist the search text in search page, dataset page and distribution pages. Reason being we already have the params `q`, it's more common and it's used by current dga


AlexGilleran:
Can you please update changes and connect this with its issue?

PR #596, opened 2018-03-15
Add CircleCI build for magda-web-client

maxious:
### What this PR does
Add a build/deploy workflow for CircleCI to send magda-web-client static files to cloud.gov.au eg. https://magda-web-client.apps.y.cld.gov.au/

I was doing a full "npm run build" including scala/sbt tasks but commented out to make this faster for just one component deploy (potentially you could deploy even more apps to different URLs from one build in parallel)

Also you could probably could deploy magda-web-server as a node app instead but consider this a PoC ;) Monkeypatching the fallback API url to https because all cloud.gov.au apps are HTTPS only.

Includes some eslint-ing which found an issue in some "CrappyChat" code. Fixed.
There's also a warning "Definition for rule 'jsx-a11y/href-no-hash' was not found  jsx-a11y/href-no-hash" due to some out of date dependency https://github.com/facebook/create-react-app/pull/2930

### Checklist
- [x] No issue from Zenhub/sprint

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,82 @@
+version: 2
+jobs:
+  build_web:
+    docker:
+      #- image: circleci/openjdk:8-jdk-node-browsers
+      - image: circleci/node:9
+    steps:
+      - checkout
+#      - run:
+#          name: install sbt
+#          command: sudo apt-get install apt-transport-https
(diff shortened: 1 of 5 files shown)


AlexGilleran:
Super cool - will work on making https://magda-dev.terria.io work :)

PR #597, opened 2018-03-15
Fix redirect to blank/outdated datasetId

### LINKED ISSUE #572: All datasets resolve to 404s on dev on first click
This problem only happens to the first click after refresh page.

### LINKED ISSUE #589: Distribution page doesn't load correct data
It seems more than one requests will be sent out.
May have something to do with links.

steve9164:
### What this PR does
Fixes #572
Fixes #589 

### Checklist
- [ ] ~I've updated CHANGES.md with what I changed.~ not really applicable (was only broken for a couple of days)

### DIFF
--- a/magda-web-client/src/Components/RecordHandler.js
+++ b/magda-web-client/src/Components/RecordHandler.js
@@ -94,10 +94,10 @@ class RecordHandler extends React.Component {
                     { id: "preview", name: "Preview", isActive: true }
                 ];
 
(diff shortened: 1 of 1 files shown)


steve9164:
I've used `encodeURI` here with the thought that `encodeURI . decodeURI == id`. Sounds reasonable to me, but I'm not 100% sure that's correct.

PR #655, opened 2018-03-23
Removed query language from search "by, in, from" etc.

AlexGilleran:
### What this PR does
- Removes the almost-natural-language search, replaces it with nice clean query parameters
- Converts all the old tests over to the new interface
- Converts the UI to the new interface
- Re-enables the lucene query syntax on search queries.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,115 +1,116 @@
 ## 0.0.37
 
-* Update email address for feedback including in map preview errors
-* Make facets stack on mobile
-* Use params q= 'xxx' to persist the search text in search page, dataset page and distribution pages
-* Added Download button/link click tracking via Google Analytics
-* Add CircleCI configuration to automatically build and deploy the public web interface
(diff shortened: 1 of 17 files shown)


CHANGES.md
```diff
@@ -1,112 +1,113 @@
 ## 0.0.37
```
AlexGilleran: So it turns out the latest version of prettier does markdown, which is cool, but also means that every single line of this has been cleaned up.

magda-web-client/package.json
```diff
@@ -113,5 +113,8 @@
         "docker": {
             "include": "node_modules build"
         }
+    },
+    "dependencies": {
+        "lodash.flatten": "^4.4.0"
```
kring: Should go in `devDependencies`.

kring:
Just the one minor comment, so I'll fix that and merge it.

PR #658, opened 2018-03-25
Adjust IE Polyfills and Add basic API proxy for testing without CORS problems 

maxious:
### What this PR does
There was some es5-shim but IE9 needs more than that and available for the third party JS.
Also in testing IE without CORS problems, made a hacky proxy that isn't really the gateway but close enough 

### Checklist


magda-web-server/src/index.ts
```diff
@@ -97,7 +103,9 @@ app.use(
                 "platform.twitter.com",
                 "www.googletagmanager.com",
                 "www.google-analytics.com",
-                "rum-static.pingdom.net"
+                "rum-static.pingdom.net",
+                "cdnjs.cloudflare.com", //rollbar and shims
+                "wzrd.in" // es7 shim
```
AlexGilleran: Although admittedly we're allowing unsafe-inline at the moment so it's the wild west, it'd be cool if we could be a bit more specific with these last two seeing how any javascript can be allowed on them. According to [this](https://stackoverflow.com/questions/33680150/does-the-content-security-policy-standard-support-wildcard-paths-if-not-why-do) you can put directories into the CSP rather than just domains - can we do `cdnjs.cloudflare.com/ajax/libs/es5-shim/` etc instead?

t83714:
According to [here](https://reactjs.org/docs/javascript-environment-requirements.html), it seems React 16 depends on `requestAnimationFrame`?
Do we need `raf` package to shim `requestAnimationFrame`?

maxious:
Raf added now

PR #677, opened 2018-03-26
New homepage

t83714:
### What this PR does

Implement the new homepage

### Checklist


magda-web-client/src/homepage-config.js
```diff
@@ -0,0 +1,34 @@
...
+        "twenty-1080w.jpg",
+        "twenty-2160w.jpg",
+        "twenty-2880w.jpg"
+    ],
+    tagLineTextDesktop: "An easy way to find, access and reuse public data",
+    tagLineTextMobile: "Find, share, and reuse open government data",
+    Lozenge: {
+        icon: "camara.svg",
+        text: "Explore QLD road data",
+        url: "https://search.data.gov.au/search?q=QLD%20road%20data"
```
AlexGilleran: "Queensland road data" will work way better than "QLD road data"

t83714: I think the text is the tiny button on the homepage. 
image
I guess use abbreviation is mainly for control its size --- correct me if I am wrong @tkeuneman 
This config file is included in repo only for reminding us its structure in future.
The actual copy is located at: https://s3-ap-southeast-2.amazonaws.com/data-gov-au-frontpage/development/homepage-config.js
(staging & production site will load file in `staging` & `production` folder).
Most changes to homepage later can be done via modifying this config on s3.

AlexGilleran: We can leave the text as "QLD", just change the text in the URL to "queensland" :).

I don't mind if it's in S3 or here.

t83714: Oh I see. I thought you want to change the button text.
Will change the URL now.

PR #753, opened 2018-04-04
Issue/672 visualisation tweaks

adamzerella:
### What this PR does
- Data Preview table vertical height is fixed to 10 entries, vertical scroll to see other rows.
- Had a hard time trying to find an elegant way to hide Data Preview `Table` and `Chart`, might need to discuss the filthiness of what I did... but it works!

### Checklist


magda-web-client/src/UI/DataPreviewVis.js
```diff
@@ -151,7 +151,8 @@ class DataPreviewVis extends Component {
             if (
                 previewData &&
```
AlexGilleran: So what was the problem you had?

It'd be cool if it could be more like

```
if (previewData.meta.chartFields) {
   <Chart Tab>
}
if (previewData.meta.fields) {
   <Table Tab>
}
```

is it the `defaultSelectedIndex` that's the problem?

adamzerella:
@AlexGilleran Wasn't sure if this was really necessary/allowed considering we might not add additional tabs but I think I've made the workflow a bit simpler. Let me know if this type of refactoring is useful and I might look at doing it again for other issues.

Also got rid of a couple of unrelated console warnings for incorrect `PropType` vs `propType`.

AlexGilleran:
@adamzerella really like what you've done, just code-golfed one of the functions a bit - take a look and feel free to merge if you're happy with it :).

PR #761, opened 2018-04-06
Issue/582 filters disappearing

sukhrajghuman:
### What this PR does
Everytime a search was performed and action was dispatched which reset the state of the filters to initial null state. I've stopped this action from dispatching.

### Checklist

### DIFF
--- a/magda-web-client/src/actions/datasetSearchActions.js
+++ b/magda-web-client/src/actions/datasetSearchActions.js
@@ -44,7 +44,6 @@ export function resetDatasetSearch(): SearchAction {
 export function fetchSearchResults(query: string): Store {
     return (dispatch: Dispatch) => {
         let url: string = config.searchApiUrl + `datasets?${query}`;
-        dispatch(requestResults(query));
         return fetch(url)
             .then((response: Object) => {
                 if (response.status === 200) {
--- a/magda-web-client/src/reducers/datasetSearchReducer.js
+++ b/magda-web-client/src/reducers/datasetSearchReducer.js
@@ -45,20 +45,6 @@ const datasetSearchReducer = (
     action: SearchAction
 ) => {
     switch (action.type) {
-        case "REQUEST_RESULTS":
-            return Object.assign({}, state, {
-                isFetching: true,
-                error: null,
-                apiQuery: action.apiQuery && action.apiQuery,
-                temporalRange: initialData.temporalRange,
-                publisherOptions: initialData.publisherOptions,
-                formatOptions: initialData.formatOptions,
-                activePublishers: initialData.activePublishers,
-                activeRegion: initialData.activeRegion,
-                activeDateFrom: initialData.activeDateFrom,
-                activeDateTo: initialData.activeDateTo,
-                activeFormats: initialData.activeFormats
-            });
         case "FETCH_ERROR":
             return Object.assign({}, state, {
                 isFetching: false,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 *   Removed excess vertical whitespace from hamburger menu
 *   Dataset page: Change icon for distribution page
 *   Empty search won't be saved as recent search item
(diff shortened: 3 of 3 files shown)


magda-web-client/src/actions/datasetSearchActions.js
```diff
@@ -44,14 +48,17 @@ export function resetDatasetSearch(): SearchAction {
 export function fetchSearchResults(query: string): Store {
     return (dispatch: Dispatch) => {
         let url: string = config.searchApiUrl + `datasets?${query}`;
-        dispatch(requestResults(query));
```
AlexGilleran: (comment 1/3)

You want to leave this here, _but_ you want to have it pass the `Query` object instead of the string.

So you need to change `fetchSearchResultsIfNeeded` to pass through the query object

```javascript
export function fetchSearchResultsIfNeeded(urlQueryObject: Object): Store {
     const apiQuery = parseQuery(urlQueryObject);
     return (dispatch, getState) => {
         if (shouldFetchSearchResults(getState(), urlQueryObject.q, apiQuery)) {
             return dispatch(fetchSearchResults(apiQuery, urlQueryObject));
         }
     };
 }
```

So now this function can be 
```javascript
 export function fetchSearchResults(query: string, queryObject: Object): Store {
     return (dispatch: Dispatch) => {
         let url: string = config.searchApiUrl + `datasets?${query}`;
        dispatch(requestResults(queryObject));
```

magda-web-client/src/actions/datasetSearchActions.js
```diff
@@ -10,10 +10,14 @@ import type {
     SearchAction
 } from "../helpers/datasetSearch";
 
-export function requestResults(apiQuery: string): SearchAction {
+export function requestResults(
+    apiQuery: string,
+    json: DataSearchJson
```
AlexGilleran: (comment 2/3, read the next one first!)

... so now you can change this to

```javascript
export function requestResults(
    query: Object
): SearchAction {
```

Accepting the query object, then pass that on below instead of `{apiQuery, json}`

magda-web-client/src/reducers/datasetSearchReducer.js
```diff
@@ -46,13 +46,17 @@ const datasetSearchReducer = (
...
         case "REQUEST_RESULTS":
+            console.log(action.json);
             return Object.assign({}, state, {
                 isFetching: true,
                 error: null,
                 apiQuery: action.apiQuery && action.apiQuery,
-                temporalRange: initialData.temporalRange,
-                publisherOptions: initialData.publisherOptions,
-                formatOptions: initialData.formatOptions,
+                temporalRange: [
```
AlexGilleran: (comment 3/3)

... so then here, `activePublishers`, `activeRegion` etc can be set to the values in the query object that you've passed through as `action.query`.

There's some wrinkles here - I'm not sure whether you still need the `apiQuery` string (what does it do? :/) and you might need to smooth over what happens with the filters if you've come to the search page for the first time and they don't have options yet but do have active values.

sukhrajghuman:
Okay ignore my latest branch. Will have a look into it after lunch. Cheers

magda-web-client/src/reducers/datasetSearchReducer.js
```diff
@@ -46,18 +46,31 @@ const datasetSearchReducer = (
 ) => {
     switch (action.type) {
         case "REQUEST_RESULTS":
+            let queryObject = action.queryObject;
+            console.log(action.apiQuery);
```
AlexGilleran: Need to remove this before merging :)

sukhrajghuman: Yea i realised that when you select a single publisher, there is no array of publishers, just a string. And in FacetHeader.js it is rendering activeOptions.length. And in the case where there is 1 publisher selected it is just rendering the length of text. So when you're selecting Geoscience Australia it is rendering 48 as in 48 characters of string 😅 .

PR #771, opened 2018-04-09
Issue568 home fix

chloeleichen:
### Checklist

### DIFF
--- a/magda-web-client/src/Components/Search/SearchBox.js
+++ b/magda-web-client/src/Components/Search/SearchBox.js
@@ -179,7 +179,8 @@ class SearchBox extends Component {
                     <button
                         onClick={this.onClickSearch}
                         className="search-btn"
-                        type="button">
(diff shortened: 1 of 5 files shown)


t83714:
Changes checklist:

- [ ] No line (or in this case a dot) should appear in the underline until a user starts typing
- [ ] Show down transition for news just a tad (maybe 1/2 a second more?)
- [ ] Line needs to be longer and underneath search icon. See Figma for more info
- [ ] The right end of the lozenge needs to be aligned with the news. See Figma for more info
- [ ] There should be 16px gaps between these columns
- [ ] Mobile tag line should be: An easy way to find, access and reuse public data (sorry about this one!)
- [ ] just a heads up that the news card and mobile search field shadows have been updated for consistency across the whole site. Also, at this stage all cards and the mobile search field have a 4px corner radius.

PR #776, opened 2018-04-09
501mobile qa

chloeleichen:
### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 *   Empty search won't be saved as recent search item
 *   Fixed an issue that accessURL won't be displayed on distribution page.
 *   Will not show distribution page if only accessURL is available.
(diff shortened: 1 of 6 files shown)


chloeleichen:
@t83714 Done! changed the interaction because the mobile nav was obscuring the search box on mobile

t83714:
My test check list:
- [ ] Capital 'S' in Search for open data
- [ ] Should only show four lines of text before 'full description is prompted (400 characters on mobile view)
- [ ] 'Show full description' is always highlighted when I go to any dataset page, and I'm not sure why
- [ ] Data quality underline should end after the word, not include the ':' and the stars
- [ ] There should be lines underneath each file type. See mockup for more information

t83714:
Good Job~ Thanks for the fix on mobile menu~

Issue #580, opened 2018-03-15
Omit 'unspecified' from the publishers filter


tkeuneman:

AlexGilleran:
@tkeuneman should it also be omitted from Format?

AlexGilleran:
In general this would be a nice introduction to the search api for someone who was keen to work on it :)

tkeuneman:
Lose 'unspecified' from all the filters, but still keep them in the results until a user starts filtering...hope that makes sense.
In other words, if I have no file filters applied, I can still see unspecified file types in the results. If I only filter by 'csvs', the 'unspecified' files are filtered out.

PR #778, opened 2018-04-10
Issues/580 Remove 'unspecified' item from publisher & format aggregation

t83714:
### What this PR does

Remove 'unspecified' item from publisher & format aggregation #580 

### Checklist
- [x] Modified test cases to suit new logic.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 *   Empty search won't be saved as recent search item
 *   Fixed an issue that accessURL won't be displayed on distribution page.
 *   Will not show distribution page if only accessURL is available.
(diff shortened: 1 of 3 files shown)


t83714:
@AlexGilleran  
Modified the test cases to suit the new logic.
Hope I did the correct thing

AlexGilleran:
@t83714 Good stuff, implementation is right but we shouldn't need to keep checking for Unspecified in the facet tests so might as well get rid of it altogether :). I've changed the code (otherwise I wouldn't have been sure it was right!), have a look and merge if you like :).

t83714:
Merged! Thanks a lot~

Issue #749, opened 2018-03-31
Performance Optimisation for Registry API


AlexGilleran:
A whole performance optimisation on registry API, to also solve the below problem.

It's possible for the Registry API to max out the CPU while doing very little
See this incident (if you can): https://app.google.stackdriver.com/incidents/0.kqbhybqnvvgj

Essentially the registry api was using its node's entire CPU - it was still responsible enough to survive livesness checks but slow enough that further-away stackdriver checks started to time out.

Restarting the pod fixed it.

Not sure what caused it or what the best mitigation is - perhaps more stringent liveness checks for now?

t83714:
@AlexGilleran 
It seems change MAX_EVENTS in this line:
https://github.com/TerriaJS/magda/blob/79e9340fcd48d94bd3498de533f66ee785bf1061/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/WebHookActor.scala#L113
to 10 solve most my CPU usage problem locally.
Do you think it could be a potential fix?

t83714:
Good idea! will do~

aneesha09:
Do 8 points worth of optimisation :D

mwu2018:
A quick search found that blocking code (Await) is used extensively, which should be avoid as it will exhaust resources.

PR #801, opened 2018-04-16
Fixed bad values yml for SDGA deploy.

AlexGilleran:
### What this PR does
Some of the resource requests in the search.data.gov.au values file in helm were formatted wrong - in particular this mean the registry api had no resource requests at all. This remedies that, and removes the config for database nodes which are no longer used in the sdga deploy.

### Checklist
- [x] No changes.md, this isn't part of the underlying software.
- [x] No issue

### DIFF
--- a/deploy/helm/magda/charts/registry-api/templates/deployment.yaml
+++ b/deploy/helm/magda/charts/registry-api/templates/deployment.yaml
@@ -50,7 +50,7 @@ spec:
             port: 80
           initialDelaySeconds: 60
           periodSeconds: 10
-          timeoutSeconds: 3
+          timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds | default 10 }}
         ports:
         - containerPort: 80
         resources:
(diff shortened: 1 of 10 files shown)


deploy/helm/magda/charts/registry-api/templates/deployment.yaml
```diff
@@ -50,7 +50,7 @@ spec:
             port: 80
           initialDelaySeconds: 60
           periodSeconds: 10
-          timeoutSeconds: 10
+          timeoutSeconds: 3
```
t83714: What's the reason we lower the timeout value?
Asked because of I quite often have the situation (in my local minikube) that registry API is killed due to high CPU usage. And the killing & re-creation action itself add more burden to the node and causing other pods starting response slower (which may trigger more pods being killed).

AlexGilleran: Yeah good question. The answer is that it mitigates #749 - right now sometimes the registry gets into a state where it uses every CPU cycle it can but doesn't actually process any webhooks... it still responds to API requests within 10 seconds, but when these requests com from outside the network they end up being 15 seconds + :(.

So for now I've set a really tight `timeoutSeconds` value to ensure it's rebooted as soon as it slows down.

In the gitlab PR I disabled liveness probes except in prod but you have a good point in that this is going to make life miserable on minikube until that gets merged... I'll make this prod only.

PR #804, opened 2018-04-17
Hide asterisk * in searchbox

maxious:
### What this PR does
Hide the * when I click on the Dataset link in header or click through from a Publisher

### Checklist


AlexGilleran:
This means that you can't actually type "*" as a search term, which a problem because it's a legitimate search - I use it all the time to demo how many datasets are searchable, for instance.

Maybe we could make it so that the Datasets link passes some kind of `searchAll=true` parameter, or make it so that the `/search` endpoint searches for `*` by default while still displaying "" as the query?

maxious:
... Or do this just when the asterisk comes via the query string. This also still works correctly for recent/saved searches and you can type * if you wish.

AlexGilleran:
As discussed on Gitter - this is nearly fine but it's a bit weird typing in an `*` and then having it disappear when you hit "Enter".

PR #806, opened 2018-04-17
Issue792 mobile filter

chloeleichen:
### What this PR does

Implement new design for the mobile filter https://invis.io/QUGT8RNBYJ3


AlexGilleran:
Looks pretty good, some comments:


- Should this say "FILTERS" and have a grey background? Also the little  ⬇️ caret looks like it should have more padding from the right.
image
(design)
image


- The "X" looks like it's not centered:
image

- There's a big gap between "RESULTS" and the first result - it's the same as what it is for desktop but it looks way bigger on mobile.
image

- No CHANGES.md

AlexGilleran:
Everything looks good except now there's a big FILTERS and a small FILTER:

image

Shouldn't it just be the one at the bottom, but with "FILTERS" not "FILTER"?

PR #808, opened 2018-04-17
Issue/396 index format

t83714:
### What this PR does

- Made indexer index format information from format sleuther
- 

### Checklist
- [ ] There are unit tests to verify my changes are correct | Unit tests aren't applicable (delete one)
- [ ] I've updated CHANGES.md with what I changed.


magda-indexer/src/main/scala/au/csiro/data61/magda/indexer/crawler/RegistryCrawler.scala
```diff
@@ -25,7 +25,13 @@ class RegistryCrawler(interface: RegistryExternalInterface, indexer: SearchIndex
 
   var lastCrawl: Option[Future[Unit]] = None
 
-  def crawlInProgress: Boolean = lastCrawl.map(!_.isCompleted).getOrElse(false)
+  def crawlInProgress: Boolean = lastCrawl.map(!_.isCompleted).getOrElse(false) match {
+    case false => 
+        lastCrawl = None
```
AlexGilleran: I think it'd be better to fix this bug in crawl() - instead of matching on `lastCrawl` do something like

```scala
   def crawl(): Future[Unit] = {
     if (crawlInProgress) lastCrawl.get else {
         lastCrawl = Some(newCrawl())
     }
   }
```

t83714: I think that is a better idea~
Have updated the code~ 
Thank you!

Issue #173, opened 2017-07-28
Useless record/aspect PATCHes get written to the Events table


kring:
If you do a PUT, and the new record is the same as the existing one, no event will be written to the Events table.  But if you invoke a PATCH service, it will write that patch as an Event without considering whether that patch actually changes anything.

PR #820, opened 2018-04-20
Issue/173 Useless patch should not trigger any event creation

t83714:
### What this PR does

Useless patch should not trigger any event creation

See #173 

- There are three sets of PUT/PATCH APIs. 
  - One is part of record services: [PUT/PATCH] /records/{id}
  - One is part of record aspect services: [PUT/PATCH] /records/{recordId}/aspects/{aspectId}
  - One is part of aspect services: [PUT/PATCH] /aspects/{id}
- /records/{id} endpoint (no matter is PUT or PATCH) will ONLY create event when record name field is changed. There should never be any unnecessary events created. 
  - Both PUT/POST endpoint will actually ignore any aspects modification attempts
  - Thus, no fix is required
- PATCH /records/{recordId}/aspects/{aspectId} requires fix
- PATCH /aspects/{id} requires fix as well

### Checklist


AlexGilleran:
👏

PR #823, opened 2018-04-23
Stopped duplicated feedback

sukhrajghuman:
I can’t send feedbacks from localhost, so it’s hard to test this out.
But I am pretty sure this will solve the issue.

### What this PR does
Currently once you press the feedback ‘send’ button, it is not
disabled. I can click it multiple times, making duplicate requests.

This will make sure the button is disabled in the ‘sending feedback’
state.

### Checklist


AlexGilleran:
I'm still able to leave duplicate feedbacks while the initial report is sending - actually I never see the button enter a "Sending..." state at all, it just stays the same until the thank you message comes up. I think there's a bug in feeding that state through that'll need to be fixed for this to work.

The idea in general is sound though 👍. You might want to make it so that the `onClick` handler is only set when `isSendingFeedback` is false, the `disabled` attribute doesn't always stop it from being submitted :(.

magda-web-client/src/Components/FeedbackForm.js
```diff
@@ -56,6 +60,7 @@ class FeedbackForm extends React.Component {
          * Instead of having button disabled, we will check here to make sure feedback is only sent when no validation error.
          */
         if (this.state.validationErrorMessage) return;
+        this.props.sendFeedbacks();
```
AlexGilleran: This'll work but I think it's more idiomatic redux to do it in the `fetchFeedback` function in `feedbackActions.js` - put `dispatch(sendFeedbacks())` at the start the same way as `dispatch(requestDataset(id));` is at the start of `fetchDatasetFromRegistry` in `recordActions.js`

PR #825, opened 2018-04-23
Apply Design System skip link/links styles

maxious:
### What this PR does
Apply Design System typography/links styles. ~~Changes the font across the whole site but~~ many link styles are overridden to match the existing designs for now.

### Checklist
- [x]  Unit tests aren't applicable 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 ## 0.0.39
+*   Apply Design System skip link/links styles
 
 *   Use colours from DTA design guide
 *   Fixed distribution previews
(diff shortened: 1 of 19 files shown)


maxious:
Fixed the white background and viewport issue (related). 
Stories have more below padding. There's excess spacing now with manual break tags and paragraph tag padding but the stories are in markdown. 

Includes fonts from #829

AlexGilleran:
Looks good now, just noticed this though:

image

Is that supposed to be screen reader only?

AlexGilleran:
Getting there! A few link styles are a bit weird which is ok for now, but there's a few things that I'm not super comfortable merging:

image
This has grown a scrollbar :(

image
This isn't visible with the new colour

image
image
image
image
image
image
These have two underlines now, can we get rid of one of them?

There's unnecessary underlines under other links too:
image

These have grown bullet points:
image

maxious:
@AlexGilleran heh, we were importing designguide, overriding then importing designguide again for every component 😂 Taking that out fixes most of the glitches and the tags have been replaces with design system format

PR #833, opened 2018-04-24
Issue/728 Fixed an issue that search may not be added to recent search list in some cases

t83714:
### What this PR does

Fixed an issue that search may not be added to recent search list in some cases based on @tkeuneman 's feedback.

it likely happens on the home page. 

Also, made a tiny adjustment to the recent search icon position.

### Checklist


t83714:
@AlexGilleran we probably want to merge this fix for this deployment as well. Thus, changed the targeted branch to 0.0.38

PR #819, opened 2018-04-20
Issue/395 change linked data rating broken status

adamzerella:
### What this PR does
Linked data rating sleuther now uses `PUT` instead of `PATCH` when writing to registry.
Removed linked data rating takes output from broken link sleuther

### Checklist
- [x] There are unit tests to verify my changes are correct
- [x] I've updated CHANGES.md with what I changed.
- [x] I've linked this PR to an issue in ZenHub (or there is no issue) and dragged it to the _Pull Request_ column

PR #864, opened 2018-04-26
Made data quality consistent between search and details

AlexGilleran:
### What this PR does
- Makes the search page quality rating have the same appearance as the dataset page (⭐️⭐️⭐️).
- Makes the dataset page use the same data as the search (weighted mean of quality aspect) so they're consistent. This will mean they're slightly inflated until #819 lands.

### Checklist


t83714:
tested locally. doesn't find anything unusual. Code reviewed

PR #875, opened 2018-04-26
Distribution previews

steve9164:
### What this PR does
Fixes distribution previews 🎉 . Running at  https://magda-previews.surge.sh

Improvements:
- Distribution previews actually work
- Makes way for changes to where preview compatibility is decided (move towards backend)

Things wrong with this PR:
- There are no changes to viz sleuther, but expects "visualization-info" to someday provide a `compatiblePreviews` object
- It doesn't use compatiblePreviews to determine dataset previews (due to some extra logic there about prioritising specific formats etc. - this should end up in the sleuther as well)
- It leaves some flow types hanging around

### Checklist


magda-web-client/src/UI/DataPreviewNews.js
```diff
@@ -0,0 +1,127 @@
+import React, { Component } from "react";
+import { config } from "../config";
+import fetch from "isomorphic-fetch";
+import getDateString from "../helpers/getDateString";
+import defined from "../helpers/defined";
+import Spinner from "../Components/Spinner";
+
+import type { ParsedDistribution } from "../helpers/record";
+
+import "./News.css";
```
t83714: maybe should be `DataPreviewNews.css`?

t83714:
Code reviewed & Tested locally~

PR #892, opened 2018-05-01
Fix CSP Reports

adamzerella:
### What this PR does
~HelmetJS seems to recommend base urls for URL directives. See: https://helmetjs.github.io/docs/csp/~

~Modified `script-src` directives to only use base URLs.~

- Removes the use of directories in the CSP, now just mentions rollbar directly
- Removes AWS S3 as we don't use it anymore
- Removes unsafe-eval from the allowed sources
- Removes VWO from third-party.js because it's in Google Tag Manager anyway.
- Importantly: Changes the report uri to report-uri.io so it doesn't spam feedback anymore.
- Apparently prettier formatted a bunch of stuff, thanks prettier.
- Deleted and gitignored all the package locks because I'm sick of dealing with them and yarn is imminent.

### Checklist


AlexGilleran:
The thing is that this means anything on S3 or cloudflare can run, which is a pretty broad attack surface.

We can get rid of the S3 one completely, maybe we could copy the rollbar JS over to our own domain?

adamzerella:
I completely agree _but_...if we wanted to be _totally_ secure we should be serving our own copies of scripts. We already are doing worse by not specifying `imgSrc`, `fontSrc`, `styleSrc` and `mediaSrc` which default to `*`

If we do keep using major CDNs though we should assume that S3/Cloudflare are _safe_ no? Ignoring all of this then we should remove the trailing slash for the directives or be doing wildcards as versions can change

e.g requesting: `https://cdnjs.cloudflare.com/ajax/libs/rollbar.js/2.3.9/rollbar.min.js` should 
probably have either directive`cdnjs.cloudflare.com/ajax/libs/rollbar.js/**` OR `cdnjs.cloudflare.com/ajax/libs/rollbar.js/2.3.9/rollbar.min.js` ?

AlexGilleran:
Eh, I wouldn't say `imgSrc` et al are worse - you can do a lot more damage with cross-site-scripting than with cross-site images. I'd definitely prefer to only trust the rollbar part of CDNJS if we can - there's > 3000 libs in there, and I'm not sure how much of an audit they perform.

The wildcard thing sounds awesome if it works - I thought you couldn't do wildcards in CSPs?

t83714:
@AlexGilleran 
Tested with Gitlab & works very well 👍 
Only two things (I believe nothing to do with this PR. So I will still merge it):
- Map Preview report `preview-map/build/TerriaMap.js` 404 error
- Shouldn't `Login link` in top Navigation bar be hidden?

PR #909, opened 2018-05-04
Fixed up the registry api chart

AlexGilleran:
### What this PR does
This was breaking all helm deployments but now it is not. 🎉 

### Checklist
- [x] Doesn't need unit tests
- [x] No need for CHANGES - it corrects a problem that wasn't in the last version
- [x] No issue, we only just noticed it.


t83714:
@AlexGilleran 
It seems also got issue with authApi deployment.yaml. Got error:
```
Error: UPGRADE FAILED: error validating "": error validating data: found invalid field imagePullSecrets for v1.Container
```

t83714:
Test locally and now helm works fine. 
It seems we still have some issues with Gitlab

PR #952, opened 2018-05-09
Tagged the correspondence api as a typescript api.

AlexGilleran:
### What this PR does
The correspondence api was never tagged as an api, which means `yarn run in-modules` doesn't pick it up, and it breaks on the auto deploy to master. This fixes that.

### Checklist
- [x] No issue


t83714:
works well on gitlab

PR #956, opened 2018-05-10
Fixed up docker script to support multiple versions of same yarn pkg

AlexGilleran:
### What this PR does
When I changed the docker scripts that bundle node packages into the images, I had a very naive view of yarn's dependency resolution - they just put everything they could find into the root `node_modules` in the image. Unfortunately if you have multiple versions of the same package anywhere in magda (which we do, obviously) this falls apart... in particular all the sleuthers and the preview-map currently fail because they can't find a submodule of one of their dependencies, which is a much older dependency than the one that's getting bundled into the image.

This makes the scripts actually look at the individual children of each npm module and makes it put it into the node_modules of that package if required, instead of assuming that all packages are declared at the root of `yarn list` and should be present at the root of `node_modules`.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,7 @@
 *   Switched to using yarn as the package manager
 *   Made frontend GA post to both terria and dga google analytics.
 *   Tagged correspondence api as a typescript api.
+*   Fixed up docker scripts to make them handle multiple versions of yarn packages coexisting.
 
 ## 0.0.38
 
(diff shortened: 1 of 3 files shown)


t83714:
Tested in local `minikube` & works well:
`sleuther-format` 
```
Listening at http://sleuther-format/hook
Adding aspect defs Details about the format of the distribution
Successfully added aspect defs
Webhook is already registered
Attempting to resume webhook sleuther-format
Successfully resumed webhook - last event id was 66425
```
`preview-map`:
```
TerriaJS Server 2.7.1
Using config file "/usr/src/app/component/devserverconfig.json".
Serving directory "/usr/src/app/component/wwwroot" on port 6110 to the world.
(TerriaJS-Server running with pid 1)
Launching 2 worker processes.
Convert warning: ogr2ogr (gdal) is not installed or inaccessible, so the format conversion service will fail.
```
Also, had a play via `http://192.168.99.100:30100/preview-map/` & works well

PR #958, opened 2018-05-10
Made master-> dev step deploy connectors too

AlexGilleran:
### What this PR does
Before we weren't updating the connectors (which live outside helm at the moment) when we deployed master via Gitlab. This meant that the connectors kept using the old version of images and the old version of the config. This updates the deploy to master step to update the config/images for the connector cron jobs.

### Checklist


AlexGilleran:
DON'T MERGE THIS YET it's well broken.

PR #966, opened 2018-05-15
Added tooltip to future dates

sukhrajghuman:
### What this PR does

Added tooltip filter to explain future dates when they are selected in the datepicker

### Checklist


chloeleichen:
@sukhrajghuman can you attach a screenshot for what you have done? The description is a bit confusing

chloeleichen:
screen shot 2018-05-15 at 2 15 01 pm
1. I think it's better to have normal font weight for the tooltip text? It's a bit hard to read.
2. Adding the icon on the side squeeze the second date picker and make the two date picker not even width. Can you make them the same width please? Maybe by moving the question mark icon inside the input element?

sukhrajghuman:
Hmmm...not sure why tooltip text is in bold on your screen...
here is what i see:

image

AlexGilleran:
Tooltip text looks fine to me, but it is showing even for dates in the past:

image

PR #976, opened 2018-05-17
Stopped open data rating always showing 0 stars

AlexGilleran:
### What this PR does
Right now the star ratings only show 0 stars because at some point the `QualityIndicator` component, which translated from fractions of 1 to numbers of stars, was removed at some point. This reinstates it and does some other housekeeping around the star ratings - particularly making the link for the tooltip on the questionmark (the link in the tooltip text isn't clickable), and fixing the alignment.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,7 @@
 *   Made magda-web-server look for magda-web-client using require.resolve.
 *   Header adjustment for new design system.
 *   Updated notification with govau design systems components & fixed a few minor error handling related issues
+*   Fixed up datasets always showing as 0 stars.
 
 ## 0.0.38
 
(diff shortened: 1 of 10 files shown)


t83714:
@AlexGilleran 

Tested on gitlab and it looks good!

Only problem is most datasets on 1st page (when search *) show 5 stars but on dataset page show 3 stars.

e.g. http://974-0-star-ratings.magda-dev.terria.io/dataset/ds-dga-52c7294f-dfdc-410e-ba83-539b0bf83931/details?q=*

It seems search API return 0.91 for it but registry return link-rating aspect as 0.6.

AlexGilleran:
😬 I assume that's because it pulled an out-of-date copy of the search index when it created the 974-0-star-ratings, will refresh the search index and see if it's still present.

AlexGilleran:
@t83714 should be consistent now 👍

t83714:
Tested on gitlab~ All good now~

PR #1014, opened 2018-05-18
Simplified elasticsearch query definition and boosted title importance.

AlexGilleran:
### What this PR does
- Boosts the importance of title in the search
- While I was monkeying around with ES, I realised that we could do our query in a much simpler way. Before we were combining 3-4 queries in order to make quotes and english analysis work, turns out we can do it with just a nested query for distributions and a simplequerystringquery for everything else 👍 . This also improves situations where searching for two words (e.g. "government contracts" would return datasets with one of the words features a bunch of times in the description above one with both the words mentioned, which was the better result.
- Testing is a bit hard because it's very down to "feel", but I think this works better - check it out! In particular this should make "government contracts" and "austender" work better.

### Checklist
- [x] There are unit tests to verify my changes are correct (the existing ones still work)


t83714:
Tested using frequent search keywords that used during the usability test.
Works much better than before.

PR #1015, opened 2018-05-21
Issue/935 ds call to action

sukhrajghuman:
### What this PR does
Added CTA on register link. 
Separated register from sign in.

### Checklist

### DIFF
--- a/magda-web-client/src/Components/Account/Login.js
+++ b/magda-web-client/src/Components/Account/Login.js
@@ -64,15 +64,7 @@ export default function Login(props) {
             {props.providers.indexOf("ckan") !== -1 && (
                 <div className="col-sm-6 col-md-5">
                     <h2>Sign In with Data.gov.au</h2>
-                    <p>
-                        This will use your existing data.gov.au account. To
-                        register a new data.gov.au account,{" "}
-                        <AUctaLink
-                            link="http://data.gov.au/user/register"
-                            target="_blank"
-                            text="click here"
-                        />.
-                    </p>
+                    <p>This will use your existing data.gov.au account.</p>
                     <form
                         action={makeLoginUrl("ckan")}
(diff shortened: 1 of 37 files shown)


sukhrajghuman:
@maxious @chloeleichen 

I've noticed the right arrow in the CTA is still in a blue color? I didn't want to change it manually. Could you point me in the direction of how to set up the new colors so that this is also purple. Cheers

image

maxious:
@sukhrajghuman The way the design system website tells you to use the react components is technically correct but doesn't allow our colors. I've done a massive change of the way we import components but it is now purple! 💜

magda-web-client/src/Components/Account/Login.js
```diff
@@ -112,6 +103,16 @@ export default function Login(props) {
                             />
                         </div>
                     </form>
+                    <br />
+                    <h2>Register</h2>
+                    <p>
+                        To register a new data.gov.au account,{" "}
+                        <AUctaLink
+                            link="http://data.gov.au/user/register"
```
AlexGilleran: Can you change this to `https`?

magda-web-client/src/Components/Publisher/PublisherDetails.js
```diff
@@ -49,14 +49,16 @@ class PublisherDetails extends Component {
                             <h3 className="section-heading">Overview</h3>
                             <OverviewBox content={description} />
                         </div>
+                        <br />
                         <div>
-                            <Link
-                                to={`/search?organisation=${encodeURIComponent(
+                            <AUctaLink
```
AlexGilleran: Using the react version means that it won't use the `react-router` link and will cause a full page reload right?

Maybe use the non-react version of the cta link (`<Link className="au-cta-link">`) instead?

PR #1020, opened 2018-05-21
Added class au-body to MarkdownViewer

adamzerella:
### What this PR does
Added class `au-body` to `MarkdownViewer`

### Checklist


AlexGilleran:
@adamzerella I don't think this is quite right - the markdown content is already picking up `au-body` styles, this is more about the fact that in the dataset details section they're getting overridden by specific styles:

image

I've updated https://github.com/terriajs/magda/issues/1008 to be a bit more specific 👍

PR #1022, opened 2018-05-21
Issue/994 Recent search caches images to avoid flickering icons issue on the first display

t83714:
### What this PR does

Recent search now caches images to avoid flickering icons issue on the first display

Tested on Gitlab demo site.

### Checklist

### DIFF
--- a/magda-web-client/src/Components/Search/SearchSuggestionBox.js
+++ b/magda-web-client/src/Components/Search/SearchSuggestionBox.js
@@ -39,10 +39,23 @@ class SearchSuggestionBox extends Component {
             recentSearches: this.retrieveLocalData("recentSearches"),
             selectedItemIdx: null
         };
+        this.cacheImgs();
         this.createSearchDataFromProps(this.props);
         this.searchInputRef = null;
         this.onSearchInputKeyDown = this.onSearchInputKeyDown.bind(this);
         this.containerRef = null;
+        this.cacheImages = [];
+    }
+
+    cacheImg(img) {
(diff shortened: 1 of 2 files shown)


magda-web-client/src/Components/Search/SearchSuggestionBox.js
```diff
@@ -39,12 +39,25 @@ class SearchSuggestionBox extends Component {
...
             selectedItemIdx: null
         };
+        this.cacheImgs();
         this.createSearchDataFromProps(this.props);
         this.searchInputRef = null;
         this.onSearchInputKeyDown = this.onSearchInputKeyDown.bind(this);
         this.containerRef = null;
     }
 
+    cacheImg(img) {
```
AlexGilleran: minor nitpick: could this live inside `cacheImgs()` like 

```javascript
const cacheImg = () => {
    const imgLoader = new Image();
    imgLoader.src = img;
    this.cacheImages.push(imgLoader);
}
```

Just a bit concerned that it depends on `cacheImgs()` being called first to initialize `this.cacheImages`, so it'd be really easy for some other developer to come and call it first by mistake.

AlexGilleran:
Works good for me with chrome simulating a slow 3G connection 👍

magda-web-client/src/Components/Search/SearchSuggestionBox.js
t83714: I think it's a good idea~ 
Will update it.
Thank you~

PR #1030, opened 2018-05-22
Made all environments send email through mailgun.

AlexGilleran:
### What this PR does
Fixes some ports and hosts so all our servers can find mailgun and send emails.

### Checklist
- [x] No issues.


AlexGilleran:
@adamzerella - nah, it always gets overridden. `deploy/helm/magda/charts/correspondence-api/values.yaml` is effectively a set of defaults separate from the environment, and I can't think of a good default to have there given that it's supposed to be usable outside of search.data.gov.au / magda-dev. Better to leave it blank so the pod fails if it's not overridden imho :)

adamzerella:
Everything seems fine to me 👍

PR #1031, opened 2018-05-22
Color system

chloeleichen:
### What this PR does
This is the start of making everything purple


t83714:
@chloeleichen 
not sure whether it's important or. The button text (in screenshot below) is hard to recogonise with the current background.
image

magda-web-client/src/UI/Pagination.js
```diff
@@ -19,7 +19,7 @@ class Pagination extends Component {
         return (
             <button
                 onClick={this.onClick.bind(this, currentIndex - 1)}
-                className="btn-prev au-btn au-btn--secondary"
```
t83714: @chloeleichen 

Those `au-btn` classes were added for solving issue #1044 and lost during the merge I think.

It seems pagination bar won't show differences between `selected` / `unselected` button now.

image

chloeleichen: The `au-btn` style isn't really suitable in this case since they are actually links. Added a different class `current` to identify which is the current page
screen shot 2018-05-28 at 3 13 39 pm

chloeleichen: Fixed:D thanks!

AlexGilleran:
image

The + and - buttons have grown an underline 😬

AlexGilleran:
Also could we make the region outlines purple instead of white? I thought this was in the issue for making the states selectable by default :(.

image

PR #1034, opened 2018-05-22
remove feedback

chloeleichen:
### Checklist


AlexGilleran:
Release blocker so @t83714 to take over:

- Needs to be based on release/0.0.39
- Needs to remove the link from the mobile footer
- Needs to have a CHANGES.md entry

t83714:
@AlexGilleran 
tested and feedback button had been actually removed from both desktop & mobile view correctly.
Only update changes.md to make description more clearly.
Merged with branch 0.0.39 & re-tested it works well

PR #1070, opened 2018-05-23
Refactored react-tooltip to vanilla tooltip

adamzerella:
### What this PR does
Refactored react-tooltip to vanilla tooltip, if this is _safter_ than `react-tooltip` I can tidy it up to make it more user friendly and replace all instances of `react-tooltip`...Yet another package with hundreds of dependencies :(

The emulation mobile testing in Chrome seemed to be ok, haven't done real mobile testing though...might spin up a gitlab preview?

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,6 +64,7 @@
 *   Modified `DatasetSummary` description style to be default `au-body`
 *   Tweaked `au-btn` border-right to be white for search facets.
 *   Made all data.gov.au environments use mailgun.
+*   Implemented `Tooltip` component to replace `react-tooltip` from `QualityIndicator`
 
 ## 0.0.38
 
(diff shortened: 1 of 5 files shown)


tkeuneman:
Helper text for homepage search bar is missing: 
screen shot 2018-05-25 at 1 09 29 pm

adamzerella:
@AlexGilleran I think it's quite weird that I can't replicate any of those issues locally? Is it possible this is only related to the GitLab instance? 

@tkeuneman I don't think I can access events like when a user taps somewhere on screen. We could omit the one second delay on mobile entirely though?

AlexGilleran:
Yeah I think dev might have had a hiccup for not being able to find datasets - http://issue-1000-replace-rules-from-tooltip.magda-dev.terria.io/search works now.

Publishers is a 404 now on dev because it's moved to "/organisations" - we should probably put a redirect in. Unrelated to this ticket though :).

adamzerella:
@tkeuneman The mobile animation delay has now been abolished! Seems more response in emulated chrome testing

PR #1072, opened 2018-05-24
Issue/834 Map preview: Set default camera & mute errors to user

t83714:
### What this PR does

- Set default camera
- Also, set the map to 2d to reduce the animation duration
- Mute errors from user but report to console
- Report select distribution to console so that it's easier to investigate in future
- Make yarn dev running

To test mute errors from users, you can add the followings after [here](https://github.com/TerriaJS/magda/blob/master/magda-preview-map/index.js#L75):

```
window.raiseErrorToUser = e => {
    raiseErrorToUser(terria, e);
};
```
Then, you can trigger an error by (if embeded as iframe):
```
window.iframes[0].raiseErrorToUser(new Error("sssss"))
```
or (if access map module directly)
```
window.raiseErrorToUser(new Error("sssss"))
```

### Checklist

### DIFF
--- a/magda-preview-map/wwwroot/init/terria.json
+++ b/magda-preview-map/wwwroot/init/terria.json
@@ -5,5 +5,6 @@
         "south": -45,
         "west": 109
     },
+    "viewerMode": "2d",
     "catalog": []
 }
--- a/magda-preview-map/wwwroot/init/terria.json
+++ b/magda-preview-map/wwwroot/init/terria.json
@@ -5,6 +5,5 @@
         "south": -45,
         "west": 109
     },
-    "viewerMode": "2d",
     "catalog": []
 }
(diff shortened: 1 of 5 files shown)


magda-web-client/src/UI/DataPreviewMap.js
```diff
@@ -57,6 +57,10 @@ class DataPreviewMap extends Component {
             });
             return;
         } else {
+            if (console && console.log) {
```
AlexGilleran: Are you sure you want to leave this in for production?

magda-preview-map/wwwroot/init/terria.json
```diff
@@ -5,5 +5,6 @@
         "south": -45,
         "west": 109
     },
+    "viewerMode": "2d",
```
AlexGilleran: Maybe let's leave it as 3D on desktop? Especially seeing how it still goes to 3D if the width is too low like I showed you :)

PR #1077, opened 2018-05-25
Fixed doc counts on facet search for publisher/format

AlexGilleran:
### What this PR does
Essentially this is a quick and dirty fix for searching for facet options, which right now is pretty messy - it shows a completely different view and has the wrong document counts if there's no keyword query.

This
- Removes the react autocomplete so it uses the normal view for search results, which makes them display properly.
- Drops the old method of getting every possible format and publisher then filtering client side for the search api again... given that we stopped using react autocomplete it was either this or write another client-side autocomplete. With integration with DAP and its 10,000 distribution datasets on the horizon I decided it was better to go back to searching.
- Hooks the generalQuery up to redux instead of relying on the current URL and removes the old react autocomplete component 

Still needs doing later:
- Making format search results the right case
- Making the results order by count
- Making the search work more like an autocomplete and less like a search

### Checklist


t83714:
@AlexGilleran 
May I clarify what's the requirement for this one? 
Should the doc count meaning be how many more result I can get if I select this format option?
If so, all `generalQuery` should only send search text without include any format parameters, I think?
If it's true, the previously described the incorrect behaviors seem only reproducible on URL `https://magda-dev.terria.io/search` (i.e. https://magda-dev.terria.io/search?q=%2A works alright)?
And, maybe caused by line:
https://github.com/TerriaJS/magda/blob/release/0.0.39/magda-web-client/src/Components/SearchFacets/Format.js#L46
`queryString.parse(this.props.location.search).q` is undefined?

t83714:
@AlexGilleran 
Tested & works great.
Only thing --- not sure whether we should look at it.
It seems publisher will only return result after 3 charaters (i.e. NASA will return no result untill you put NASA in full).
The similar thing applies to format: e.g. you won't get CSV option untill you input CSV in full.
Input `c` will give `X-C(0)` and `cs` will return no result.

PR #1083, opened 2018-05-25
Issue/1047 make end of content spacing consistent

adamzerella:
### What this PR does
Put `PublisherViewer` content into appropriate class container.

Removed `min-height` property for `.app_container`. It seems that only the `dataset-quality` page was actually using this property, other pages exceed 1000px. Mobile testing (emulated) seems to be chill without having the `min-height` property as well. 

**New publisher distance (150px):**
image

### Checklist
- [x]  Unit tests aren't applicable


adamzerella:
@AlexGilleran Made some changes to this PR.

screen shot 2018-06-01 at 11 23 20 am

Did you want the text to be 100% of the row?

Also, unrelated problem but the `min-height:100vh` with the `.app-container` causes whitespace:

screen shot 2018-06-01 at 11 34 27 am

Which footer `margin-top` only promotes...Finally, I was unable to resolve why `OtherPages` caused an annoying whitespace for content that was less than a certain page height, resolved with setting min height.

AlexGilleran:
Sorry, the full width thing was a bum steer - I meant full _height_, which is a problem still:
image

The footer needs to always stick to the bottom of the page.

Also looks like the gap between the content and the footer is now consistent on most of the pages, but it's also zero - needs to be a gap there consistent with the one on the datasets page:
image
- No gap in "Organisers" page

image
- No gap in Organiser page

image
No gap on about page

adamzerella:
@AlexGilleran Sorry for the confusion, everything seems resolved now locally for me, just using the stock `au-footer` props. I've noticed the footer spacing can sometimes get stuck but a page re-render (refresh) fixes that stuff for me.

AlexGilleran:
Nice, coming along now. A few more things:

image
- There's a gap above the "A new look for Australia's data portal" thing - goes away when you click the X so I guess it's attached to that component.

image
- This gap is nice and consistent between the pages now which is great, but it's a bit big... I'd say the gap below "1 - 11 of 1274 results" here should be the same as the gap between the pagination links and the last result.

image
The footer is usually at the bottom now which is awesome, but if you get a really long portrait screen it ends up unsticking eventually.

adamzerella:
@AlexGilleran Fixed the gap at the top (I overlooked the style parent). I've matched the footer to the padding from the pagination result but the `OtherPage` just doesn't stick to the bottom when the page height is less than a certain height...Honestly, no idea how to resolve that or how to refresh the DOM when the page loads faster than the footer style is applied (Leaving it stuck in the middle of screen)...

PR #1085, opened 2018-05-25
Replaced remaining instances of publishers to organisations

adamzerella:
### What this PR does
Renamed remaining instances of `publishers`

Added redirect for /publishers -> /organisations, with quick 302 test

### Checklist


magda-int-test/src/test/scala/au/csiro/data61/magda/api/LanguageAnalyzerSpec.scala
```diff
@@ -73,7 +73,7 @@ class LanguageAnalyzerSpec extends BaseSearchApiSpec {
     def termExtractor(dataSet: DataSet) = dataSet.publisher.toSeq.flatMap(_.name.toSeq)
 
     def test(dataSet: DataSet, publisherName: String, routes: Route, tuples: List[(DataSet, String)]) = {
-      Get(s"""/v0/facets/publisher/options?facetQuery=${encodeForUrl(publisherName)}&limit=10000""") ~> routes ~> check {
+      Get(s"""/v0/facets/organisations/options?facetQuery=${encodeForUrl(publisherName)}&limit=10000""") ~> routes ~> check {
```
AlexGilleran: The facet is still called "publisher" so this will break.

magda-typescript-common/src/JsonConnector.ts
```diff
@@ -321,7 +321,7 @@ export default class JsonConnector {
             });
         });
 
-        app.get("/v0/datasets/:id/publisher", (req, res) => {
+        app.get("/v0/datasets/:id/organisations", (req, res) => {
```
AlexGilleran: I think we can leave this for now, given that we'd have to update admin-client too.

magda-web-client/src/actions/facetPublisherSearchActions.js
```diff
@@ -36,7 +36,7 @@ export function fetchPublisherSearchResults(generalQuery: string) {
         dispatch(requestPublishers(generalQuery));
         return fetch(
             config.searchApiUrl +
-                `facets/publisher/options?generalQuery=${encodeURIComponent(
+                `facets/organisations/options?generalQuery=${encodeURIComponent(
```
AlexGilleran: Ditto re: the publisher facet not being renamed.

magda-web-client/src/stories/index.js
```diff
@@ -177,7 +177,7 @@ storiesOf("API viewer", module)
     ))
     .add("organisations viewer", () => (
         <ApiViewer
-            url="http://search.data.gov.au/api/v0/search/facets/publisher/options?generalQuery=*&facetQuery=*&limit=562&orderedBy=hitCount%22"
+            url="http://search.data.gov.au/api/v0/search/facets/organisations/options?generalQuery=*&facetQuery=*&limit=562&orderedBy=hitCount%22"
```
AlexGilleran: Ditto re: the publisher facet not being renamed.

magda-web-server/src/index.ts
```diff
@@ -213,7 +213,7 @@ const topLevelRoutes = [
     "sign-in-redirect",
     "dataset",
     "projects",
-    "publishers"
+    "organisations"
```
AlexGilleran: Good catch!

However it'd probably be best to leave both of these so that react-router can handle the routing.

magda-web-server/src/index.ts
```diff
@@ -225,6 +225,14 @@ topLevelRoutes.forEach(topLevelRoute => {
     });
 });
 
+/**
+ * For whatever reason a user still requests /publishers, redirect to /organisations
+ */
+app.get("/publishers", function(req, res) {
```
AlexGilleran: This kinda works but it'd be even better to have it done at the react-router level: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md.

This way if we missed a `<Link to="/publishers" />` it'll catch that too.

AlexGilleran:
Seeing how adam is on holiday today I've taken this over :)

PR #1095, opened 2018-05-28
Made prettier format package.json in a compliant way.

AlexGilleran:
### What this PR does
Stops the constant tug of war between yarn, which wants to make package.json have a tabWidth of '2' and prettier, which wants to give it '4'. Turns out that package.json is only ever supposed to have a tabWidth of 2 so that pretty much settles it - this configures prettier to make an exception of all `package.json`s for tabWidth, and also commits the result.

### Checklist


maxious:
Oh that explains a lot :P

PR #1101, opened 2018-05-28
Added `baseExternalUrl` to `server-config.js` &  Make `open in national map button` uses `baseExternalUrl`

t83714:
### What this PR does

- Added `baseExternalUrl` to `server-config.js`
- Make `open in national map` button uses `baseExternalUrl`
- As this will only change the baseUrl used by `open in national map` button, it shouldn't impact any other function.

Cannot test locally or in a minikube (or gitlab demo site) as https connection is required for the cross-domain request from national map site.

Probably the only way to test is merged to master and tested from the dev site.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,277 +1,277 @@
 ## 0.0.40
 
--   Changed Publishers to Organisations in UI
--   Modified `MonthPicker` tooltip icon's right padding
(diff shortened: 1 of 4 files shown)


t83714:
we don't need this anymore~

PR #1103, opened 2018-05-29
Merge release 0.0.40

AlexGilleran:
### What this PR does
- Merges release 0.0.40
- Fixes national map links pointing at `/` instead of the external magda api url.


deploy/helm/magda-dev.yml
```diff
@@ -131,7 +131,6 @@ correspondence-api:
     requests:
       cpu: 100m
       memory: 0
-  externalUrl: "https://magda-dev.terria.io"
```
t83714: Will this cause any trouble for dev site deployment (master branch)?
It seems `externalUrl` now is from `externalUrl=http://$CI_COMMIT_REF_SLUG.magda-dev.terria.io`
Does that mean dev site won't get value "https://magda-dev.terria.io" for `externalUrl`?
`Open in National Button` seems only work with `https` URL for cross-domain request

magda-web-server/src/index.ts
```diff
@@ -146,6 +146,7 @@ app.get("/server-config.js", function(req, res) {
         disableAuthenticationFeatures: argv.disableAuthenticationFeatures,
         baseUrl: addTrailingSlash(argv.baseUrl),
         apiBaseUrl: apiBaseUrl,
+        baseExternalUrl: argv.baseExternalUrl,
```
t83714: Do we need `addTrailingSlash` it like `baseExternalUrl: addTrailingSlash(argv.baseExternalUrl)`?

deploy/helm/magda-dev.yml
AlexGilleran: Dev site should be ok - externalUrl is still present on line 4, I've just made correspondence api use that one.

Would probably be wise to override it for the ui-only previews though 🤔

t83714: I think it's good enough then~ 👍

magda-web-server/src/index.ts
AlexGilleran: Well, it works like it is, but I suppose it'll work with a trailing slash too - I'll add one and see.

deploy/helm/magda-dev.yml
t83714: Probably don't need to override it as I guess one day gitlab preview will support `https`

PR #1104, opened 2018-05-29
Made stop words get filtered out of search tests more reliably

AlexGilleran:
### What this PR does
Right now in the `LanguageAnalyzerSpec`, the test input generator will occasionally throw in a `for` or a `was` and cause a failing test case because these words get filtered out and can't be reliably searched on.

This adds more robust checking for this case so it won't happen and slow down the tests any more.

### Checklist


t83714:
Good stuff~ 
May just need it 👍 
Thanks!

PR #1121, opened 2018-05-29
style org page

chloeleichen:
### What this PR does
screen shot 2018-05-29 at 4 26 05 pm
### Checklist


AlexGilleran:
I'm happy with this code-wise - @tkeuneman or @philipgrimmett can you QA the UI part? it's at http://org-page.magda-dev.terria.io/organisations

philipgrimmett:
Looking good! Seems like the Heading styles are different from the About page though? Also the heading should be Sentence case, not upper case.

It would be great to have the number of datasets listed under each publisher (in the same style as on the datasets results page), but not essential if we're short on time.

tkeuneman:
+1 to what @philipgrimmett suggested. We can do the dataset count under a seperate ticket.

chloeleichen:
@philipgrimmett the page heading on about page is black, should I make them all purple?

chloeleichen:
done: 
screen shot 2018-06-04 at 5 07 52 pm

PR #1123, opened 2018-05-30
Bumped regions version

AlexGilleran:
### What this PR does
Bumps the regions version so that it creates itself a new index with crispy's abbreviations in it.

You can test it at http://1120-bump-regions-version.magda-dev.terria.io/search - there's no datasets because that's broken but the regions are already loaded if you use the locations facet.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -82,6 +82,7 @@
 *   Map Preview: Mute errors from users
 *   Map Preview: Report selected distribution to console
 *   Adjusted pagination button styles
+*   Bumped regions index version to 31 because of region shortname
 
 ## 0.0.38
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -82,6 +82,7 @@
 *   Map Preview: Mute errors from users
 *   Map Preview: Report selected distribution to console
 *   Adjusted pagination button styles
+*   Bumped regions index version to 31 because of region shortname
```
t83714: Only a tiny one: here should be 21.
In the meantime, I will do a test in minikube

t83714:
@AlexGilleran 
It seems indexer won't re-index region automatically after the version changes.
How can I manually trigger it?

PR #1124, opened 2018-05-30
Added redirect for CKAN dataset URLs (UUID or slug)

maxious:
### What this PR does
Added redirect for CKAN dataset URLs (UUID or slug) including resource URLs (just back to the main dataset page)

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@
 -   Fixed a issue that state region layer could be removed from region filter panel map
 -   Adjust search box placeholder color to be more visible on mobile and more consistent on desktop
 -   Corrected email template style to be inline with designs.
+-   Added redirect for CKAN dataset URLs (UUID or slug) to their new canonical location
 
 ## 0.0.40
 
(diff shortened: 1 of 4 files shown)


AlexGilleran:
Works well for datasets hosted directly by data.gov.au, but it doesn't support externally harvested datasets like https://data.gov.au/dataset/aurora-australis-voyage-1-2015-16-track-and-underway-data. I don't think we're ever gonna be able to nail it 100%, but maybe we could do a quoted search on the de-slugged url (e.g. http://magda-dev.terria.io/search?q=%22aurora%20australis%20voyage%201%202015%2016%20track%20and%20underway%20data%22) and if there's one result then redirect there as a last resort?

AlexGilleran:
@maxious nice - sorry I didn't get to this sooner, I think there was a rebase or something that confused me.

Anyway looks great except for this, which is a bit confusing for new users:
image

Is there an easy way to leave the search bar blank?

maxious:
Heh, I copied code that intentionally preserves the query, should be able to take that out 👍

PR #1128, opened 2018-06-01
Issue/986 Make Elasticsearch supports synonyms & merged Alex's branch for quote

t83714:
### What this PR does

- Updated `elasticsearch` docker image to include WordNet synonym lib
- Created custom `english_with_synonym` analyzer &  `english_without_synonym_for_search` analyzer
- Applied `synonym`  analyzer to the following fields:
  - description
  - distributions.description
  - themes
- In order to pass `LanguageAnalyzer` test suit, a `CustomAnalyzer` was create in order to use `KStemmer` (another name of `light_english` stemmer in apache `lucene` codebase)
  - Use `CustomAnalyzer` because the `KStemmer` class & `stem` method is not public in the package `org.apache.lucene.analysis.en`. That's the easiest way I can figure out to ultilise the `KStemmer` function in our code

We didn't apply `synonym` analyzer to `title` field ( and also lifted up its weight) to make sure the first-page result would be similar to the current version.

Eventually, I only added `synonym` token filter to `index analyzer` rather than `search analyzer` as:
there is no way to assign `weight` to `synonym token` generated. Thus, `search analyzer` seems largely impact the return result sorting.

- Also merged Alex's branch for quote #1166 

### Checklist


magda-scala-common/src/main/resources/common.conf
```diff
@@ -7,7 +7,7 @@ elasticSearch {
     }
 
     datasets {
-      version = 32
+      version = 34
```
AlexGilleran: 33?

magda-scala-common/src/main/scala/au/csiro/data61/magda/search/elasticsearch/IndexDefinition.scala
```diff
@@ -46,17 +58,72 @@ case class IndexDefinition(
...
+    Some("analysis/wn_s.pl"),
+    Set.empty,
+    None,
+    Some("wordnet")
+  )
+
   val dataSets: IndexDefinition = new IndexDefinition(
     name = "datasets",
-    version = 32,
+    version = 34,
```
AlexGilleran: 33

magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/ElasticSearchQueryer.scala
```diff
@@ -321,12 +321,12 @@ class ElasticSearchQueryer(indices: Indices = DefaultIndices)(
     val clauses: Seq[Traversable[QueryDefinition]] = Seq(
       query.freeText flatMap { inputText =>
         val text = if (inputText.trim.length == 0) "*" else inputText
-        val queryString = new SimpleStringQueryDefinition(text).defaultOperator(operator).quoteFieldSuffix("keyword")
+        val queryString = new SimpleStringQueryDefinition(text).defaultOperator(operator).quoteFieldSuffix(".quote")
 
         // For some reason to make english analysis work properly you need to specifically hit the english fields.
         def foldFieldsEnglish(query: SimpleStringQueryDefinition, fields: Seq[Any]) = fields.foldRight(query) {
```
AlexGilleran: Forgot to do this in mind - you can get rid of this function, it's now a copy paste of foldFields

magda-scala-common/src/main/resources/common.conf
t83714: Bump to 34 just in case you merged your branch to master earlier
Will revert back to 33

t83714:
It's a good idea to merge next week so we can give everyone a chance to feel the search result changes.

t83714:
Some adjustment for quote test case:
- Added a new Analyzer
  - Make sure quoted query string can match a portion of the field content
  - But the whole quoted string still needs to be matched extactly
  - Use white space tokenizer to match our test case
- Quote test case: filter out the blank term

t83714:
@AlexGilleran 
Previously the test cases issue seems purely performance of `elasticsearch` on gitlab.
Google it and most answers suggest tuning the resource related config (e.g. more memory) --- and this probably not an option.
Tried other changes in code --- mainly on two things: slow down thread running & smaller sample wordNet file
Changes made since the last comment:
- replaced `blockUntilNotYellow` call with `blockUntilNotRed` so tests can run a little bit faster
- call `blockUntilNotRed` before some test . --- it seems `beforeAll` only works for testSuits not `funSpec`?
- Adjustment on some fields' `quote` subfield analyzer setting --- use `keyword` tokenizer for search analyzer
Re-index in indexer has been started in dev cluster --- still running

t83714:
closed as this PR has been merged into a new PR:
https://github.com/magda-io/magda/pull/1364

PR #1129, opened 2018-06-01
For Jacky: Changes to publisher acronym search

AlexGilleran:
Unless you are @t83714 ignore me!

### DIFF
--- a/magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/Queries.scala
+++ b/magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/Queries.scala
@@ -22,24 +22,15 @@ object Queries {
   def publisherQuery(strategy: SearchStrategy, searchForAcronym: Boolean = false)(publisher: FilterValue[String]) = {
     handleFilterValue(publisher, (publisherString: String) =>
       strategy match {
-        case SearchStrategy.MatchAll => if(!searchForAcronym) {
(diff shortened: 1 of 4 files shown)


magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/Queries.scala
```diff
@@ -19,27 +19,18 @@ import au.csiro.data61.magda.model.misc.Region
...
-                .minimumShouldMatch("-50%")
-        } else {
-            multiMatchQuery(publisherString)
-                .fields("publisher.acronym", "publisher.name", "publisher.name.english")
-                .minimumShouldMatch("-50%")
-        }
-      }, "publisher.name"
-    )
+        case SearchStrategy.MatchAll =>
+          matchQuery("publisher.name.keyword_lowercase", publisherString)
```
t83714: Can I confirm that the reason we don't need to worry about the `buildQueryWithAggregations(inputQuery, start, limit, MatchAll, requestedFacetSize)` call [on line 82](https://github.com/TerriaJS/magda/blob/83a32f5d88be0ee1a3c891adbb45b0dec71fc45d/magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/ElasticSearchQueryer.scala#L82) is: `acronym` is unlikely to match any full name and that call will eventually fall back to `MatchPart` call `buildQueryWithAggregations(inputQuery, start, limit, MatchPart, requestedFacetSize)` on line 88?

Issue #1142, opened 2018-06-05
Refreshing suggest page at /suggest gives an error


steve9164:
### Problem description
Navigating to /suggest using browser history API is fine, but refreshing the page gives an error:
> Cannot GET /suggest

### Problem reproduction steps
1. Click "Suggest a dataset" in footer
1. Once the suggest form is loaded, click refresh

adamzerella:
@steve9164 I'm not able to replicate this issue locally? Go to `/suggest`, wait for it to load and refresh? I've also tried sending a request then refreshing and it seems fine.

What was the error?

steve9164:
![image](https://user-images.githubusercontent.com/13863530/41017820-0e8805f8-699a-11e8-8da3-f9ab1aea75b3.png)

This is an incompatibility between magda-gateway routes and the /suggest route in React Router, so it will only happen when running the UI through the gateway.

Best fix is probably to move `/suggest` to `/page/suggest`.

steve9164:
Looks like /page/... routes are for static html pages generated from markdown. So instead the fix should be to add the /suggest route into magda-gateway.

PR #1189, opened 2018-06-06
Add /suggest route to web server config

tephenavies:
### What this PR does
For #1142.

### Checklist
- [x] **I don't think this should have a CHANGES.md entry** (this is fixing a bug from an unreleased feature that is documented in the changelog)


steve9164:
Looks like that formatting change is due to an update of prettier.

PR #1191, opened 2018-06-07
Made format and publisher autocomplete give correct doc counts.

AlexGilleran:
### What this PR does
Short term fix before https://github.com/TerriaJS/magda/issues/979 gets done.

Right now when they get doc counts, the publisher and format filter parse the current url directly, using the "q" param and passing that all the way to the search api. Unfortunately since then we've made it that the "q" param isn't always present, so this defaults it to "*" in line with the actual search.

### Checklist


t83714:
Tested locally by:
- Click dataset link in the header
- Check the publisher panel for entry `Australian Securities`
- The count matches the number

PR #1193, opened 2018-06-07
Added Zendesk feedback widget activated via footer link

maxious:
### What this PR does
Added Zendesk feedback widget activated via footer link

### Checklist


magda-web-client/public/index.html
```diff
@@ -38,17 +38,32 @@
...
+    </script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/es5-shim/4.5.10/es5-shim.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/es5-shim/4.5.10/es5-sham.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/es6-shim/0.35.3/es6-shim.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/es6-shim/0.35.3/es6-sham.min.js"></script>
     <![endif]-->
-
+    <!-- Start of datagovau Zendesk Widget script -->
+    <script>/*<![CDATA[*/
+    window.zE || (function (e, t, s) {
```
AlexGilleran: Can this live in thirdparty.js?

magda-web-client/src/config.js
```diff
@@ -110,7 +110,8 @@ export const config = {
                             ? "mailto:data@digital.gov.au"
                             : "suggest"
                     ],
-                    ["Privacy Policy", "page/privacy-policy"]
+                    ["Privacy Policy", "page/privacy-policy"],
+                    ["Give feedback", "feedback"]
```
AlexGilleran: This will mean it only shows on desktop, is that the plan?

maxious: Mobile/small already had one of these links included in the config: https://github.com/TerriaJS/magda/blob/f5616ae1875cb27fc28adbf74955d0fd79a00c77/magda-web-client/src/config.js#L97

maxious:
Fixed CSP error as per developer.zendesk.com/embeddables/docs/widget/csp

AlexGilleran:
Can get the form now, looks pretty good.

Some issues:
image
- This comes up - I couldn't see it when I tested this branch before, I assume it's because the CSP is blocking something Zendesk needs?

image
- This happens when I try to submit the form.

Is there a way to show the "? Feedback" floaty button by default, instead of only if I cancel the form? Like the same functionality that we had with the old feedback form/button. @tkeuneman might want to weigh in on that.
image

AlexGilleran:
Cool, looks good and works for me now. One last nitpick - the footer link doesn't have the right cursor when you hover over it.

@tkeuneman do you want to make sure this is all good? http://1038-replace-feedback-form.magda-dev.terria.io/

AlexGilleran:
Just realised tash is away so 🚢ing it

PR #1196, opened 2018-06-07
Cache the yarn cache in gitlab, not node_modules

AlexGilleran:
### What this PR does
Rather than cache `node_modules`, this uses the yarn cache. This means that best-case-scenario builds are a bit slower because everything in `node_modules` has to be rebuilt, but also means that the size of the cache being uploaded/downloaded each time is smaller, and importantly should help with the problems we've been seeing recently where the cached `node_modules` folder doesn't quite match `yarn.lock` or `package.json` and yarn ends up taking up to an hour and failing on step 

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,6 @@ stages:
   - release
   - deploy-prod
 
-
 builders-and-yarn:
   stage: builders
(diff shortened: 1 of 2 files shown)


t83714:
Thanks~
Looks works alright in gitlab. Merged

PR #1197, opened 2018-06-08
Restricted markdown strong tag

adamzerella:
### What this PR does
Restricted markdown strong tag

### Checklist


adamzerella:
Preview of changes: 
screen shot 2018-06-08 at 11 14 56 am

AlexGilleran:
This looks like it's actually making `<strong>` tags smaller than the surrounding text:

image

Also is there any way to make it so that we're only restricting the markdown on the search results page rather than on both the search results and the dataset page? I think rendering the markdown normally on the dataset page still makes sense.

It'd also be good if we did something about making `text inside backticks` render normally inside search results.

adamzerella:
@AlexGilleran Made some changes to isolate overridden markdown styling to just be under `.dataset-summary-description`. Couldn't see a better solution besides overriding the `code` styling to mimic the `au-body` font styling though...

## Before:
screen shot 2018-06-12 at 3 00 53 pm

## After:
screen shot 2018-06-12 at 3 00 38 pm

## DatasetDetails now retains the markdown styling:
screen shot 2018-06-12 at 2 52 08 pm

PR #1215, opened 2018-06-12
Issue 1178/show number of filters selected

sukhrajghuman:
### What this PR does
show number of filters applied when searching
### Checklist

### DIFF
--- a/magda-web-client/src/Components/Search/Search.js
+++ b/magda-web-client/src/Components/Search/Search.js
@@ -39,8 +39,7 @@ class Search extends Component {
         // it needs to be undefined here, so the default value should be from the url
         // once this value is set, the value should always be from the user input
         this.state = {
-            searchText: undefined,
-            filterCount: 0
+            searchText: undefined
         };
     }
 
@@ -139,7 +138,7 @@ class Search extends Component {
      * this is then appended to the results text on the search page
      */
     filterCount = () => {
-        var count = 0;
+        let count = 0;
         if (this.props.activePublishers.length > 0) {
             count++;
         }
@@ -151,11 +150,11 @@ class Search extends Component {
             count++;
         }
 
-        if (this.props.activeDateFrom) {
+        if (this.props.activeDateFrom && this.props.activeDateTo) {
             count++;
         }
         if (count !== 0) {
-            var filterText = count === 1 ? " filter" : " filters";
+            const filterText = count === 1 ? " filter" : " filters";
             return " with " + count + filterText;
         } else {
             return "";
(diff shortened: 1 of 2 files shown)


magda-web-client/src/Components/Search/Search.js
```diff
@@ -39,7 +39,8 @@ class Search extends Component {
         // it needs to be undefined here, so the default value should be from the url
         // once this value is set, the value should always be from the user input
         this.state = {
-            searchText: undefined
+            searchText: undefined,
+            filterCount: 0
```
AlexGilleran: When does this get used?

magda-web-client/src/Components/Search/Search.js
```diff
@@ -133,6 +134,34 @@ class Search extends Component {
         });
     }
 
+    /**
+     * counts the number of filters that have active values
+     * this is then appended to the results text on the search page
+     */
+    filterCount = () => {
+        var count = 0;
```
AlexGilleran: Can you use `let` instead? :)

magda-web-client/src/Components/Search/Search.js
```diff
@@ -133,6 +134,34 @@ class Search extends Component {
...
+        }
+        if (this.props.activeFormats.length > 0) {
+            count++;
+        }
+
+        if (this.props.activeRegion.regionId) {
+            count++;
+        }
+
+        if (this.props.activeDateFrom) {
```
AlexGilleran: What about `activeDateTo` ?

magda-web-client/src/Components/Search/Search.js
```diff
@@ -133,6 +134,34 @@ class Search extends Component {
...
+            count++;
+        }
+        if (count !== 0) {
+            var filterText = count === 1 ? " filter" : " filters";
```
AlexGilleran: Use `const` please :)

magda-web-client/src/Components/Search/Search.js
sukhrajghuman: @AlexGilleran we don't really need check for both activeDateTo and activeDateFrom. Since once a user selects a date (either "from" or "to"), the other is automatically set as well. Basically every search query is going to have both activeDateTo and activeDateFrom...if that makes sense. Will make the other changes

sukhrajghuman: I'll chuck it in anyway

magda-web-client/src/Components/Search/Search.js
```diff
@@ -133,6 +133,34 @@ class Search extends Component {
...
+        }
+        if (this.props.activeFormats.length > 0) {
+            count++;
+        }
+
+        if (this.props.activeRegion.regionId) {
+            count++;
+        }
+
+        if (this.props.activeDateFrom && this.props.activeDateTo) {
```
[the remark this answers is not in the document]
sukhrajghuman: yup done

PR #1219, opened 2018-06-14
Refactoring/eslint fixes inc. React async rendering lifecycles

maxious:
### What this PR does
Refactoring/eslint fixes inc. React async rendering lifecycles. We updated eslint/prettier recently which have new rules. Left unmet, these files will fail to validate next time they are edited.

### Checklist


magda-web-client/src/Components/SearchFacets/FacetTemporal.js
```diff
@@ -27,7 +27,7 @@ class FacetTemporal extends Component {
         };
     }
 
-    componentWillReceiveProps(nextProps) {
+    getDerivedStateFromProps(nextProps) {
```
AlexGilleran: Shouldn't this be returning the new state now instead of using `setState` ?

magda-web-client/src/UI/DataPreviewMap.js
```diff
@@ -70,7 +70,7 @@ class DataPreviewMap extends Component {
         }
     }
 
-    componentWillMount() {
+    UNSAFE_componentWillMount() {
```
AlexGilleran: Looks like as long as you have `getDerivedStateFromProps` lower down you can just delete this altogether?

magda-web-client/src/UI/DataPreviewMap.js
```diff
@@ -82,7 +82,7 @@ class DataPreviewMap extends Component {
         window.removeEventListener("message", this.onIframeMessageReceived);
     }
 
-    componentWillReceiveProps(props) {
+    getDerivedStateFromProps(props) {
```
AlexGilleran: This and `createStateFromProps` should be returning the new state right?

magda-web-client/src/UI/MonthPicker.js
```diff
@@ -37,7 +37,7 @@ class MonthPicker extends Component {
         });
     }
 
-    componentWillReceiveProps(nextProps) {
+    getDerivedStateFromProps(nextProps) {
```
AlexGilleran: Should be returning state?

AlexGilleran:
Just had a test on http://react16willmount-eslint-fixes.magda-dev.terria.io, unfortunately there's a bunch of stuff broken :(.

- Can only search once, after that the search results don't update
- No search suggestions drop down
- Date filter doesn't work
- Map preview doesn't load

maxious:
@AlexGilleran Okay, these should be fixed or downgraded to just adding UNSAFE_

AlexGilleran:
Still can't get the date filter to work - it opens but clicking on any button fails.

Also going to certain datasets (I assume ones with maps) like https://react16willmount-eslint-fixes.dev.magda.io/dataset/ds-dga-fa0b0d71-b8b8-4af8-bc59-0b000ce0d5e4?q= seems to create some kind of infinite loop where the dataset never loads but the URL at the top of the browser keeps oscillating between two values.

PR #1222, opened 2018-06-15
Issue/581 filters not applied

sukhrajghuman:
### What this PR does
date and region filter can be applied when clicking outside the box
### Checklist


magda-web-client/src/Components/SearchFacets/FacetTemporal.js
```diff
@@ -29,10 +29,10 @@ class FacetTemporal extends Component {
     }
 
     componentWillMount() {
-        this.componentWillReceiveProps(this.props);
+        this.getDerivedStateFromProps(this.props);
     }
 
-    componentWillReceiveProps(nextProps) {
+    getDerivedStateFromProps(nextProps) {
```
AlexGilleran: The idea is to return the state from `getDerivedStateFromProps` instead of using `setState` :)

PR #1253, opened 2018-06-19
issue/1236 - improve dataset page

sukhrajghuman:
1. Removed unneccessary underscores from file names
2. Made file and API names wrap in screen.
3. Made Modal fit better on smaller screen
4. Improved design of modal
5. Removed underlines from filenames.
6. Added Animation for modal opening
7. Added blur when opening modal

Modal fits better on mobile now. Though the feedback form button I
believe is
stuffing up the body grid.

### Checklist

### DIFF
--- a/magda-web-client/src/Components/Dataset/DatasetSuggestForm.scss
+++ b/magda-web-client/src/Components/Dataset/DatasetSuggestForm.scss
@@ -2,13 +2,17 @@
 
 .ask-dataset-form {
     padding: 10px;
-    border: solid;
-    border-color: $AU-color-foreground-focus;
-    background-color: $AU-colordark-foreground-action;
     overflow: -webkit-paged-y;
     position: relative;
     float: right;
     width: 800px;
+    background-color: rgb(248, 247, 247);
+}
+
+.modal-header {
+    color: white;
+    background-color: #4b3081;
+    text-align: center;
 }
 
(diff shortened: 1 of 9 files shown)


magda-web-client/src/Components/Dataset/DatasetSuggestForm.scss
```diff
@@ -2,13 +2,17 @@
...
 .ask-dataset-form {
     padding: 10px;
-    border: solid;
-    border-color: $AU-color-foreground-focus;
-    background-color: $AU-colordark-foreground-action;
     overflow: -webkit-paged-y;
     position: relative;
     float: right;
     width: 800px;
+    background-color: rgb(248, 247, 247);
```
chloeleichen: Can we use $AU-colordark-foreground-action instead? it is rgb(241,247,238) which is very close to this color here and can help keep the theme consistent

sukhrajghuman: hi @chloeleichen  that color looks really bad. Please see below

image
****

chloeleichen: yeah, i agree. Happy to use the new color 💃

sukhrajghuman: 😸

PR #1289, opened 2018-06-25
Issue/1280 Change registry /records API `pageToken` parameter to a Long type

t83714:
### What this PR does

Change registry /records API `pageToken` parameter to a Long type to avoid the possible runtime exceptions.

Currently, it's declared as String type and convert to Long in the runtime before sent to DB.

Any incorrect format will trigger an uncaught `NumberFormatException`, which will print an error log entry.

Declare it as `number` type parameter will prevent uncaught exception be thrown (thus,, no log will be printed) and users will get correct `400 bad request` response:

```
The query parameter 'pageToken' was malformed:
'dsd12800' is not a valid 64-bit signed integer value
```

Updated:
- to avoid impact typescript API, the parameter was converted manually 

### Checklist

-   [x]  Unit tests aren't applicable

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 -   Add breadcrumb for dataset page and distribution page
 -   Added UNSAFE\_ annotations or refactored to prepare for [React async rendering](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html).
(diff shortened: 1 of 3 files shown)


t83714:
@AlexGilleran 
You are right~
It's much simpler now~

PR #1293, opened 2018-06-26
Removed totalCount from registry api responses

AlexGilleran:
### What this PR does
Previously we were doing a lot of really expensive full-table `COUNT(*)`s in the registry api when we almost never needed to. This moves counts into their own endpoint, and changes the `totalCount` field that was previously present in Registry API pagination to a `hasMore` field like the Stack Overflow API.

- Removes `totalCount` field from registry api pages (we weren't using it anyway except in one spot in redux)
- Replaces use of `totalCount` in redux for organisations with the new `/count` endpoint
- Removes projects in the ui - we haven't used this for like a year and I don't think it's coming back any time soon
- Adjusts the indexer crawling to work with the new API - it wasn't using the count anyway.

### Checklist


magda-web-client/src/actions/publisherActions.js
```diff
@@ -49,18 +49,35 @@ export function requestPublisherError(error: FetchError): FacetAction {
 function fetchPublishers(start) {
     return (dispatch: Function) => {
         dispatch(requestPublishers());
-        const url = `${
+
```
t83714: Are changes in this files are intended to be included in this PR?

t83714: Please ignore my previous comment~ I see the purpose now~

PR #1325, opened 2018-06-28
Removed <button> dataset download link which looks like button

adamzerella:
### What this PR does
Removed <a> elemnt from button component
Replaced button component with DS `AUButton`.

### Checklist

# Preview: 
image


steve9164:
Is the change of 
`[Icon] Text`
to
`Text [Icon]`
on the button intended?

AlexGilleran:
Good point, I'll flip that back around.

chloeleichen:
@AlexGilleran ready for another look

Issue #986, opened 2018-05-17
Feed synonyms into elastic search engine


tkeuneman:
Find a synonym library that we can input into elastic search. For example, rivers, streams and waterways.

Searches to check on:
"companies" should find datasets for "business", particularly by asic

Note: This is unlikely to make it for the usability test unless we get really lucky, so don't stress about that too much.

AlexGilleran:
In `master` now.

PR #1364, opened 2018-07-04
Upgrade Elasticsearch to v6.3.0 & including synonyms supports

t83714:
### What this PR does

- Switched to HTTP client. 
  - We now have more control on the connection timeout time (can make very long request i.e. more than 10 minutes)
    - New settings are in common.conf or application.conf
      -  This mainly controlled `socketTimeout` setting (the name was quite misleading)
  - Test cases will be running using DefaultClientProvider as well
    - with test only setting in `TestActorSystem.config`
- Upgrade Elasticsearch & Kibana to 6.3.0
  - Set max locked memory to unlimited ( in both daemonset & es-nodes otherwise won't pass bootcheck)
  - Remove http.enabled --- it's now depreacated as HTTP is compulsory
  - Turn on auto_create_index for system indices:
    - ".security*,.monitoring*,.watches,.triggered_watches,.watcher-history*,.ml*"
- The package `com.vividsolutions.jts` is now renamed to `org.locationtech.jts`
  - the project was moved to `locationtech` so we have to change the name for es embeded node
- Move elastic4s `ElasticDsl API` to `au.csiro.data61.magda.search.elasticsearch.ElasticDsl`
  - **All code should import that file rather than original `com.sksamuel.elastic4s.http.ElasticDsl`**
  - This is our entry point to fix some bugs and fill some gap
  
## ElasticDsl API changes
- should import `au.csiro.data61.magda.search.elasticsearch.ElasticDsl` to access API
- client will now return an `Either`
- response data types are slightly different:
  - e.g. before you got `RichSearchResponse` now you got `SearchResponse`
Example:

```
client.execute(ElasticDsl.search(query)).map{
   case Right(results) =>
    //--- you can access `SearchResponse` from results.result
    //--- to acess aggregations:
    // results.result.aggregations.cardinality("name")
   Left(ESGenericException(e)) => throw e
  //--- I created more types of exceptions (available from `au.csiro.data61.magda.search.elasticsearch.Exceptions`)
}
```
At this moment, elastic4s are not 100% mapping all responses yet. 
If you really need some fields that are not available from built-in type yet, you can customise from `Excutables`

The current aggregations API are very young & primitive --- it is just a simply wrap of a `Map[String, Any]` structure

I have to add a few functions to do what I want. You can check the API from `au.csiro.data61.magda.search.elasticsearch.AggregationResults`
 

This pr are related to issue:
#1364 , #986 

### Checklist


t83714:
Travis was failed as:
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated
```
Looks like we need to reduce the amount of output to log 
For now, just restart the job

magda-int-test/src/test/scala/au/csiro/data61/magda/api/DataSetSearchSpec.scala
```diff
@@ -110,6 +113,63 @@ class DataSetSearchSpec extends BaseSearchApiSpec with RegistryConverters {
       }
     }
 
+    describe("searching for a dataset should return that datasets contains the keyword & it's synonyms") {
+      it("for synonyms group 300032733 `agile`, `nimble`, `quick` & `spry`") {
+        case class GenResult(searchKeyword: String, synonym: String, datasetWithSynonym: DataSet)
+
+        val synonyms = List("agile", "nimble", "quick", "spry")
+        val pairs = synonyms.map(s => {
+          val listWithItself = synonyms.filterNot(_===s)
```
AlexGilleran: should this be `listWithoutItself` ?

magda-int-test/src/test/scala/au/csiro/data61/magda/api/DataSetSearchSpec.scala
```diff
@@ -110,6 +113,63 @@ class DataSetSearchSpec extends BaseSearchApiSpec with RegistryConverters {
...
+              dataset <- Generators.dataSetGen(scala.collection.mutable.HashMap.empty)
+              datasetWithSynonym = dataset.copy( description = Some(synonym))
+            } yield GenResult(searchKeyword, synonym, datasetWithSynonym)
+            gen.retryUntil(_ => true).sample.get
+        }
+
+        val synonymDataset = synonymTestData.map(_.datasetWithSynonym)
+
+        val searchKeywordGen = for {
+          genResult <- Gen.oneOf(synonymTestData)
```
AlexGilleran: Is it possible to put the `randomDatasets` and `synonymTestData` into this generator instead of using `.retryUntil(_ => true).sample.get` ? Or was that too slow?

magda-int-test/src/test/scala/au/csiro/data61/magda/test/util/MagdaMatchers.scala
```diff
@@ -26,6 +27,19 @@ object MagdaMatchers extends org.scalatest.Matchers {
     if (stemmer.stem) stemmer.getCurrent else string
   }
 
+  def lightEnglishStem(string: String): String = {
+    val stemmer = new LightEnglishStemmer
+    return stemmer.stem(string)
+  }
+
+  def stemString(term: String, useLightEnglishStemmer:Boolean = false) = {
```
AlexGilleran: So do we still need the PorterStemmer? Can't we use the light english one for everything?

magda-int-test/src/test/scala/au/csiro/data61/magda/test/util/SharedElasticSugar.scala
```diff
@@ -0,0 +1,236 @@
...
+import org.elasticsearch.ResourceAlreadyExistsException
+import org.elasticsearch.transport.RemoteTransportException
+import org.slf4j.{Logger, LoggerFactory}
+
+trait SharedElasticSugar extends HttpElasticSugar
+
+/**
+  * Provides helper methods for things like refreshing an index, and blocking until an
+  * index has a certain count of documents. These methods are very useful when writing
+  * tests to allow for blocking, iterative coding
```
AlexGilleran: Do you mean "imperative" instead of "iterative"?

t83714:
@AlexGilleran 

If looked the content of the au.csiro.data61.magda.search.elasticsearch.ElasticDsl, it mainly imports other packages from elastic4s.
Inside the file, I only override the package that we need to fix.
This is relatively low impact way to fix things.
At this moment, I can't find a better to fix things (especially, some of the request generation issue)

One more thing, elastic4s HTTP API is very immature at this moment (TCP client has been killed already in master). We will have to change something from time to times. 

Considering the simplicity of HTTP client, if we want to have more control over this area, we probably should think about fork elastic4s and have our own version.

You probably noticed I didn't use their testkit & embed node package as well (to overcome some spatial package issue & elasticsearch packages issue). 

I also have a long list of issues haven't got a chance to submit PRs to `elastic4s` project yet. 

The other option is (I actually planned to do this when I got a chance):
- I will gradually go through our problems and submit PRs to elastic4s projects 
- Once the project is more stable, we then do a merge again: We probably don't need to a file merge -- as our changes are more likely to make fields available or fix incorrect field value in requests

magda-int-test/src/test/scala/au/csiro/data61/magda/test/util/MagdaMatchers.scala
t83714: The stemmer outcome is different. 
And `PorterStemmer` is part of standard built-in `english` analyzer
I was thinking a better way should be:
- we have a new function can read index definition 
- And construct the same analyzer process
- Then we can guarantee there is no exception --- at this moment, it's possible some combination of filters /tokenizerd results empty input as well

magda-int-test/src/test/scala/au/csiro/data61/magda/api/DataSetSearchSpec.scala
t83714: @AlexGilleran 
I am actually not very familiar with the generator API.
Those lines are copies of your `acronym` test cases example.
What I was trying to do is:
- Generate 20 random datasets as `background noise`
- Use built-in synonyms list to generate a few datasets contains synonyms words
- Put all datasets above into `elasticsearch` and search synonym keyword
Do you suggest putting both datasets generation into searchKeywordGen? Why do we want to do that?

AlexGilleran: `.retryUntil(_ => true).sample.get` is a bit dodgy because it breaks ScalaCheck's ability to shrink input... say you generate `randomDatasets` as `List(a, b, c)`. If you're using the generator inside a `forAll`, then if the test fails Scalacheck is able to try it with `randomDatasets` = `List(a, b)`, `List(b, c)` etc until it finds the smallest possible failing case. 

I guess this is ok though, because if you had all the generated datasets in the `forAll` you'd have to insert a new index 100 times instead of once, which'd be too slow 🤷‍♂️

t83714: Thanks a lot for your comments 👍 
Good to know the difference & I'd better also have a read on ScalaCheck documents

t83714:
@AlexGilleran 

Further to our conversation: 
Do you think it would pratical option that we:
- Still merge this PR fo now (so this enable all future deveopment talk to new interface)
- I created another issue for moving this all my `elastic4s` changes to either a sub-module or sperate maven artifacts

The benefit of this option would be:
- We don't have to re-write all new development later (merging `master` on Monday does take me very long time to debug)
- Allow us to spend more time on `creating an elastic4s` fork so we don't have to do things in a rush
- Project management wise. We can avoid creating a giant work block & we are more confident on creating new feratures
- The future `creating an elastic4s` fork changes would be `transparent` to all new development. So we don't have to worry about re-writiing anything

We probably can have a chat tomorrow if you have time~

t83714:
@AlexGilleran 
I think I probably will close this PR as another version with all elastic4s files removed is available at another branch:
https://github.com/magda-io/magda/tree/feature/es-http-changeover-maven

t83714:
Stop old gitlab preview site. Will create another one for new PR.

PR #1391, opened 2018-07-09
Neatened up (some of the) broken-looking charts

AlexGilleran:
### What this PR does
Essentially I just went through the top CSVs and added code to make them chart a bit better:

- Adds more heuristics to guess data types from column names for charting CSV files
- Rearranges the code to process data for echarts a bit - particularly pulling static methods out of the class definition, pulling constants up to the top of the file
- Adds a retry if the initial CSV parsing doesn't work with `\n` forced as the newline character, for SoE CSVs which don't work with autodetection for some reason
- Makes it so that different fields are aggregated together with `sum` for everything except scatter graphs - this means that there's only ever one line / set of bars / etc, and helps the problem where bars would stack on top of each other.
- Makes `Count` an option for the Y axis even if there are other axes, just in case the other axes don't make sense to graph.
- Turns the x axis labels 45 degrees so that you can read more categories on a bar graph
- Parses dates for `time` series if possible
- Sorts the series so we don't get any more back-and-forwards line graphs.

### Checklist


magda-web-client/src/helpers/ChartDatasetEncoder.js
```diff
@@ -12,47 +12,145 @@ import takeRight from "lodash/takeRight";
...
+    "year"
+];
+/** Columns with these terms in them will be overridden as time types */
+const OVERRIDE_TO_TIME_COLUMNS = ["date", "time"];
+/** Columns with these terms in them will be overridden as numeric types */
+const OVERRIDE_TO_NUMERIC_COLUMNS = [
+    "amt",
+    "amount",
+    "sum",
+    "count",
```
t83714: I think probably the regex expression (testKeywords) doesn't enforce the word edging (or string edge). So `country` is now set to a numeric column.  
One example: https://1140-preview-chart-dates.dev.magda.io/dataset/ds-dga-ab7eddce-84df-4098-bc8f-500d0d9776d1/details?q=*
A quick fix (if the regex is hard to fix) could be:
- remove  the `count` from `OVERRIDE_TO_NUMERIC_COLUMNS `
- or add `country` to `OVERRIDE_TO_CATEGORY_COLUMNS `

magda-web-client/src/helpers/ChartDatasetEncoder.js
```diff
@@ -12,47 +12,145 @@ import takeRight from "lodash/takeRight";
...
+const LOW_PRIORITY_X_AXES = ["name"];
+/** Columns with these names are more likely to be picked as the default column on the y axis */
+const HIGH_PRIORITY_Y_AXES = [];
+/** Columns with these names will not be picked as the default column on the y axis unless there's nothing else */
+const LOW_PRIORITY_Y_AXES = ["count", "id"];
+
+/** Columns with these terms in them will be overridden as category types */
+const OVERRIDE_TO_CATEGORY_COLUMNS = [
+    "id",
+    "code",
```
t83714: - Probably we want to keep `age` as category columns? ---- it likely a user wants to see other data according to different ages?
- `lic num` probably is another one we want to keep as category columns? --- otherwise, the result will be license number added up. 
- "longitude", "latitude" probably should be treated as category columns (it won't produce any meaningful result by adding the value up)?  However, move them to xAxis could also likely lead to too many columns (or very busy chart). Probably, another option is to remove those columns from both `xAxis` & `yAxis`. i.e. Exclude them from the chart as they won't be useful for either `xAxis` & `yAxis` columns?

magda-web-client/src/helpers/ChartDatasetEncoder.js
```diff
@@ -155,173 +395,30 @@ class ChartDatasetEncoder {
...
-                isAggr: true,
-                isAggrDone: false
-            });
-            //--- we cann't generate coutn data here yet as we don't know user's selection
-        }
+        //--- Present data by selected col's count
+
+        const newFieldName = "count";
+        this.fields.push({
+            idx: 1,
```
t83714: Since we are adding `count` for all chart, we'd better make sure it's unique among all columns (it (and probably `name`) will be used as column reference by eCharts).
We can set to: 
```
idx: this.fields.length+1,
```

magda-web-client/src/helpers/ChartDatasetEncoder.js
```diff
@@ -123,6 +222,147 @@ const defaultChartOption = {
...
+    let newFields = map(disFields, (field, key) => ({
+        ...field,
+        idx: indexOf(headerRow, key),
+        name: key,
+        label: startCase(key),
+        category: !field.time && !field.numeric,
+        isAggr: false
+    }));
+    //--- filter out fields that cannot be located in CSV data. VisualInfo outdated maybe?
+    newFields = filter(disFields, item => item.idx !== -1);
```
t83714: This dataset now shows a blank column (in the xAxis dropdown):
https://1140-preview-chart-dates.dev.magda.io/dataset/ds-sa-5db51968-5902-4bf5-a177-ad83d805bd38/details?q=income
But production site looks ok:
https://search.data.gov.au/dataset/ds-sa-5db51968-5902-4bf5-a177-ad83d805bd38/details?q=income
I think the issue might be around this line (`disFields` should be `newFields`) but couldn't be certain

t83714:
Well done! @AlexGilleran ~ It works much better and the code is much cleaner now~ 👍 
Merge data points together (by xAxis value) helps a lot when rendering some datasets.

magda-web-client/src/helpers/ChartDatasetEncoder.js
AlexGilleran: I added the country to override category - I still want to have the regex super-greedy so it catches column names like `BLAHCOUNT` or `COUNTBLAH`

magda-web-client/src/helpers/ChartDatasetEncoder.js
AlexGilleran: - Yeah good point re: age as a category. I've kept age as a numeric for now because given the choice between being able to have age on the Y axis or the X axis, I prefer the Y axis (so you can graph age by category, for instance) but we should make it so you can put numeric columns on either axis at some point.
- Added `lic` to overrides for category columns
- Filtered out `lat` and `long` from all axes.

magda-web-client/src/helpers/ChartDatasetEncoder.js
AlexGilleran: Done :)

magda-web-client/src/helpers/ChartDatasetEncoder.js
AlexGilleran: Yeah oops 😬

AlexGilleran:
Made some fixes and made blank values format as "Unknown".

AlexGilleran:
Oh also added (Total Sum) when aggregating.

magda-web-client/src/helpers/ChartDatasetEncoder.js
t83714: @AlexGilleran 
Yeah, ideally the regex should be super-greedy to avoid miss matches. I had a few tries initially but give up due to the timelines --- one issue related to this is we probably want the regex greedy on three different boundaries:  non-word boundary, $ (ending string) & ^ (beginning of the string).
Second thought on this, I think it might be easier if we use regex to split the string to words and reformat them before the match a list of keyword.
Having said that, do you want to merge this PR for now (for the spirit of agile 😄 ) since all fixes are in place and continue the development on a new issue / new PR?

AlexGilleran:
:shipit: !

PR #1392, opened 2018-07-09
wrap chart title

chloeleichen:
### What this PR does
Wrap chart title
screen shot 2018-07-09 at 11 08 39 am

### Checklist

### DIFF
--- a/magda-web-client/src/UI/DataPreviewChart.js
+++ b/magda-web-client/src/UI/DataPreviewChart.js
@@ -64,7 +64,7 @@ class DataPreviewChart extends Component {
 
                 if (!chartType) chartType = defaultChartType;
                 this.chartDatasetEncoder.setChartType(chartType);
-                const chartOption = this.chartDatasetEncoder.getChartOption();
+                const chartOption = this.chartDatasetEncoder.getChartOption("");
 
                 this.setState({
                     error: null,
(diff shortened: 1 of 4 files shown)


magda-web-client/src/UI/DataPreviewChart.js
```diff
@@ -64,9 +64,7 @@ class DataPreviewChart extends Component {
 
                 if (!chartType) chartType = defaultChartType;
                 this.chartDatasetEncoder.setChartType(chartType);
-                const chartOption = this.chartDatasetEncoder.getChartOption(
-                    this.state.chartTitle
-                );
+                const chartOption = this.chartDatasetEncoder.getChartOption();
```
t83714: Since we don't need chart lib to generate title anymore, we probably could just remove the parameter from this method and set title field to([docs](https://ecomfe.github.io/echarts-doc/public/en/option.html#title.show)):
```
            title: {
                show: false
            },
```
around this line:
https://github.com/magda-io/magda/blob/fe1379c12724188328c44defc242eb63f271c352/magda-web-client/src/helpers/ChartDatasetEncoder.js#L585 

Currently, the field was actually set to `undefined` --- it didn't create any errors but I guess the chart lib didn't expect that value.
Alternatively, we could simply try this: `this.chartDatasetEncoder.getChartOption("")` to make sure that field is set to a valid value.

PR #1393, opened 2018-07-09
trim empty rows before table rendering

chloeleichen:
### Checklist

To test, go to http://localhost:6108/dataset/ds-dga-70f76a9e-d450-4567-9760-290dea54229c/details?q=medicare and sort by any column


magda-web-client/src/UI/DataPreviewTable.js
```diff
@@ -85,6 +85,24 @@ export default class DataPreviewTable extends Component<
             });
     }
 
+    removeEmptyRows(data) {
+        let dataForTable = [];
```
AlexGilleran: How would you feel about 
```javascript
return data.filter(row => 
    Object.values(row).some(column => 
        column.trim().length > 0
    )
);
```
?

magda-web-client/src/UI/DataPreviewTable.js
```diff
@@ -117,16 +135,17 @@ export default class DataPreviewTable extends Component<
             Header: item,
             accessor: item
         }));
+        const rows = this.removeEmptyRows(this.state.parsedResults.data);
```
AlexGilleran: I'm a bit wary about looping through every row on every render - could we use `memoize-one` like in `DataPreviewMap` ?

chloeleichen: it actually only renders 3 times: 1st on initially request data, 2 on requesting data, 3 on receiving all data. subsequent sorting does not cause re-rendering. Changing tab do cause re-render but that is because we unmount the current tab when switching to another. Not sure how much use would `memoize-one` be in our case

magda-web-client/src/UI/DataPreviewTable.js
chloeleichen: updated! Thanks!

magda-web-client/src/UI/DataPreviewTable.js
AlexGilleran: OK then 😄 🤷‍♂️

PR #1418, opened 2018-07-11
Upgraded TerriaJS and made gateway more configurable

### LINKED ISSUE #1348: Lock down our CORS policy
### Problem description
From the pen test

Right now for all CORS we simply reflect back the `Origin` header:
E.g.
image

I think it's fair enough that we should allow CORS for our non-authenticated APIs in general, but we should change it to `*` to be clear, and disable `Access-Control-Allow-Credentials: true`.

Any APIs that require authentication should have a specific CORS whitelist, and only use `Access-Control-Allow-Credentials` in development.

This is probably something that should happen at the gateway level.

AlexGilleran:
### What this PR does
- Uses the latest, eval-free TerriaJS version
- Turns off terrain in the Terria Map
- Makes the externalUrl for the web pod independently overridable, and overrides it to `dev.magda.io` for UI previews.
- Moves the config for a bunch of stuff the gateway does - CSP, CORS, HelmetJS, Proxying etc into helm configmaps. This means that you can now have different security settings in each environment, particularly making it so that the dev instance can be embedded in an iframe.
-  As a result this also Fixes #1348 🎉 

### Checklist


t83714:
@AlexGilleran 
Map preview works but Open in national map button not work properly.
It seems there is no `externalUrl` Preview helm chart and global default "http://localhost:6100" was used.
What do you think it's the best way to solve that?
Options I can think about:
- Change the hardcoded value to `https://dev.magda.io`
- Or is there any helm parameter can set the value over command line?

AlexGilleran:
@t83714 addressed that and some other stuff (see description)

t83714:
@AlexGilleran 
https://1417-upgrade-terriajs.dev.magda.io/dataset/ds-dga-33c46400-4c6f-45db-8476-de216f6aac1e/details?q=city
It seems gateway now won't proxy request to correct search API node:
```
::ffff:10.0.2.192 - - [13/Jul/2018:01:46:37 +0000] "GET /api/v0/search/datasets?query=&start=0&limit=10 HTTP/1.1" 500 - "https://1417-upgrade-terriajs.dev.magda.io/search?q=" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.99 Safari/537.36"
{ Error: connect ECONNREFUSED 127.0.0.1:6102
    at Object.exports._errnoException (util.js:1020:11)
    at exports._exceptionWithHostPort (util.js:1043:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1099:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6102 }
```
probably gateway didn't load routes.json from in /etc/config?

AlexGilleran:
Yeah I think I did a final test, renamed a variable then forgot to change all references to it :(. Rebuilding...

AlexGilleran:
@t83714 working again.

a

deploy/helm/magda/charts/gateway/values.yaml
```diff
@@ -26,3 +26,51 @@ resources: {}
...
+  correspondence:
+    to: http://correspondence-api/v0/public
+helmet: {}
+csp:
+  browserSniff: false
+  directives:
+    scriptSrc:
+    - "''self''"
+    - "''unsafe-inline''" # For VWO, hopefully we can remove this soon
+    - "''data:''" # Also for VWO
```
t83714: A very minor thing:
The `data:` should be included in CSP without single quote `'`
image

.gitlab-ci.yml
```diff
@@ -448,19 +448,19 @@ dockerize:dockerExtensions:
     # Helm upgrade
     - helm init --upgrade
   script:
-    - helm upgrade $CI_COMMIT_REF_SLUG deploy/helm/magda --install --recreate-pods --namespace $CI_COMMIT_REF_SLUG -f deploy/helm/preview.yml --set global.image.repository=registry.gitlab.com/magda-data/magda/data61,global.image.tag=$CI_COMMIT_REF_SLUG,ingress.hostname=$CI_COMMIT_REF_SLUG.dev.magda.io,correspondence-api.externalUrl=http://$CI_COMMIT_REF_SLUG.dev.magda.io --timeout 1200 --wait
+    - helm upgrade $CI_COMMIT_REF_SLUG deploy/helm/magda --install --recreate-pods --namespace $CI_COMMIT_REF_SLUG -f deploy/helm/preview.yml --set global.image.repository=registry.gitlab.com/magda-data/magda/data61,global.image.tag=$CI_COMMIT_REF_SLUG,ingress.hostname=$CI_COMMIT_REF_SLUG.dev.magda.io,correspondence-api.externalUrl=https://$CI_COMMIT_REF_SLUG.dev.magda.io --timeout 1200 --wait
```
t83714: Sorry. I really should check everything in one go.
Should be this line:
```
correspondence-api.externalUrl=https://$CI_COMMIT_REF_SLUG.dev.magda.io
```
or this line at the bottom:
```
web-server.externalUrl
```
both be:
```
global.externalUrl=https://$CI_COMMIT_REF_SLUG.dev.magda.io
```
currently, the value is still return localhost in server-config.js

t83714: I actually not sure the outcome of 
```
"--baseExternalUrl", {{ .Values.externalUrl | default .Values.global.externalUrl | quote }}
```
So please correct me if I suggest something wrongly...

AlexGilleran: Yeah so the idea is that I override the externalUrl but just for the web-server pod, so that the link to nationalmap works.

But now I look at it with fresh eyes it doesn't make much sense, the real solution is to make it so that the nationalmap link goes to `baseUri` instead of `externalUri` because it'll be pointing at the API.

Also I'm not sure why `correspondence-api` needs its own external uri, I'll look at that too.

AlexGilleran:
@t83714 I've improved it a bit more:

- Removed the `correspondence-api.externalUrl` stuff from gitlab-ci. This was never used. There's just one global externalUrl again now, which the web-server uses for the sitemap and nothing else.
- Made the UI pass urls based off the API url that it's using to nationalmap. If this is `/`, then it prepends its own url (minus path) to make it external. This means that it'll pass `dev.magda.io` in ui-only previews but pass its own url outside of that.

t83714:
Looks good~ Well done! 
Merged~

Issue #1106, opened 2018-05-29
remove duplicate tooltip component 


chloeleichen:
### Problem description
We have two different component doing the same thing in the UI for tooltip : 
![screen shot 2018-05-29 at 1 44 29 pm](https://user-images.githubusercontent.com/7508939/40637162-91ec80a4-6346-11e8-8b31-53208cce4b48.png)

The `Tooltip` component should be removed and all instance replaced with ReactTooltip

adamzerella:
@chloeleichen I think we are going to transition away from `react-tooltip` and use the custom solution to avoid some problems `react-tooltip` has, like using `dangerouslySetInnerHTML`

t83714:
We probably need an alternative option to replace `transition` CSS for this issue:  #1152

aneesha09:
Closing it

PR #1444, opened 2018-07-13
Tooltip style

chloeleichen:
### What this PR does
1. unified the two mentioned tooltip styles, set font-size to 0.9rem
2. aligned with star rating box

Eventually, we want to unify tooltip design/implementation, as mentioned here: https://github.com/magda-io/magda/issues/1106. But that would be a different issue

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,6 +33,7 @@
 -   Improved styling of suggest dataset form
 -   Improve distribution link alignment
 -   Improve UI for chart loading error
+-   Improve tooltip look
 
 ## 0.0.43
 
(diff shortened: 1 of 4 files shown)


AlexGilleran:
@tkeuneman or @philipgrimmett can you look at this? https://tooltip-style.dev.magda.io/

It's consistent but looks a bit cramped to me now.

tkeuneman:
Sigh. We've lost the previous work in this PR. There's meant to be a second delay once a user moves away from the tooltip.

The tooltip on the results page is still out of alignment.

I think increasing the padding around the text so that it's 16px or 12px would help make this look like it can breathe a bit more.

AlexGilleran:
Merging into https://github.com/magda-io/magda/issues/1447 because that has another tooltip.

PR #1446, opened 2018-07-13
Issue/1438 Added Auto GZip compress to gateway

t83714:
### What this PR does

Issue/1438 Added Auto GZip compress to gateway
Used `compression` lib: https://github.com/expressjs/compression
Checked the source code. By default, It can:
- Only compress `compressible` content by looking at `Content-type` header of res
  - It uses [compressible](https://www.npmjs.com/package/compressible) for this
- Will add `Accept-Encoding` [Vary](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary) header (and [this](https://stackoverflow.com/questions/43217103/google-cdn-random-not-serving-gzip-content)) so CDN can serve non-compress copy to upsupported client and compressed copy to supported client.
- Will only compress if client accept gzip or deflate encoding (by looking at Accept Encoding from request)

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,6 +33,7 @@
 -   Improved styling of suggest dataset form
 -   Improve distribution link alignment
 -   Improve UI for chart loading error
+-   Added auto gzip compress to gateway
 
 ## 0.0.43
 
--- a/magda-gateway/package.json
+++ b/magda-gateway/package.json
@@ -40,7 +40,8 @@
     "pg": "^6.4.0",
     "tsmonad": "^0.7.2",
     "urijs": "^1.18.12",
-    "yargs": "^8.0.2"
+    "yargs": "^8.0.2",
+    "compression": "^1.7.2"
   },
   "devDependencies": {
     "@magda/scripts": "^0.0.44-0",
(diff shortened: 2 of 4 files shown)


AlexGilleran:
It mostly works but seems to GZIP some things and not others :/.

- Responses from the web server are gzipped
- Most API responses aren't GZIPped - e.g. https://issue-1438.dev.magda.io/api/v0/search/datasets?start=0&limit=10
- ... but some are, e.g. https://issue-1438.dev.magda.io/api/v0/search/region-types

Weird, those last two both have `content-type: application/json` 🤔

AlexGilleran:
Maybe it's the size of the latter? 🤔

t83714:
@AlexGilleran 
Yes. The source code indicates it only gzip content if it's bigger enough --- too small response you won't be benefited too much from compress (and considering the overhead might not worth it )
The default value it's 1024. 
The lib will auto check this by looking at Content-Length header.
But it's configurable.

AlexGilleran:
But it's the other way around - the big one doesn't zip and the small one does 😢

t83714:
...argh...if so, I think we just discovered a bug here.
emm... from the source code doesn't suggest that way:
https://github.com/expressjs/compression/blob/7ebf291087d69156df4b086e2020e660a1dffb17/index.js#L158
I will have a few tests as well.

AlexGilleran:
@t83714 don't worry I got it the wrong way around 😂

AlexGilleran:
Yep confirmed that once I've put some data in, https://issue-1438.dev.magda.io/api/v0/search/datasets?start=0&limit=10 has GZIP 🎉

PR #1472, opened 2018-07-18
Resolves #1260 by making GET requests for sources where HTTP HEAD ret…

### LINKED ISSUE #1260: Make broken link enhancer test remote endpoint / url using HTTP `GET`
### Problem description

Currently, we have a backend program to test all remote links liveness.

For performance consideration, we used HTTP `HEAD` request (instead of `GET`) to test the liveness.

As per standard here: https://tools.ietf.org/html/rfc7231#section-4.1

`HEAD` requests should trigger same behaviour as `GET` requests.  

However,  many servers block `HEAD` method incorrectly.

We might want to find a way to use `GET` request to test liveness and disconnect the connection once we receive the headers. 

This will improve the compitiablity of our broken link enhancer program

```
curl -I http://mapping.launceston.tas.gov.au/arcgis/rest/ces/Public/DetailSurvey/MapServer/2
HTTP/1.1 405 Method Not Allowed
Content-Length: 662
Content-Type: text/html;charset=utf-8
Date: Wed, 20 Jun 2018 00:57:46 GMT
Server: 
Vary: Origin
Connection: keep-alive
```

nahidakbar:
Made broken links sleuther perform a get request with content range when head request returns 405 (method not allowed).

### Checklist


AlexGilleran:
So I was trying to test this with the original Launceston open data system that caused it to be required but it turns out that now every single distribution link they return is 404ing even for GET lol 😂 .

@nahidakbar the code looks good to me, can you add yourself to `contributors.md` and sign the CLA?

Hopefully launceston fixes their data.json endpoint soon.

magda-sleuther-broken-link/src/onRecordFound.ts
```diff
@@ -266,6 +266,36 @@ function retrieveHttp(
                         response.statusCode === 429
                     ) {
                         resolve(response.statusCode);
+                    } else if (response.statusCode === 405) {
```
AlexGilleran: This makes sense, but can we just make it retry with `GET` if it's any bad code? Unfortunately some servers (like launceston open data, see here: https://1260-broken-link-head-405-to-get.dev.magda.io/api/v0/registry/records/ds-launceston-http%3A%2F%2Fdata-launceston.opendata.arcgis.com%2Fdatasets%2Fb02fbc0eee4c4ec2a374ab817d2af199_5?dereference=true&aspect=dcat-dataset-strings&optionalAspect=dcat-distribution-strings&optionalAspect=dataset-distributions&optionalAspect=temporal-coverage&optionalAspect=dataset-publisher&optionalAspect=source&optionalAspect=source-link-status&optionalAspect=dataset-quality-rating) respond with `400` to `HEAD` but `200` with `GET`

AlexGilleran: Or better yet if it times out too 🤔

nahidakbar:
This is ready to go as well @AlexGilleran

AlexGilleran:
Looks like the test might need updating? https://gitlab.com/magda-data/magda/-/jobs/85212398

Issue #1460, opened 2018-07-17
Content gets bumped down the page when searching for organisation


tkeuneman:
### Problem description
When searching (or clearing search), a small space appears above the breadcrumbs which pushes all the content below it down and makes the user's eyes bleed 👁👁 💦.

### Reproduction
On organisations page, search for something, and clear the search afterwards. You'll notice the content jump up and down.

PR #1475, opened 2018-07-18
Issue/1460 Fixed organisation page content jumping issue

t83714:
### What this PR does

Fixed organisation page content jumping issue (#1460)

What happened was the loading bar component didn't play well with `Breadcrumbs` (Don't know the root cause).

Once loading bar is under the same parent as `Breadcrumbs`,  `Breadcrumbs` will get extra margin --- although loading bar is an `absolute positioned` element.

Move the loading bar away (into content part) as other pages fixed this.

### Checklist

### DIFF
--- a/magda-web-client/src/Components/Publisher/PublishersViewer.scss
+++ b/magda-web-client/src/Components/Publisher/PublishersViewer.scss
@@ -65,6 +65,3 @@
         min-height: 650px;
     }
 }
-body {
-    overflow-y: scroll !important;
-}
--- a/magda-web-client/src/Components/Publisher/PublishersViewer.scss
+++ b/magda-web-client/src/Components/Publisher/PublishersViewer.scss
@@ -61,4 +61,10 @@
             cursor: pointer;
         }
     }
+    .org-result-page-body {
+        min-height: 650px;
+    }
+}
+body {
+    overflow-y: scroll !important;
 }
--- a/magda-web-client/src/index.scss
+++ b/magda-web-client/src/index.scss
@@ -2,7 +2,7 @@
 @import "./pancake/sass/pancake.scss";
 html {
 }
-.au-body {
+body {
     overflow-y: scroll;
 }
 .au-body .au-grid {
--- a/magda-web-client/src/index.scss
+++ b/magda-web-client/src/index.scss
@@ -2,7 +2,9 @@
 @import "./pancake/sass/pancake.scss";
 html {
 }
-
+.au-body {
+    overflow-y: scroll;
+}
 .au-body .au-grid {
     p,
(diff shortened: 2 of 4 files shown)


AlexGilleran:
Nice work @t83714 - it still jumps sideways though, because the scrollbar disappears and reappears? Can we set `overflow-y: scroll` on `body` to ensure there's always a scrollbar?

t83714:
Good Job~ 
I think you must have used IE for the testing 👍 
Scrollbar on mac didn't take any space... 😄 
Will update the program and let you know~

t83714:
Added  `overflow-y: scroll` to body (also added min-height to result body)
Please have a look again~

magda-web-client/src/Components/Publisher/PublishersViewer.scss
```diff
@@ -61,4 +61,10 @@
             cursor: pointer;
         }
     }
+    .org-result-page-body {
+        min-height: 650px;
+    }
+}
+body {
+    overflow-y: scroll !important;
```
AlexGilleran: I think this belongs in `global.scss`

t83714:
That's a very good point~
Lucky you spot it otherwise it cloud be hard for someone to find out where the global body CSS kicked in later 👍 
it seems we didn't `load global.scss` anymore.
I put in `index.scss` instead.

magda-web-client/src/index.scss
```diff
@@ -2,7 +2,9 @@
 @import "./pancake/sass/pancake.scss";
 html {
 }
-
+.au-body {
```
AlexGilleran: This still has to be `body` instead of `.au-body` because `.au-body` gets used in a bunch of places including the footer, mobile header, banner :(.

t83714:
Thanks for the pickup 👍 
It's updated now~

AlexGilleran:
Searching is 💯 , but I just realised while testing that it's calling `pushState` for every single character you type into the search, which means that if you search for "geoscience" then press "back", it goes back to "geoscienc" :(. Sorry, I realise you didn't cause this but it needs fixing before we can release I think.

t83714:
That's really a good job to spot this issue 👍 
Clean up the code a little bit to fix this issue.
Basically stick to the following rules:
- Requesting data will only be triggerred by URL changes
- URL changes will only be triggered by debounced version `updateSearchQuery`
- All other user interactions (e.g. input changes or button clicks) talk to debounced version `updateSearchQuery`
- Maintaince text input status in Component state rather `location` property
Hopefully this fixes all~ 🤞

AlexGilleran:
:+1: :+1: 💯 🎉

Issue #1469, opened 2018-07-17
"Clear" button inside filter dropdown is the same as the apply button


AlexGilleran:
### Problem description
If you open the dropdown for a filter when something is selected and click "Clear", it doesn't clear - it has the same effect as "Apply".

### Problem reproduction steps
- Go to the search page
- Apply a filter (e.g. PDF)
- Reopen the dropdown and click "clear"
- Note that the filter value is still selected.

### Screenshot / Design / File reference
![image](https://user-images.githubusercontent.com/900555/42804386-3ce8eabc-89ec-11e8-830b-07676fb1c202.png)

PR #1476, opened 2018-07-19
Issue/1469 Fixed "Clear" button inside filter dropdown is the same as the apply button

t83714:
### What this PR does

Issue/1469 Fixed "Clear" button inside the filter drop-down is the same as the apply button (#1469 )

It's probably easy to make `clear` button functional again but hard to keep the `click outside the box to apply filter` feature work at the same time.

We have to refactor the code a little bit (I didn't touch `Temporal` as it works well).  Basically,
- Make facet body live longer so we can maintain the relevant information required for the logic
- Instead of hook up to `componentWillUnmount` for auto apply function. Monitor property changes via `componentDidUpdate`

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,7 @@
 -   Added auto gzip compress to gateway
 -   Made CSW connector retrieve country field from alternative JSON path
 -   Fixed Sitemap times out if input is invalid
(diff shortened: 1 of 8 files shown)


AlexGilleran:
+1+1+1+1+1+1

PR #1504, opened 2018-07-24
Fixed scatter graphs.

AlexGilleran:
### What this PR does
Stops the scatter graphs from displaying nothing.

### Checklist

-   [x] No CHANGES, this resolves something that was working last version


AlexGilleran:
@t83714 updated to fix the issues with blank labels - now it says unknown in the legend and tooltip.

Re: the rollbar thing - I get it too but at random times... e.g. I can't reproduce it now at all. I've got no idea what's causing it, are you sure it's to do with the charts?

Re: putting `random()` on the end of the name for the count column - why? Is there ever more than one count?

t83714:
Works well~ 👍 
Merged.
We'd better look at the random rollbar issue when we can reproduce it rather than now

PR #1528, opened 2018-07-31
Resolves #1397 initial crack at api documentation

### LINKED ISSUE #1397: SPIKE - Investigate API documentation
### Problem description
How should we document APIs in different microservices?

nahidakbar:
### Checklist


nahidakbar:
:thumbsup: @AlexGilleran

AlexGilleran:
Looks awesome. How do you feel about having apidoc run in its own pod though? So every time we build the apidoc gets generated and dropped on top of something like https://github.com/nishanttotla/DockerStaticSite, so that it's not dependent on the built-in magda web server, then we route it through the gateway at `/apidoc`?

Also could you add a top-level `yarn run` command and put it into a gitlab ci job?

nahidakbar:
@AlexGilleran Thanks for the config fixes. Changes look good.

PR #1545, opened 2018-08-01
facet layout

chloeleichen:
### What this PR does
Fix the problem that Search facet options get cut off 

screen shot 2018-08-01 at 5 03 25 pm

### Checklist


t83714:
It seems still a little bit out. See screenshot below:
I think it's easier to reproduce it on `format` panel.
Probably because we make `format` panel always popup to the left. It make sense as its normal position is at far right end of the all facet buttons and make it positioned to the left can avoid going over the right screen edge.
However, when Both publisher & region is selected and screen is not wider enough, the `format` button will be moved to the second row --- the left end now. It is very easy to go over the left screen edge if the panel is still openned to the left.
image

t83714:
Works well~ 💯 👍

PR #1554, opened 2018-08-02
Made CKAN harvesters run hourly

AlexGilleran:
### What this PR does
Makes the CKAN harvesters run hourly, as per Gordon's instructions. Fortunately CKAN servers are pretty good and our process for harvesting them is pretty low-resources (unlike the XML parsing of CSW). 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -53,6 +53,7 @@
 -   Updated URL of City of Launceston connector.
 -   Keep search text in synch.
 -   Make pagination on mobile responsive.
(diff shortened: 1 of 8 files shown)


maxious:
A potential optimisation is to only ask for records since the last time https://github.com/ckan/ckanext-harvest/blob/f715aa5e1271bbc758de0ee72cc636577351a5aa/ckanext/harvest/harvesters/ckanharvester.py#L198 but let's try this for now 👍

PR #1555, opened 2018-08-03
Issue/1066 Fixed Publisher Out of order

t83714:
### What this PR does

- Updated all connectors to make sure all non-wording chars are removed from the starting of the title
- Name field & id field still remain unchanged --- so we still can do the CSW redirection thing (if we need one day)
- Also fixed an issue with current CSW publisher details fetching logic. If a CSW dataset doesn't have any contact info, the first responsible party will be returned.

Test URL: 
https://issue-1066.dev.magda.io/

To deploy:
- Recrawl all CSW connectors (since we adjusted the CSW connector logic)
- Manually trigger the full reindex. Just realize indexer won't be notified as it won't be looking at `organisation-details` aspect changes. 

### Checklist


AlexGilleran:
Works good, but can we put `cleanOrgTitle` into `magda-typescript-common` ?

t83714:
@AlexGilleran 
Sounds a good idea~ 👍 
Will put into `magda-typescript-common` and make it available through Transformer creation options.

t83714:
@AlexGilleran 
Updated & tested locally
I think it's ready for having a look at again~

PR #1568, opened 2018-08-09
Made the fallback banner url configurable

AlexGilleran:
### What this PR does
Previously we made it so that you could remove the fallback banner... this makes it so you can configure the url that the fallback banner redirects you to, or turn it off completely by not specifying one.

### Checklist

### DIFF
--- a/deploy/helm/magda/charts/web-server/templates/configmap.yaml
+++ b/deploy/helm/magda/charts/web-server/templates/configmap.yaml
@@ -7,5 +7,5 @@ data:
   web.json: '{
     "disableAuthenticationFeatures": {{ .Values.disableAuthenticationFeatures }},
     "baseUrl": {{ .Values.baseUrl | default "/" | quote }},
-    "showFallbackBanner": {{ .Values.showFallbackBanner }}
+    "fallbackUrl": {{ .Values.fallbackUrl }}
   }'
--- a/deploy/helm/magda/charts/web-server/templates/configmap.yaml
+++ b/deploy/helm/magda/charts/web-server/templates/configmap.yaml
@@ -5,7 +5,7 @@ metadata:
 data:
   # When the config map is mounted as a volume, these will be created as files.
   web.json: '{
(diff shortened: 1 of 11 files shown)


deploy/helm/magda/charts/web-server/templates/configmap.yaml
```diff
@@ -7,5 +7,5 @@ data:
   web.json: '{
     "disableAuthenticationFeatures": {{ .Values.disableAuthenticationFeatures }},
     "baseUrl": {{ .Values.baseUrl | default "/" | quote }},
-    "showFallbackBanner": {{ .Values.showFallbackBanner }}
+    "fallbackUrl": {{ .Values.fallbackUrl }}
```
t83714: Do we need double quote around `{{ .Values.fallbackUrl }}`.
i.e. `"{{ .Values.fallbackUrl }}"`
Asked because I got invalid `Invalid JSON config file: /etc/config/config.json` error
and the config map seems missing " around https://data.gov.au
```
web.json:
{ "disableAuthenticationFeatures": false, "baseUrl": "/", "fallbackUrl": https://data.gov.au }
```

deploy/helm/magda/charts/web-server/values.yaml
```diff
@@ -20,4 +20,4 @@ resources: {}
   #  memory: 128Mi
 disableAuthenticationFeatures: false
 service: {}
-showFallbackBanner: true
\ No newline at end of file
+fallbackUrl: "https://data.gov.au"
```
t83714: We probably want to remove this line?
Asked because if the top-level config e.g. `minikube-dev.yml` doesn't have a `fallbackUrl:` line.
The default value from `deploy/helm/magda/charts/web-server/values.yaml` will be used.
So to turn redirection bar off, you will have to make sure the following in top-level config (e.g. `minikube-dev.yml`) :
```
web-server:
  fallbackUrl: ""
```

Issue #1447, opened 2018-07-13
Organisation dataset filter: Desktop and Mobile


tkeuneman:
### Description
As a Magda user
I want to be able to view all data sets of an organisation on desktop as well as mobile
So that I can browse them regardless of the device I am on

### Acceptance Criteria

1. Remove filters from organisation page on mobile. The org filter would be executed in the background but the user does not need to see this. When a user types in a new search, the filter will be cleared. This is an interim step until we have filters back in mobile.
2. The first time a user goes to the search results from an organisation page on desktop, they see the tooltip. 
3. The tooltip is dismissed either by the user clicking away from the tooltip, or by clicking 'Got it'
4. The tooltip is only available on desktop
5. Once the user dismisses the tooltip, the browser remembers this and does not show the tooltip again unless they clear their cookies

### Screenshot / Design / File reference

Designs: https://www.figma.com/file/6V1RKGaGFsE5IJiPG3Tc4u/Organisation

AlexGilleran:
No, still going! 😬

PR #1577, opened 2018-08-10
Added tooltip when using organisation filter for the first time

AlexGilleran:
### What this PR does
All the things in #1447. Principally:
- Removes search filters from mobile
- Makes it so that if you search for something on mobile and you have a search filter active (from following a URL), it strips all the filters
- Makes it so that if you follow a link to datasets by an organisation on desktop, it shows you a tooltip explaining that a filter is active. Once you click away it records that you've seen it in localstorage so that it doesn't show again on that computer
- To make this work, there's a new tooltip component that is used by both the new tooltip and the old data quality one. This works way more dynamically, using javascript to centre itself. I also changed the colour of both tooltips to purple, which was @philipgrimmett approved ™️ .

### Checklist


magda-web-client/src/UI/Tooltip.js
```diff
@@ -1,24 +1,110 @@
...
+
+    componentDidMount() {
+        if (this.props.requireClickToDismiss) {
+            document.addEventListener("mousedown", this.handleClickOutside);
+        }
+        this.adjustOffset();
+    }
+
+    componentDidUpdate() {
+        this.adjustOffset();
```
chloeleichen: change state here cause an infinite loop, eg, if you go to http://localhost:6108/dataset/ds-dap-csiro%3A17050/details?q=water and wait for a few seconds

chloeleichen:
1. new tooltip component looks good & works well
2. >  if you follow a link to datasets by an organisation on desktop, it shows you a tooltip explaining that a filter is active
I can't actually find it, can you upload a screenshot?

AlexGilleran:
Yeah it's not working anymore :'(

AlexGilleran:
Better again now - want to give it one last look @chloeleichen ?

chloeleichen:
Eh I must be looking at the wrong thing :(. I still don't see the tooltip. Maybe you can show me when I'm in tomorrow

PR #1588, opened 2018-08-14
Slowed down cron jobs outside of production.

### LINKED ISSUE #1587: Make dev jobs execute on a slower schedule
### Problem description
Right now prod and dev connector jobs have the same schedule. Because the dev cluster is smaller than prod, this means that the hourly CKAN jobs end up stacking up and taking up all the room in the cluster. They should run much more seldom on dev... like once a week or something.

AlexGilleran:
### What this PR does
Fixes #1587.

Makes it so that the connector cron jobs in dev only run every week... they'll just all queue up at midnight on sunday and then execute then.

This is to alleviate what's happening at the moment where running the CKAN jobs every hour is causing the cluster to fill with slow connector jobs and stopping us from deploying.

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -64,7 +64,7 @@ generate-connector-jobs:
   cache: {}
   script:
     - cd deploy
-    - yarn run generate-connector-jobs --prod true --in ./connector-config --out ./kubernetes/generated/prod --imageVersion=$CI_COMMIT_REF_SLUG --imagePrefix=registry.gitlab.com/magda-data/magda/data61/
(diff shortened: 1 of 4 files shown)


AlexGilleran:
@t83714 right, I've changed it to `0 14 * * 6`, which is 2pm UTC on Saturday, which if you add 10 hours to it should be 12am Sunday in Sydney...

... right? 😬 

Trying to figure out timezones hurts my head.

t83714:
I think you are right~  👍 
I was trying to take 10 hours off the UTC time which was wrong~

PR #1620, opened 2018-08-21
Issue/1617 Fixed dataset / organisation debounced search request not cancel upon URL changes

### LINKED ISSUE #1617: Recent searches not working properly
### Problem description
Picking up a search suggestion from recent searches doesn't work properly

1. Search for ballarat bbq and see the results so that "ballarat bbq" gets added to your recent search
2. Type ba in the search box so you see ballarat bbq as a suggestion.
3. Click on the suggestion

Expected result:
You see the results for ballarat bbq

Actual result:
Page gets refreshed and you see results for ba

[Prod issue.zip](https://github.com/magda-io/magda/files/2301260/Prod.issue.zip)

t83714:
### What this PR does

Fixes #1617 

What happened is the search input box (for both dataset search & organization search ) will always delay search request for 3 seconds.  This delayed timer should be canceled upon browser URL changes --- otherwise, users will be dragged back to old search page if he types something into the search input and goes to another page (e.g. homepage, about page) before the 3 seconds. 

Why monitor through `history.listen` rather than `componentDidUpdate`?
According to the [doc](https://reacttraining.com/react-router/web/api/withRouter): `withRouter does not subscribe to location changes`

### Test URL: 

https://issue-1617.dev.magda.io

@aneesha09 @AlexGilleran @chloeleichen 

It seems a lot of logic running around this search box component. 

Please help to test (especially search related function e.g. Search filter panel) just in case my fix broke something else~ 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 -   Uses the Header component from Design System
 -   Fixed an issue that connector record triming might not fully completed
 -   Fixed an issue that indexer webhook event types not properly setup
+-   Fixed an issue that dataset / organisation debounced search request not cancel upon URL changes
 
 ## 0.0.46
 
--- a/magda-web-client/src/Components/Publisher/PublishersViewer.js
+++ b/magda-web-client/src/Components/Publisher/PublishersViewer.js
@@ -17,6 +17,7 @@ import "./PublishersViewer.css";
 import search from "../../assets/search-dark.svg";
 import { Medium } from "../../UI/Responsive";
 import AUpageAlert from "../../pancake/react/page-alerts";
+import { withRouter } from "react-router-dom";
 
 class PublishersViewer extends Component {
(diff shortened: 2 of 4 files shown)


t83714:
@chloeleichen 
It's ready to look at again~
I believe the issue you found can be triggered by a successful search. And should happen after every keyword has been saved into local storage.
if you issue can be reproduced in a different way, please let me know as we must have a different bug around there 😄

chloeleichen:
It works, except one thing, when you click on a suggestion and updated search url, it also needs to update the search text inside the search box. 

To reproduce, 
1. search for water, get to result page
2. change search text to w, suggestion comes up, click "water"
3. search starts, shows results for "water" but search text is still "w". 

Search text in search box should reflect whatever is being searched at the moment so it should be updated when url updates

Issue #1599, opened 2018-08-15
API Documentation - Authorization API


AlexGilleran:
### Problem description
Create apidoc documentation for the authorization api.

PR #1630, opened 2018-08-24
Document public portions of authorization API (#1599)

nahidakbar:
### What this PR does

Fixes #<insert issue number here if relevant>

### Checklist


AlexGilleran:
Looks fine in apiDoc, but because there's no return types specified the swagger isn't good enough to generate a client. Do you know of a good way to do that with api doc and swagger?

nahidakbar:
@AlexGilleran fixed up swagger warnings. No there isn't a nice way through apidocjs. If this is important for our users, we can look into it separately.

PR #1688, opened 2018-09-05
Add No results label when there are no results in organisation and lo…

### LINKED ISSUE #856: Search Filters: There should be feedback when no publisher/format found
### Problem description

When I search for a publisher (also for Format and Location filters), if there is no publisher found for that particular query, there isn't any feedback for it. Maybe a little message below saying "no results found" would be good.

Try typing in a publisher in the publisher drop down filter that doesn't exist

image

### maybe have something like this:

image

When the user clicks away, clear the search

nahidakbar:
…cation facets. Fixes #856

### What this PR does

Fixes #856

### Checklist

### DIFF
--- a/magda-web-client/src/Components/SearchFacets/FacetRegion.js
+++ b/magda-web-client/src/Components/SearchFacets/FacetRegion.js
@@ -169,6 +169,7 @@ class FacetRegion extends Component {
             <div className="facet-body facet-region">
                 <RegionSearchBox
                     renderOption={this.renderOption}
+                    renderNoOption={this.renderNoOption}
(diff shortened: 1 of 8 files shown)


magda-web-client/src/Components/SearchFacets/FacetRegion.js
```diff
@@ -122,6 +123,19 @@ class FacetRegion extends Component {
         );
     }
 
+    renderNoOption() {
```
AlexGilleran: Can we pull this out into a separate little component seeing how we're using it in two places?

nahidakbar:
@AlexGilleran

PR #1709, opened 2018-09-12
Issue/1695 Allow scss variables (e.g. Color ) to be changed in the runtime

### LINKED ISSUE #1695: Further Investigate the colour scheme configurable solution
### Problem description

Related to #1640 

I made some progress on issue #1640 but failed to work out a concept-proof demo in our dev cluster.
Therefore, the scope of this issue was to continue the investigation so that we can have a workable demo in google cloud (to prove it works in the production environment )

The current problem is:
- `pancake-sass` v2.0.0 's upstream lib `sass` introduced an unnecessary dependency `fsevents` that is not available in our docker environment. This will cause our gitlab pipeline stops at `dockerize:ui` step
- Upgraded `pancake-sass` to v2.1.0 and it removed the dependency `sass`. But its compile result seems buggy (yet to find out the reason)

Possible solution:
- Stay on v2.0.0 and modify our `create-docker-context-for-node-component` script to skip `fsevents`

~Estimated: roughly 2 days~
I will probably spend some time to fix the existing pancake issue then spend more on the `turn the solution into a Job` plan. - Re-estimated as `5`

t83714:
### What this PR does

Fixes #1695 

Allow scss variables to be changed in the runtime. We can use this to alter the UI color etc.
Summary of the changes I've done:
- Created a new docker image.
  - Packed with pancake & node-sass for scss file re-compile
  - Compile result will always be sent to `content API`: `/api/v0/content/stylesheet.css`
  - Read vars (that's required to be overwritten. By default, should be '{}' or empty string) from configMap `scss-compiler-config`
   - This config map is part of `contentAPI` chart
- Added a new job to `contentAPI` chart. This job is hooked with helm`post-install,post-upgrade` hook
- Modified `content API` so it can support CSS type content
- Overwritten `create-react-app` setting. So, css will be pointed to `/api/v0/content/stylesheet.css` for production built `html`. Development version remains unchanged.
- Create a new command line toole `set-scss-vars` for updating scss. Run `yarn set-scss-vars --help` to look at help.
  - Will create a new job for compiling and updating css
  - The tool will monitor the job excution progress and print status every 10 seconds
  - It will delete the job once it's completed 

### Checklist


magda-scss-compiler/setup.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+cd /usr/src/app
+# delete node-sass comes with pancake-sass 2.1.0 T_T
+rm -Rf node_modules/@gov.au/pancake-sass/node_modules/node-sass
+echo "Rebuild node-sass"
+npm install node-sass@4.9.3 stringstream@1.0.0
```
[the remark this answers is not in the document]
t83714: @AlexGilleran 
Good idea~ The node-sass reinstallation is only for pulling Linux version binary to docker.
`stringstream` put there just for some wired version issue.
I think, instead of reinstalling this two, I might try something else (that doesn't require specifying a version) to see if it works.

t83714:
@AlexGilleran 
- The underline is gone now (after the change to `always` pull image)
image
- The loop issue seems an existing issue --- you can easily reproduce it on `Master` branch. (Afterall, there is code changed to the client module. I actually implemented a separate (and different) recompilation process in a separate docker container. I only turned off one option for `HtmlWebpackPlugin` -- it's for HTML template.)
- The new docker image carries the node-scss. so that job won't need re-build node-scss and re-run pancake.
- No version number is used. I used npm rebuild.

I think it's ready to have another look now~

deploy/helm/magda/charts/content-api/templates/init-scss-compiler.yaml
```diff
@@ -16,9 +16,11 @@ spec:
       containers:
       - name: init-scss-compiler
         image: "{{ .Values.image.repository | default .Values.global.image.repository }}/magda-scss-compiler:{{ .Values.image.tag | default .Values.global.image.tag }}"
-        imagePullPolicy: {{ .Values.image.pullPolicy | default .Values.global.image.pullPolicy }}
-        command:
-            - "/usr/src/app/component/setup.sh"
+        imagePullPolicy: Always
```
AlexGilleran: I think this should stay as it is - with the config that was here before it'll be "always" in dev/preview but "ifnotpresent" in prod, which is better.

Wasn't it the job that had the problem with the pull policy?

t83714: You're right~ 👍 I will change it back~

t83714:
@AlexGilleran 
Pull policy is set to `imagePullPolicy: {{ .Values.image.pullPolicy | default .Values.global.image.pullPolicy }}` now~ Thanks for the tips 👍

PR #1715, opened 2018-09-13
Added contact point to dataset page

### LINKED ISSUE #1660: Display contact point details on dataset page
As an internal Magda user, I want to see who the contact point for a dataset is so I can ask them a question without having to send an email through the Magda system.

### Problem description
For internal magdas, it's possible that the person who finds the data won't have access to it, so they'll need to ask the person who looks after the data to give them access. Allowing to see who that person is (instead of hiding it behind the "Ask a question" button means that they're able to choose how they approach that person, and use a more secure means of doing so (i.e. not an email going through mailgun or another SMTP server provider).

~This might not be as appropriate for external magdas, so it should be toggleable through config.~

### Acceptance Criteria
- Dataset contact point information is visible on UI
- ~When turned off, dataset contact point information is not visible.~
- Track how many emails are sent through the system
- Hide the contact information behind a click and track the number of clicks by dataset to know how many times users actually try to see the contact information to contact the custodian outside the system

AlexGilleran:
### What this PR does

Fixes #1660 

- Shows the `dcat-dataset-strings` `contactPoint` on the dataset page
- Makes it so you have to click to reveal it, Gumtree-style
- Moves the "reveal" button out into its own component so that both the contact point and the description can use it
- Tracks that click with Google Analytics so we can see who is being contacted
- Renders the contactPoint with markdown so that emails get highlighted with mailto even if they're embedded in a bunch of other cruft

@philipgrimmett should probably take a look at this too... interaction looks like so (ignore the loop, it doesn't go back to "Click to reveal" if you click on the email:
contactpoint reveal

### Checklist


magda-web-client/src/UI/ToggleButton.scss
```diff
@@ -0,0 +1,17 @@
...
+    font-size: 1rem;
+    color: $AU-color-foreground-action;
+    background: transparent;
+    border: none;
+    cursor: pointer;
+    position: relative;
+    padding-left: 0px;
+
+    &:hover {
+        focus {
```
nahidakbar: Does this need a colon ?

PR #1720, opened 2018-09-17
Made new dataset suggestion form move around according to search score

### LINKED ISSUE #1521: 'Suggest a new dataset' in search results: move to further down the results page
### Problem description
The 'Suggest a new dataset' is showing as the second result in every search that I run. If it continues to do so, it will either become part of the wallpaper to the user, or worse, people will start submitting requests even if there's a dataset that matches further on the page. 
It should show based on the relevance of the keywords to the datasets. 
- If it's set as relevance, lower the relevance threshold. 
- If it's not set as relevance and has just been done to show as the second result, change it so that it is based on relevance.

Run a search on prod

screen shot 2018-07-27 at 7 44 58 pm

AlexGilleran:
### What this PR does

Fixes #1521 

So this changes the search results to put the dataset suggestion form between the last dataset to score over a certain threshold and above the first one to score below it - I had a play around and `65` seemed like a good sweet spot, it's just about where results go from obviously-relevant to "ok I can kinda see why that came up"-relevant. It can be changed on the fly though.

- When searching for nothing (e.g. when you click the "Datasets" link), the suggest box should show up at the end of the first page
- When searching for something with no high-scoring results (e.g. "I am a gigantic elephant") the suggest box should show up at the start of the first page
- When searching for something where the first result under the score threshold is the first one in a page, the suggest box should show up at the end of the previous page rather than at the top of the current one.

Live at https://1521-move-suggestion.dev.magda.io/

### Checklist


AlexGilleran:
When searching for nothing (i.e. "Datasets" link), suggest link should be at bottom of the first page.

PR #1724, opened 2018-09-20
Issue/1655 Allow cluster to be protected by a password

### LINKED ISSUE #1655: Put instances behind a password
As a user of an internal Magda, I don't want everyone on the internet to be able to see my data catalogue.

### Problem description
Neither of our internal Magda partners want their magda instances to be freely visible, so we need some way to restrict them.

The ideal end state is that they'll live behind the firewall and/or have SSO auth, but for now we should do something quick and easy like put everything behind basic HTTP auth. Which means we need to built it in _somewhere_, most likely in the Gateway.

### Acceptance Criteria
- It's possible to set up Magda so that the entire thing requires a password (not just the UI, but access to APIs etc)
- The password is securely stored in kubernetes secret
- This secret setup is integrated into the secret creation script
- Nothing can be accessed without the password

t83714:
### What this PR does

Fixes #1655 

Changes:
- Make public/whoami return 200 status. Therefore, login status won't be reset by 401 status. This is particularly for google chrome
- Make sure `fetch` call initiated with option `credentials: "same-origin"`. So safari will send Auth headers.
- This feature can be enabled by enableWebAccessControl = true in top level helm config (under gateway section)
- Can use `create-secrete` tool to setup the secrets

Test site:
https://issue-1655.dev.magda.io/

user name: `magda`
password: `Faegaehie7chaiye`

### Checklist


scripts/create-secrets/askQuestions.js
```diff
@@ -326,6 +326,86 @@ const questions = [
         validate: input =>
             trim(input).length ? true : "secret cannot be empty!"
     },
+    {
+        type: "list",
+        name: "use-web-access-secret",
+        message:
+            "Do you need to setup the username & password used to access Magda web interface?",
```
AlexGilleran: This is a bit confusing because there's two sets of username and password - maybe specifically say "Do you want to setup HTTP Basic authentication" ?

magda-web-client/src/actions/userManagementActions.js
```diff
@@ -16,15 +16,25 @@ export function requestWhoAmI() {
         });
 
         fetch(config.authApiUrl + "users/whoami", {
-            credentials: "include"
+            credentials: "same-origin"
```
AlexGilleran: This is going to break logging into dev.magda.io from other instances :(

t83714:
@AlexGilleran 
when you have a chance, could you please have another look and let me know if further changes are required? 👍

PR #1729, opened 2018-09-21
Upload and run CSV connector for spreadsheet data

### LINKED ISSUE #1653: Build CSV Connector
As a Magda core team member, I want to be able to create a data catalogue using a CSV file 
So that I can load CSV data from government agencies into internal Magda

### Problem description
Both Agriculture and Environment have Excel spreadsheets that contain their data catalogues. The important columns are:

- Title
- Description
- Share drive link
- Custodian email address

We need a tool that can run, read through the spreadsheet, and create very basic records with this metadata.

The contents of these spreadsheets are not massively secret like the actual data they describe, but they are confidential enough that we can't just have them out in the open at a URL to be picked up, so that's a concern that needs to be addressed. Possibly rather than configure it as a job, we configure it as a deployment that accepts the spreadsheet as a `POST`?

We have to produce these for two entirely different spreadsheets (and probably more in the future) so this needs to be super configurable in terms of how it crawls through, maps columns etc. also.

### Acceptance Criteria
- It's possible to take one of these spreadsheets that's been provided to us, load it in to a fresh Magda instance and see that the records described therein are searchable
- It's possible to download from the share drive links assuming you have access and one was present in the spreadsheet
- It's possible to see the email address of the custodian assuming that was present
- One connector can be applied to both spreadsheets, and be extended to others
- If a new CSV is uploaded, it should override the data catalogue

nahidakbar:
### What this PR does

Fixes #1653

### Checklist


AlexGilleran:
Actually I could just change the column name, duh

AlexGilleran:
... nevermind!

PR #1744, opened 2018-09-27
Made CORS work for localhost connecting to dev.magda.io again

### LINKED ISSUE #1743: Can't access dev.magda.io API when running from localhost or preview
### Problem description
Because we're returning `*` as the origin on dev but also including credentials in the fetch request, chrome is getting v. cranky:

```
Failed to load https://dev.magda.io/api/v0/auth/users/whoami: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. Origin 'http://localhost:6108' is therefore not allowed access.
```

We should make it so that in dev (and definitely NOT in prod), the `origin` header gets reflected back as the Access-Control-Allow-Origin header.

AlexGilleran:
### What this PR does
Makes the gateway reflect the CORS origin for preview and dev so that we can continue using it as the API when running locally or in ui-only previews. This involved changing the config and also stripping the CORS plugin out of the scala APIs.

Fixes #1743 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 -   Made the CSW connector look for names for service-based datasets in another place
 -   Made CSW connector gracefully handle datasets without ids instead of crashing
 -   Fixed Gitlab UI only preview failed to download main CSS file
+-   Removed CORS handling from Scala APIs, should be entirely handled by the gateway.
 
 ## 0.0.48
 
--- a/deploy/helm/magda-dev.yml
+++ b/deploy/helm/magda-dev.yml
@@ -29,6 +29,7 @@ gateway:
     googleClientId: "275237095477-f7ej2gsvbl2alb8bcqcn7r5jk0ur719p.apps.googleusercontent.com"
   cors:
     credentials: true
+    origin: true
   helmet:
     frameguard: false
   csp:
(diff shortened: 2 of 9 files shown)


t83714:
@AlexGilleran 
- Tested using a local web-client connect to the gitlab deployment.
It looks good 👍  And `access-control-allow-origin` now show as my localhost
image
- Also tested put the instance behind a password. Didn't find any issues for both Chrome & Safari
- Only issue is when you connect to the gitlab deployment from a local web-client, the map preview seems not work:
```
Refused to display 'https://issue-1743-make-cors-header-reflect-origin-in-dev.dev.magda.io/preview-map/#mode=preview&hideExplorerPanel=1' in a frame because it set 'X-Frame-Options' to 'sameorigin'.
```
But I guess we set that option for security reason, right?

t83714:
By the way, the instance is now behinds a password. if you want to have a try, the username is `magda` & password is `Faegaehie7chaiye `

AlexGilleran:
@t83714 I've turned frameguard off on previews too, so now you should be able to run on localhost and connect to the preview URL with maps working again 👍

t83714:
Tested and preview map works well at local now 👍 
Merging~

Issue #1713, opened 2018-09-13
Make Google Analytics id configurable


AlexGilleran:
As a magda admin, I want to be able to set my Google Analytics id.

### Problem description
We use Google Analytics in a bunch of places for events and will probably end up using it more - we need to be able to set the analytics id so that instances aren't sending their events to our and dta's google analytics.

### Acceptance criteria
- Google analytics id is available to be set in helm
- Google analytics events can be observed in the google analytics account whos id has been set
- You can set multiple ids (e.g. in search.data.gov.au we send events to DTA and to us)

PR #1754, opened 2018-10-02
[WIP] Make Google Analytics tracker(s) configurable

adamzerella:
### What this PR does

WIP for #1713 allowing a data admin to access the Google Analytics identifier set in Helm, assuming `admin-api` is a suitable module. 

`admin-api` helm chart is configured from static value --> `magda-admin-api` accepts helm input --> `magda-admin-api` offers API endpoint `/gapi`.

### Checklist


AlexGilleran:
Great to see you again @adamzerella 😄.

> assuming admin-api is a suitable module.

It _kinda_ makes sense, but I think it'd be easier to just have it supplied to the `magda-web-server` module instead and then passed to the client with the rest of the config, seeing how it's only the client that's going to use it.

adamzerella:
@AlexGilleran what's crackin' :wave: 

Yeah I think the wording of the issue threw me off a little, I've moved the code into the `web-server` module and added a test for it. Haven't tested locally yet, hopefully will soon

AlexGilleran:
OK that's a bit better, but rather than provide an endpoint specifically for this I'd just put it in the `webServerConfig` that the client javascript (`magda-web-client`) is already grabbing on startup.

Also keep in mind that it needs to be connected up in `magda-web-client` - right now that's at https://github.com/magda-io/magda/blob/master/magda-web-client/public/assets/third-party.js#L34 but it should probably done better - we should just be downloading the Google Analytics code once and configuring it with one or more Google Analytics ids as passed through the config from the web server pod rather than relying on pasting code snippets.

adamzerella:
@AlexGilleran I've swapped around this PR a little, the `webServerConfig` config now accepts a list of tracker ids that are passed into `react-ga`; This module seems to accept an array of tracker ids: https://github.com/react-ga/react-ga/blob/92205207684e5df19c172f93f7a9e1d0b3edb179/types/index.d.ts#L85

I've refactored and removed the current implementation of sending `ga` events in favour of using `react-ga`. Still needs to be tested on deployment as I'm not sure a `.yml` list value will translate to js well.

Let me know if you prefer the `magda-web-client` to handle ga / ga events another way

magda-web-client/src/AppContainer.js
```diff
@@ -23,6 +24,28 @@ import "./AppContainer.css";
 class AppContainer extends React.Component {
     componentDidMount() {
         this.props.requestWhoAmI();
+
+        const _trackers = [];
+
+        // Create GA trackers from the list provided to us
+        config.gapiId
```
adamzerella: Should clean this up a little and discuss if better places to put this

magda-web-client/public/assets/third-party.js
```diff
@@ -31,8 +31,6 @@
     "ga"
 );
```
AlexGilleran: You can get rid of the google analytics snippet above too :)

magda-web-client/src/AppContainer.js
AlexGilleran: Yeah I think just refactor it into some kind of `initGoogleAnalytics.js` file that exports a function that does this.

I'm a bit confused by the use of the `?` and the `[config.gapiId]` thing here though - isn't `config.gapiId` already an array?

magda-web-server/src/index.ts
```diff
@@ -168,7 +173,8 @@ const webServerConfig = {
     ),
     fallbackUrl: argv.fallbackUrl,
     datasetSearchSuggestionScoreThreshold:
-        argv.datasetSearchSuggestionScoreThreshold
+        argv.datasetSearchSuggestionScoreThreshold,
+    gapiId: argv.gapiId
```
AlexGilleran: It'd probably be better to call it `gapiIds` because it's an array :)

deploy/helm/magda/values.yaml
```diff
@@ -13,6 +13,8 @@ global:
   logLevel: INFO
   noDbAuth: false
   namespace: default
+  gapiId: 
+    - UA-92539508-3
```
AlexGilleran: This should probably be `[]` by default then we can add this and the other id in https://github.com/magda-io/magda-config-data.gov.au :)

adamzerella:
@AlexGilleran Updated the PR from your notes, naming convention for the analytics module is in the air.

Maybe making `ga.js` an actual module and tidying it up to something similar to: https://github.com/react-ga/react-ga/blob/master/demo/app/index.jsx is the next step. It seems that we'd need to initialise the component before every call. 

> GA must be initialized using this function before any of the other tracking functions will record any data. The values are checked and sent through to the ga('create', ... call.

From: https://github.com/react-ga/react-ga#reactgainitializegatrackingid-options

Side note, I'll open another PR to add `gapIds` to https://github.com/magda-io/magda-config-data.gov.au/blob/master/config.yaml

AlexGilleran:
Made some updates to this but I can't put the comments into here or raise a separate PR, closing this one

AlexGilleran:
OK new PR is up at https://github.com/magda-io/magda/pull/1783/files.

@adamzerella do you want to take a look and see if you're ok with everything I did? Essentially boiled down to:
- Making sure the trackers always got passed to ReactGA
- Smoothing out the helm stuff

Thanks again! 😁

AlexGilleran:
Definitely! 💯

PR #1758, opened 2018-10-03
Issue/1710 Make email template configurable

### LINKED ISSUE #1710: Make email text configurable
As a (non data.gov.au) Magda admin, I want Magda to send emails that are configured with text I approve of and don't have any reference to data.gov.au in the text or logo.

### Problem description
Currently the email templates are hard-coded into the correspondence API, and they're very very data.gov.au specific.

We want to make it so that the admin can write simple markdown-based templates and have these rendered into emails by the correspondence API. This should involve a very simple template engine (something like moustache?). This might mean getting rid of the current pug templates which are super complex. So it'll work something like:

- Correspondence api gets called with variables
- Render variables into markdown template
- Render markdown to HTML and insert it into nicely styled email
- SEND

### Acceptance Criteria
- It's possible to send an email via Magda that has no reference to data.gov.au in it
- It's possible to change the text of the email by configuration - this doesn't require a UI yet, only some kind of API-capable way of changing it

t83714:
### What this PR does

Fixes #1710 

Changes:
- ~Content API can now accept any type of content without modifying its code~
  - ~The new behavior only works for URL pattern `/v0/commonAssets/*`~
- Move `magda-correspondence-api` template to ~`emailTpls`~ `emailTemplates` directory
  - When `correspondence-api` starts up, all files in this folder will be `synced` to ~`contentAPI`/commonAssets/emailTpls/*~ `contentAPI`/emailTemplates/*
    - If a file already exists, it won't be replaced
    - If sync failed, `correspondence-api` pod will restart
- Replace Template engine with [mustache](https://github.com/janl/mustache.js)
  - re-wrote the `mustache` tpl
  - Implement customise function so that:
```html
{{#inlineImg}}
    <img class="image" src="emailTpls/assets/Logo.jpg" alt="Data gov logo" />
{{/inlineImg}}
``` 
will become:
```
<img class="image" src="cid:inlineImg_0_968719913851184" alt="Data gov logo" />
```
The program will automatically try to:
- retrieve `emailTpls/assets/Logo.jpg` from ~`contentAPI/v0/commonAssets/emailTpls/assets/Logo.jpg`~ `contentAPI/v0/emailTemplates/assets/Logo.jpg`
- encode it
- inline attached it

### Test:
The current branch has been deployed to namespace: `issue-1710`
The current url is: 
https://issue-1710.dev.magda.io
There is no data at this moment but you can test using the footer link (Suggest a Dataset)
The content-api NODE_ENV has been changed to `development`. 
You can get `magada session token` from `/public/jwt`

To change the email template, you need to update them from conentAPI.
Their URLs are:
- ~`/commonAssets/emailTpls/request.html`~ `/emailTemplates/request.html`
- ~`/commonAssets/emailTpls/question.html`~ `/emailTemplates/question.html`

### Checklist


magda-content-api/src/createApiRouter.ts
```diff
@@ -15,10 +21,138 @@ export default function createApiRouter(options: ApiRouterOptions) {
 
     const router: express.Router = express.Router();
 
+    const ADMIN = mustBeAdmin(options.authApiUrl, options.jwtSecret);
+
     router.get("/healthz", function(req, res, next) {
         res.status(200).send("OK");
     });
 
+    function parseUrlPath(pathString: string) {
```
nahidakbar: Is /commonAssets necessary?. Custom assets can be configured by use route option. See csv example in content.ts. With that, separate route can be used for different types of resources as well.

magda-correspondence-api/src/CContentApiDirMapper.ts
```diff
@@ -0,0 +1,92 @@
+import buildJwt from "@magda/typescript-common/dist/session/buildJwt";
+import * as rp from "request-promise-native";
+import * as mime from "mime-types";
+import * as recursiveReadDir from "recursive-readdir";
+import * as fse from "fs-extra";
+import * as path from "path";
+import * as typeis from "type-is";
+
+class CContentApiDirMapper {
```
nahidakbar: Why not use content API exclusively? Instead of maintaining a separate file system thing.

magda-correspondence-api/src/CEmailTplRender.ts
```diff
@@ -0,0 +1,82 @@
+import { Attachment } from "./SMTPMailer";
+import CContentApiDirMapper from "./CContentApiDirMapper";
```
nahidakbar: What does the first C stand for in CContent* and CEmail*? Just a personal preference but would rather prefer to see descriptive names for things instead of short forms.

magda-correspondence-api/src/CEmailTplRender.ts
```diff
@@ -0,0 +1,82 @@
...
+            cid
+        });
+    }
+    return content.replace(
+        /(<img\s+[^>]*src=["'])([^"']*)(['"])/i,
+        `$1${"cid:" + cid}$3`
+    );
+}
+
+export default class CEmailTplRender {
```
nahidakbar: Just a pet peeve but could we please use descriptive names? E.g. Template instead of Tpl.

magda-content-api/src/createApiRouter.ts
t83714: 😄 the `/commonAssets` is not for enforcing a particular route. It's more for providing a little bit different API without impact existing functions.
The logic behind `/commonAssets/*` are aiming for more generic resource storing purposes. The differences are:
- Both GET / POST are using the same resource name.
  - Existing API you need to GET `style.css` but post to `style`
- Handle any type of file without modifying content API or config
- Allow `/` in the resource name. So I reference `emailsTpl/assets/a.png` in email templates

magda-correspondence-api/src/CContentApiDirMapper.ts
t83714: Yes. It's fully rely on `content API`.
The `CContentApiDirMapper` is more for auto-deployment. 
It only uploads all local files in `emailTpls` to content API --- suppose new deployment or production site upgrade have no those resources in the beginning.
`CContentApiDirMapper` will make sure if any of those files missing, they will be auto - sync to `content API`.
The actual resource retrieving is all from `content API` directly.
It will also help future deployment.
Let's say we modify the default template and added a more image file.
The deployment doesn't require manual uploading the new file to the `content API` of the existing site.

magda-correspondence-api/src/CEmailTplRender.ts
t83714: Ah~ I think I am a little bit old school on this~ 😄
`C` indicates it's a class that requires initialization.
If you think it's confusing, how about I just remove the `C`?

magda-correspondence-api/src/CEmailTplRender.ts
t83714: I think it's a good idea~
Have updated the name to `CEmailTemplateRender`

magda-correspondence-api/src/CEmailTplRender.ts
nahidakbar: Hmmm... never seen it before.
Probably remove. If it starts with a capital letter, it should be a class.

magda-correspondence-api/src/CEmailTplRender.ts
nahidakbar: Thanks. Did a quick check and there are a few other instances of *Tpls* as well.

magda-content-api/src/createApiRouter.ts
t83714: - Just want to make it less job to do when uploading content. It will save some code every time you want to upload to content API. Auto figure it out from the extension name could minimize the chance of sending the wrong mime-type
- I agree. It's just a randomly picked name. We probably need a better name 😄 Here just try to make things auto handled and minimize the work when applying the changes. If we want to go for the more configurable direction, we then probably needs some flexibility. However, we probably do need to have some plan with naming.

magda-correspondence-api/src/CContentApiDirMapper.ts
t83714: Yes. We can put it to `content DB migrator`. However, if we decide to put it to `content DB migrator`, we have to update it everything we have some changes. And it could including some images files (think about future UI changes).
Right there is no extra work needs to be done and all new files will be auto-included. Was trying to avoid some maintenance troubles (for whoever has to work on it) 😄

magda-correspondence-api/src/CEmailTplRender.ts
t83714: Removed C 👍

magda-correspondence-api/src/CEmailTplRender.ts
t83714: Removed Tpls
And move assets to `emailTemplates` 👍

magda-content-api/src/createApiRouter.ts
AlexGilleran: Yeah this sounds more like functionality that should be added to the regular API but I don't think I have the full picture, I'll ask Jacky after lunch 🤔

t83714:
Hi @nahidakbar & @AlexGilleran ,
Had a play on https://issue-1710.dev.magda.io
It seems it runs ok now. Could you please have another look?
To test the email, you could use the `Suggest a dataset` link in the footer.

magda-content-api/src/createApiRouter.ts
```diff
@@ -45,64 +71,63 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+    async function getContent(req: any, res: any) {
+        const requestContentId = req.params.contentId;
+        const requestFormat = req.params.format;
         try {
-            const maybeContent = await database.getContentById(contentId);
-            const content = maybeContent.caseOf({
-                nothing: () => null,
-                just: x => x
-            });
+            const contentPromise = await database.getContentById(
```
AlexGilleran: I've rewritten this part to get rid of the mutability because I wasn't sure how to put it into a comment, let me know if it's ok.

nahidakbar: Ok. This does look cleaner =)

t83714:
@AlexGilleran The followings have been done:
- Comments added for `ContentApiDirMapper.ts`
- Rename logo name & replace logos with Magda logos
- Remove DGA related content from email templates
- A PR has been created for config-data.gov.au: https://github.com/magda-io/magda-config-data.gov.au/pull/5
I think it's ready to have another look (test site: https://issue-1710.dev.magda.io)
Please let me know if further changes are required~

PR #1765, opened 2018-10-04
Issue/1671 Better way of retrieving hook actor status

### LINKED ISSUE #1671: Investigation - Registry API: Hook service / Get Hook Endpoint may report timeout when Actors are very busy
### Problem description

We recently added the feature of making Actor status available from hook info result returned from Hook service endpoints (see PR https://github.com/magda-io/magda/pull/1619).

The actor status retrieve process requires sending a `getStatus` message to the actor and wait for the response.

When the actor is too busy, actor might not be able to return the response timely (within 5 seconds) and will produce a timeout exception in the log (see log from the attachment).

There's a number of ways we might solve this:
#### Solution 1
- Incease the timeout to 30 seconds
- And this timeout parameter should be adjustable from Config
- The `isRunning` & `isProcessing` fields should both be `String` (Default value is `Unknown`)
  - All available values are:
   - `Unknown`
   - `Yes`
   - `No`
   - `Timeout`
- When `getStatus` query timeout, the value for the two fields above should be "Timeout"

#### Solution 2
Have a process occasionally query the actor and get its status. When the API is invoked it gets the last status this actor got and the time it got it.

#### Solution 3
Have the webhook actor report send a message to some kind of listener whenever it changes, which stores the last state it received and returns it when the API is invoked

This an investigation to see what the best solution is.

Probably reduce the CPU resource and increase the webhook actor load will see this error.

[registry.log](https://github.com/magda-io/magda/files/2343928/registry.log)

t83714:
### What this PR does

Fixes #1671 

Trying to go for the option 3. 
i.e.
> Have the webhook actor report send a message to some kind of listener whenever it changes, which stores the last state it received and returns it when the API is invoked 

What I have done is to make all hook actors to report their status changes to their parent actor. Parent actor save the status in a local map. However, we still need to send message to the parent (main) actor to retrieve the status.

I couldn't find a way to directly access the local MAP structure held by the main actor. The `ActorRef` seems only allows to interacte with the actor via messages. If you know a `thread safe` way to access actor's data directly, please let me know~

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## 0.0.50
 
 -   Delete registry API /records/{recordId}/aspects endpoint
+-   Hook actors will report its status to its parent actor when changes
 
 ## 0.0.49
 
--- a/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
+++ b/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/HooksService.scala
@@ -76,10 +76,10 @@ class HooksService(config: Config, webHookActor: ActorRef, authClient: AuthApiCl
           HookPersistence.getAll(session)
         }
 
-        implicit val timeout = Timeout(5 seconds)
+        implicit val timeout = Timeout(30 seconds)
 
         hooks.map { hook =>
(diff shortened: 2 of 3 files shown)


t83714:
Changes are made for all previous comments. I think it's ready to look at again~ 👍

PR #1789, opened 2018-10-17
Issue/1762 Make header menu items come from content api

### LINKED ISSUE #1762: Make header details come from content api
### Problem description
Currently the details of what shows up in the header are hard-coded into the client. We should make this come from the content api so we can change it. We've already made the logo work like this, so we've just got to change the links at the side.

### Acceptance Criteria
- Details of links on the header come from content api
- There's reasonable defaults for new installs (e.g. Datasets, Organisations, About)
- There's a plan for migrating search.data.gov.au (this in particularly has its 🦄 "Community" link that doesn't make sense to have as a default)

nahidakbar:
### What this PR does

Fixes #1762

### Checklist


magda-content-api/src/createApiRouter.ts
```diff
@@ -35,14 +45,52 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+     *            "length": ...
      *        },
      *        ...
      *    ]
      */
-    router.get("/all", ADMIN, async function(req, res) {
-        res.json(await database.getContentSummary());
+    router.get("/all", USER, async function(req, res) {
+        // get summary
+        let all = await database.getContentSummary();
```
AlexGilleran: Can we do this filtering with an SQL query, or does it require something that the wildcard library does that SQL pattern matching doesn't do?

Then we could just hit the records we need, and do the inlining in the same trip.

magda-content-api/src/createApiRouter.ts
```diff
@@ -26,7 +36,7 @@ export default function createApiRouter(options: ApiRouterOptions) {
 
     /**
      * @apiGroup Content
-     * @api {post} /v0/content/:contentId Get All
+     * @api {post} /v0/content/all Get All
      * @apiDescription Get a list of content items and their type.
      * You must be an admin for this.
      *
```
AlexGilleran: Can we document the filtering/inlining and how it's intended to be used here? Looking at the way the frontend uses it looks pretty powerful but I'm not exactly sure how it works api-wise.

magda-content-api/src/filter.ts
```diff
@@ -0,0 +1,22 @@
+const wildcard = require("wildcard");
+
+export function simpleFilter(items: any[], field: any, filters: any) {
```
AlexGilleran: If we keep this function it could use some jsdoc

magda-content-api/src/createApiRouter.ts
nahidakbar: Can I please keep it 🙏 ? My rationale:

- Right now DB is just a key value store.
- Want to move DB to an even sourcing model with caching all content in memory. This works perfectly for that.

PR #1804, opened 2018-10-24
#1760 Home page items are configurable

nahidakbar:
### What this PR does

Fixes #1760

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.
-   [x] I've linked this PR to an issue in ZenHub (core dev team only)

Issue #1782, opened 2018-10-15
Remove "suggest a dataset" and "privacy policy" pages from internal magdas


AlexGilleran:
### Problem Description
The agriculture and environment internal magdas have no need for privacy policies or suggest a dataset pages - we should remove the links to these. Removing the pages themselves isn't really necessary.

### Acceptance Criteria
- The environment magda has no links to the privacy policy or suggest a dataset pages
- The agriculture magda has no links to the privacy policy or suggest a dataset pages

PR #1809, opened 2018-10-25
Issue/1756 Make footer configurable

### LINKED ISSUE #1756: Make footer content come from content-api
### Problem description
Currently the footer is driven by configuration, but that configuration is hard-coded into the client. We should move this to the content api so it can be edited. Note that this will probably mean adding to what is driven by configuration.

Most likely the entire footer configuration should be stored in one row in the content db as a blob of JSON, although we could change this.

### Acceptance Criteria
- Desktop Footer links and headers are configurable
- Mobile footer links and headers are configurable
- Vanity links ("developed by, operated by") links in footer are configurable
- There's a plan for deploying this to search.data.gov.au without any loss of content - presumably by executing some SQL beforehand to pre-load it with the old values
- New installs get set to reasonable defaults - this means that the "Feedback" link shouldn't be there because this depends on Data.gov.au's Zendesk

t83714:
### What this PR does

Fixes #1756 

Changes:
- this PR merged https://github.com/magda-io/magda/pull/1804 as it includes the latest changes for content API
- Move fetch content to `AppContainer` in order to cover footer as well
- fixed: error every child should have a key in `HeaderNav`
- Configurable footer related changes

Test site available at:

https://issue-1756.dev.magda.io/

To test DTA config, please remove any content with id starts with `footer/`. And then use dta upgrade SQL to import DTA config (Relevant PR: https://github.com/magda-io/magda-config-data.gov.au/pull/11).

### Checklist


magda-migrator-content-db/sql/V1_8__footer_default_config.sql
```diff
@@ -0,0 +1,43 @@
+INSERT INTO "public"."content" ("id", "type", "content") VALUES ('footer/navigation/categories/medium/data.gov.au', 'application/json', '{
+  "label": "Data.gov.au",
```
nahidakbar: Internal magda shouldn't reference data.gov.au

magda-migrator-content-db/sql/V1_8__footer_default_config.sql
```diff
@@ -0,0 +1,43 @@
...
+  "links": [
+    [
+      "About",
+      "page/about"
+    ],
+    [
+      "Suggest a dataset",
```
nahidakbar: Remove these links for internal magda? See #1782

magda-content-api/src/schemas.ts
```diff
@@ -50,6 +50,80 @@ export const headerNavigation: any = {
...
+        },
+        label: {
+            type: "string",
+            minLength: 1
+        },
+        links: {
+            type: "array",
+            items: {
+                type: "array",
+                minItems: 2,
```
nahidakbar: Why this length limit range?

magda-migrator-content-db/sql/V1_8__footer_default_config.sql
```diff
@@ -0,0 +1,43 @@
...
+      "suggest"
+    ],
+    [
+      "Privacy Policy",
+      "page/privacy-policy"
+    ]
+  ]
+}'),
+('footer/navigation/categories/medium/developers', 'application/json', '{
+  "label": "Developers",
```
nahidakbar: This is not valid as far as content schema specified goes. E.g. order is missing.

magda-migrator-content-db/sql/V1_8__footer_default_config.sql
nahidakbar: I'd probably also separate links and categories into two separate items. Would make migrations like #1782 easier. What do you think @AlexGilleran ?

magda-migrator-content-db/sql/V1_8__footer_default_config.sql
t83714: Good picks 👍 Has updated this

magda-migrator-content-db/sql/V1_8__footer_default_config.sql
t83714: - I thought `Suggest a dataset` will stay for our generic version but removed for internal Magda?
If not the case, I can simply remove it 😄 
- Thought about break the items into two levels: `categories` then `links` but the problem will be hard to maintain two `order` information. Right now, `order` field is for the order of `categories` and links order is determined by array item order. Also, try to make data structure similar to the current one so less code to edit. @nahidakbar Let me know if you got a better idea on the data structure 😄

magda-content-api/src/schemas.ts
t83714: Current config data is a 2 dimension array:
```
"links": [
    [
      "About",
      "page/about"
    ],
    [
      "Suggest a dataset",
      "suggest"
    ]
]
```
The first item is the label and the second item is the link.
`maxItems` used to be 3 --- it was a mistake and I've corrected it

magda-migrator-content-db/sql/V1_8__footer_default_config.sql
t83714: Have added order fields to all items

magda-migrator-content-db/sql/V1_8__footer_default_config.sql
AlexGilleran: Yeah I would've thought having them together is a bit easier, but I don't have a strong opinion 🤷‍♂️

nahidakbar:
Replaced by #1818 which is a copy paste of this

t83714:
Had a chat with @nahidakbar:
A new PR has been created due to rebase issue:
https://github.com/magda-io/magda/pull/1818
We gonna close this one and review the new PR

Issue #208, opened 2017-09-28
Handle registry data set conversion error in Indexer smoother


AlexGilleran:
Logs a giant stack trace and might be returning 500 for webhooks.

This is the error: 
```
Exception: Expected String as JsString, but got null
```

PR #1841, opened 2018-11-07
Issue/1817 Make jurisdiction information available from search api

### LINKED ISSUE #1817: Make search index support jurisdiction
### Problem description

Acceptance criteria:
- Organisations grouped by jurisdiction and name
- Add jurisdiction to search API for all organisations
- Make jurisdiction field available from search API

t83714:
### What this PR does

Fixes #1817 

Change Summary:
- Make jurisdiction information available from search api
- Organisation search result from search api will be aggregated based on jurisdiction and name
  - Records without jurisdiction will not be merged
- Added `correspondence-api: false` to values.yaml for consistency
- Updated typescript version registry API
- Added jurisdiction to org summary & detail page (if available)
- Forgot to mention: also made indexer won't get errored when received json field with `null` value. See #208

Test Site:

https://issue-1817.dev.magda.io/organisations

### Checklist


AlexGilleran:
image

I think organisations with the same or no jurisdiction should be grouped together - so there should only be one CSIRO here.

AlexGilleran:
@philipgrimmett can you take a look at the way the jurisdiction info is represented (below the org names) - https://issue-1817.dev.magda.io/organisations 

image

t83714:
> image
> 
> I think organisations with the same or no jurisdiction should be grouped together - so there should only be one CSIRO here.

No problem~ Will make organisations with same name but no jurisdiction merge into one record~

philipgrimmett:
@AlexGilleran @t83714

Could we add value by making the jurisdiction a search link? For example: clicking 'Government of South Australia' runs an org search for all datasets that fall under that jurisdiction. 

In this instance, we might display the following:
RESULTS MATCHING "GOVERNMENT OF SOUTH AUSTRALIA" (144)
or if we're running a custom query (not just a keyword search):
RESULTS UNDER THE JURISDICTION OF "GOVERNMENT OF SOUTH AUSTRALIA" (144)

Design concept:
screen shot 2018-11-08 at 3 50 24 pm

AlexGilleran:
> Could we add value by making the jurisdiction a search link

Not easily :(. That means we'd have to add that as an option in the search API, plumb it through the dataset search and then figure out some way to represent it in the UI... the plan with this was that we do as small of a change as we can in order to see whether that satisfies DTA's use case on search.data.gov.au before we go any further with it 😬.

t83714:
@AlexGilleran 
It's ready to have a look again.
All org records with the same name without jurisdiction are merged as well:
image

AlexGilleran:
OK looks good to me (technically) @t83714!

@philipgrimmett what do you think we should do about the fonts, keeping in mind that we can't search by jurisdiction/publisher yet?

philipgrimmett:
@t83714 Probably best to keep the same font styling, but change the colour to #6F6F6F

screen shot 2018-11-12 at 12 37 09 pm

t83714:
@philipgrimmett Have updated the font size & color~
The test site takes some time to be ready~
Will let you know once it's ready~

t83714:
@philipgrimmett 
The test site is ready to have a look:
https://issue-1817.dev.magda.io/organisations
image
Please me know if we need to adjust a little more.

PR #1832, opened 2018-11-01
#1824 Migrate static pages to a json structure and make it editable

nahidakbar:
### What this PR does

Fixes #1824 

Need for this migration wasn't urgent but made other things easier.

Merge after https://github.com/magda-io/magda/pull/1828

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.
-   [x] I've linked this PR to an issue in ZenHub (core dev team only)

PR #1844, opened 2018-11-08
#1797: Introduced internationalization, allowed "Organisation" string to be configured

### LINKED ISSUE #1797: Pull all UI strings that mention "Organisation" into the content API
### Problem description
Following on from https://github.com/magda-io/magda/issues/1775 - there are a number of other mentions of the word "organisation" that need to change to "sections", or whatever the instance needs.

This is a symptom of a greater problem, wherein the strings we use in the UI are hard-coded into the JSX files. Ideally we'd be able to change any text shown to the user on the fly. The first step towards that is _starting_ to pull these strings into the content API.

So this ticket is the starting point - all strings that mention "organisation" (e.g. the filter on the search page, page title etc) should be pulled into the content API and retrieved from there when rendering.

Most likely the best way to do this would be to create some kind of `/strings/` root in the content api and retrieve everything at once. Possibly there are cleverer ways too - splitting by page?

### Acceptance Criteria
- As an admin I can change any string that mentions the word "organisations" in the UI
- There's a migration for default new installations
- There's some kind of plan for migrating search.data.gov.au (quite likely this can just be handled by the default migration though, as both default and search.data.gov.au will use the "default" nomenclature)
- The way the strings are stored means that we can pull more and more of the UI strings into there over time (i.e. if everything was stored in one big JSON file, that would be hard to add things to, we need to be able to add more rows to the content api as we move more strings out of JSX files).

AlexGilleran:
### What this PR does

Fixes #1797

This introduces a library called i18next to add internationalisation to the magda web ui. This is configured to get strings from the content api. i18next has a concept called _namespaces_ - if you request a string from a certain namespace, it will automatically retrieve that string, and using its React integration it'll re-render the component when it receives it.

I've set it up so that every page is a i18next namespace. So if you go to `/organisations`, it'll retrieve the `publishersPage` namespace. There's also a `global` namespace that has stuff that's on every page, which for now is just the `appName`. As you move around between pages, you'll hit new namespaces which will result in new requests. I've written a component that will show strings as blank while the strings are being retrieved, and will substitute generic placeholders if the requests fail. These requests are batched together and cached by a custom i18next backend.

These all live in the content db under `/lang`. I did a tidy-up of some of that as a result - in particular writing unit tests for getting strings as a wildcard, and fixing a potential SQL injection.

Best to merge this after #1832 because both have SQL migrations and I've numbered mine to go after the one there.

~This also still needs a search.data.gov.au migration before the issue can be closed.~ Migration is at https://github.com/magda-io/magda-config-data.gov.au/pull/12

### Checklist


magda-content-api/src/schemas.ts
```diff
@@ -160,7 +160,6 @@ export const footerCopyright: any = {
     required: ["order", "href", "logoSrc", "htmlContent"]
```
t83714: Do we need to add schema (for lang setting record) here to make content API accept it?

t83714:
🎉 Tested on Gitlab site and it works well~
Even `organization` keywords on the search panel have been changed --- that's very thoughtful 😄

magda-content-api/src/schemas.ts
AlexGilleran: Yes :(

AlexGilleran:
@t83714 finished making the admin ui bit work, take a look again please? :)

PR #1771, opened 2018-10-11
changes during deployment attempts

t83714:
- db password secret changes
- google load balance prob url fix
- create-secrets tool should save file content to data field

### What this PR does

_Helps_ with #1656 #1657 

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.

PR #1583, opened 2018-08-10
Issue/1403 `create-secrets` utility for helping users to create k8s secrets easier

t83714:
### How to use:
Go to Magda project root directory & run:
```
`yarn bin`/create-secrets
```
You can type:
```
`yarn bin`/create-secrets --help
```
to print the help screen.

#### Save config data to a file
```
`yarn bin`/create-secrets -P > config.json
```
#### Create secrets using config from STDIN or a file

```
cat config.json | `yarn bin`/create-secrets -E -
```
or
```
CONFIG_DATA=`config.json`
echo $CONFIG_DATA | `yarn bin`/create-secrets -E -
```
or
```
 `yarn bin`/create-secrets -E ./config.json
```
#### You can choose to overwrite some settings by ENV variable
e.g. You can create secrets using config data `$CONFIG_DATA` but only overwrite `SMTP_SECRET_PASSWORD`.
```
export SMTP_SECRET_PASSWORD="qazwsx"
echo $CONFIG_DATA | `yarn bin`/create-secrets -E -
```
A list of all available ENV variables can be found from help screen.

Please note: If you answered `NO` to question `Do you want to allow environment variables to overwrite current settings at runtime?`, no ENV variables will be used (except `$CI_COMMIT_REF_SLUG` & `$CI_JOB_TOKEN` depends on your answers to other questions).

#### Other Notes for test:

- File picker will only list all json files from current directory to 2 level deeper.

### Checklist

-   [x] Unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.
-   [x] I've linked this PR to an issue in ZenHub (or there is no issue) and dragged it to the _Pull Request_ column
-   [x] There are sufficient comments for my code to be understandable - and I realise reviewers will pull me up on it if not!

PR #1853, opened 2018-11-12
Fix loading cloud sql config file instead of loading path of config file

nahidakbar:
### Checklist

non-standard


t83714:
Thanks a lot for spotting this bug~ 👍 @nahidakbar 
I think we have multiple issues here:
1> Actual data structure for `cloudsql-instance-credentials` & `storage-account-credentials` should be something like:
```
{
  "deploy-to-google-cloud": true,
  "use-cloudsql-instance-credentials": true,
  "reselect-cloudsql-instance-credentials": true,
  "cloudsql-instance-credentials": {
    "value": "/.../.../cb28d459770c.json", //--- optional value field contains file path for UI prefill purpose
    "data": { //--- actual credentials json data 
      "type": "service_account",
      "project_id": "magda-agriculture",
      "private_key_id": "...",
      "private_key": "...."
      ...
    }
  },
```
You can check the actual data strcuture by:
```
 yarn create-secrets -P
```
The execution part should read actual credentials data from `configData.cloudsql-instance-credentials.data` rather than `configData.cloudsql-instance-credentials`.
This issue has been fixed via PR https://github.com/magda-io/magda/pull/1771 but somehow didn't go to `Master`.

2> As you've pointed out, the `overrideSettingWithEnvVars` should re-format env vars depends on the question types. e.g. boolean should be converted to `true` or `false`.
I've added a new `dataType` field to better accomodate this.
For env var `CLOUDSQL_INSTANCE_CREDENTIALS` (`jsonfile` type), I think it's better to assume the input ENV var contains the file content rather than a file path.
Otherwise, probably it's hard to use this tool in a CI env.
We will be able to set the variable as:
```
export CLOUDSQL_INSTANCE_CREDENTIALS=`cat credentials.json`
CONFIG_DATA=`cat config.json` //--- create-secrets configData; you can get by create-secrets -P
echo $CONFIG_DATA | `yarn bin`/create-secrets -E -  //--- Run in silent mode with config data from stdin `-`
```
More info is here: https://github.com/magda-io/magda/pull/1583

Updated the PR accordingly, please help to have a look~ @nahidakbar

PR #1856, opened 2018-11-15
#1471 Add readiness and liveness probes to all services

### LINKED ISSUE #1471: Add readiness probes to all services
### Problem description
Right now we have liveness probes set up for everything but not readiness probes. This is a problem because it prevents deployments from being zero-downtime, and also messes up the timing of the post-upgrade helm jobs that set up the database schema. We should make it so that every pod has a readiness probe. In most cases we can just use the same method as the existing liveness probe, but in some (e.g. postgres, elasticsearch) we might have to do something different, because it needs to make sure that these are up, ready and able to serve data, not simply started but not yet actually ready to serve data.

MVP:

- [x] Common infrastructure
- [x] Content API

Connectivity:
- [x] Database
- [x] Dependencies between services
- [x] Gateway

Others:

- [x] Admin API
- [x] Auth API
- [x] Correspondence API
- [x] Registry API
- [x] Search API

nahidakbar:
### What this PR does

Fixes #1471

### Checklist

### DIFF
--- a/deploy/helm/magda/charts/registry-api/templates/_helpers.tpl
+++ b/deploy/helm/magda/charts/registry-api/templates/_helpers.tpl
@@ -51,11 +51,18 @@ spec:
 {{- if .root.Values.global.enableLivenessProbes }}
         livenessProbe:
           httpGet:
-            path: "/v0/ping"
+            path: /v0/status/live
             port: 80
-          initialDelaySeconds: 60
+          initialDelaySeconds: 10
           periodSeconds: 10
-          timeoutSeconds: {{ .root.Values.livenessProbe.timeoutSeconds | default 10 }}
+          timeoutSeconds: 10
+        readinessProbe:
+          httpGet:
+            path: /v0/status/ready
+            port: 80
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          timeoutSeconds: 10
 {{- end }}
         ports:
         - containerPort: 80
--- a/deploy/helm/magda/charts/registry-api/templates/_helpers.tpl
+++ b/deploy/helm/magda/charts/registry-api/templates/_helpers.tpl
@@ -53,9 +53,9 @@ spec:
           httpGet:
             path: /v0/status/live
             port: 80
-          initialDelaySeconds: 10
+          initialDelaySeconds: 60
           periodSeconds: 10
-          timeoutSeconds: 10
+          timeoutSeconds: {{ .root.Values.livenessProbe.timeoutSeconds | default 10 }}
         readinessProbe:
           httpGet:
             path: /v0/status/ready
--- a/magda-typescript-common/src/express/status.ts
+++ b/magda-typescript-common/src/express/status.ts
@@ -1,0 +1,204 @@
+import { Router } from "express";
+import request = require("request");
+
+export type Probe = () => Promise<State>;
+
+export interface ProbesList {
+    [id: string]: Probe;
+}
+
+export interface OptionsIO {
+    // inputs
+    probes?: ProbesList;
+    probeUpdateMs?: number;
+    forceRun?: boolean;
+    // outputs
+    ready?: boolean;
+    _installed?: any;
+    since?: string;
+    last?: string;
+    details?: any;
+    error?: any;
+}
+
+export interface State {
+    ready: boolean;
+    since?: string;
+    last?: string;
+    details?: any;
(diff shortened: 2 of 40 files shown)


deploy/helm/magda/charts/registry-api/templates/_helpers.tpl
```diff
@@ -51,14 +51,21 @@ spec:
 {{- if .root.Values.global.enableLivenessProbes }}
         livenessProbe:
           httpGet:
-            path: "/v0/ping"
+            path: /v0/status/live
             port: 80
-          initialDelaySeconds: 60
+          initialDelaySeconds: 10
           periodSeconds: 10
-          timeoutSeconds: {{ .root.Values.livenessProbe.timeoutSeconds | default 10 }}
```
AlexGilleran: Is there a reason why we don't need this anymore? The reason this got put in is because the registry tends to slow down over time, so to mitigate it we made it restart when the liveness probe went over 5 seconds in prod, until https://github.com/magda-io/magda/issues/749 is done.

deploy/helm/magda/charts/registry-api/templates/_helpers.tpl
```diff
@@ -51,14 +51,21 @@ spec:
 {{- if .root.Values.global.enableLivenessProbes }}
         livenessProbe:
           httpGet:
-            path: "/v0/ping"
+            path: /v0/status/live
             port: 80
-          initialDelaySeconds: 60
+          initialDelaySeconds: 10
```
AlexGilleran: Why reduce this to 10? The reason this is set to 60 is that it sometimes takes > 30 seconds to become responsive, which causes it to be restarted, which then causes it to take > 30 seconds to start, which causes it to be restarted etc. At the very least it needs to be configurable I would think?

Admittedly the reasons behind these could use some commenting.

magda-typescript-common/src/express/status.ts
```diff
@@ -0,0 +1,204 @@
...
+        options.probeUpdateMs * 10
+    ).unref();
+}
+
+export function installStatusRouter(
+    app: Router,
+    options: OptionsIO = {},
+    routePrefix: string = ""
+) {
+    // leagy google standard
```
AlexGilleran: legacy? Do we still need it at all if we're changing all the helm charts?

magda-typescript-common/src/express/status.ts
```diff
@@ -0,0 +1,204 @@
...
+        options.since = new Date().toISOString();
+        options._installed = {};
+    }
+    if (!options.probes) {
+        return;
+    }
+    // every probe deserves to not be blocked by state of other probes
+    // so we have parallel probe checkers
+    Object.entries(options.probes).forEach(probe => {
+        const [id, callback] = probe;
```
AlexGilleran: Pulling out readiness/liveness checks into their own router is a great idea, and I really like the idea of just passing async functions that check the status.

This bit I don't understand though - it looks a lot like a whole new readiness/liveness check system? What is the advantage of doing this in our code vs letting Kubernetes do it?

I'm worried because:
- This is really crucial, and we're now doing it via code that we have to maintain, rather than Kubernetes code that's used and tested by thousands of people.
- It removes the ability to configure the initial delay/cadence of readiness/liveness checks - e.g. some services require a bit more time to start, and this changes from environment to environment
- It's not obvious that the options in Kubernetes don't work anymore... also they do still work for some pods (e.g. scala ones) but not for others. This means that someone might set something to be checked every 5 seconds and restart after 5 failures - this will only translate to 1 actual check via this system, resulting in k8s killing the pod after only one failed check.
- Currently if a service fails due to liveness checks failing, `kubectl describe pod <pod>` will give you the details of how many failed, but now that information is going to be false because the actual checks will be done inside the pod

AlexGilleran: Re: our verbal conversation, obviously I did _not_ understand how this all worked 😂 . But it'd still be good to change the readiness check to something more like the liveness one so that any service failure doesn't take down the gateway and keep this for monitoring checks.

AlexGilleran: Oh and maybe a comment to explain that this doesn't replace all the readiness checks, just gives us our own view!

nahidakbar:
@AlexGilleran

nahidakbar:
👍 @AlexGilleran

PR #1857, opened 2018-11-16
Include Magda user agent in external HTTP resource accesses

maxious:
### What this PR does
Add a default user agent to any uses of the "requests" library. The user agent and user agent version are pulled from package.json in that project (hopefully reducing the overhead/risk of not changing the user agent as projects are copied)
The user agent can't have 2 slashes in it as per https://tools.ietf.org/html/rfc7231#section-5.5.3 but should have version after the slash so some string replace required.

### Checklist

-   [x] There are unit tests to verify my changes are correct (testing in one of the minions)


AlexGilleran:
Looks good - could we pull this snippet:

```
const request = requestLib.defaults({
    headers: {
        "User-Agent": "".concat(pkg.name.replace("/", "-"), "/", pkg.version)
    }
});
```

into `magda-typescript-common` though?

Using package.json makes it a bit complicated obviously, but you could get `pkg` inside each connector/minion then pass it into a function that accepts `pkg` and returns `request` ?

maxious:
I'll give it a go, I might even be able to pull the pkg into the common code because it's package.json relative to the final built javascript.

PR #1865, opened 2018-11-21
Issue/1845 Make sure content DB migrator won't stuck for DGA upgrade

### LINKED ISSUE #1845: search.data.gov.au migrators are going to collide with normal migrators
### Problem description
With all this content stuff we've been doing, we've been writing Flyway migrators that we build into our migrator images, and also writing manual scripts for data.gov.au, with the idea being that when the upgrade for data.gov.au happens, we can run the manual scripts then the migrators.

_However_ because in some cases both scripts are running an `INSERT` for the same content item id, this is going to break when we run the migrator. Which is annoying.

So we need to modify the migrators we've done to have `ON CONFLICT DO NOTHING` on the `INSERT` statements.

This will be fine in all our prod deployments because we haven't released the migrators yet, but it's going to make dev.magda.io very very unhappy. So we'll have to manually fix that with `flyway repair` when the change gets merged. But that's OK.

### Acceptance Criteria
- All the migrations for v0.0.50 can successfully run after all the search.data.gov.au-specific sql scripts for v0.0.50 have run.

t83714:
### What this PR does

Fixes #1845 

Changes Summary:
- Add `ON CONFLICT DO NOTHING` to insert statements -- As we're gonna run DGA upgrade script on DGA site before have migrator run, some setting records may have existed.
- Remove unnecessary update statement in `V1_10__change_org_name.sql`

Also, make some adjustments to DGA upgrade script in a seperate PR (on DGA repo): https://github.com/magda-io/magda-config-data.gov.au/pull/13

Tested on local minikube with following steps:
- `helm del --purge magda`
- `kubect delete pvc combined-db-combined-db-0`
- redeploy magda 0.0.49 Rc3 version
- Run all DGA upgrade scripts
- redeploy magda master version
- Then check the migrator logs:

```
Creating database content (this will fail if it already exists; that's ok)
ERROR:  database "content" already exists
Migrating database content
Flyway 4.2.0 by Boxfuse

Database: jdbc:postgresql://content-db/content (PostgreSQL 9.6)
Successfully validated 11 migrations (execution time 00:00.021s)
Current version of schema "public": 1.3
Migrating schema "public" to version 1.4 - default static pages contents
Migrating schema "public" to version 1.5 - Insert blank includeHtml
Migrating schema "public" to version 1.6 - update config
Migrating schema "public" to version 1.7 - insert home content
Migrating schema "public" to version 1.8 - footer default config
Migrating schema "public" to version 1.9 - migrate pages
Migrating schema "public" to version 1.10 - change org name
Successfully applied 7 migrations to schema "public" (execution time 00:00.144s).
```

### Checklist


magda-migrator-content-db/sql/V1_10__change_org_name.sql
```diff
@@ -1,4 +1,3 @@
-UPDATE public.content SET id='lang/en/global/appName' WHERE id='lang/en/appName';
```
AlexGilleran: Without this line won't search.data.gov.au get its `<title>` changed to MAGDA? 😬

AlexGilleran: On second thought I am wrong 😬

PR #1869, opened 2018-11-23
Fix scss compiler

AlexGilleran:
### What this PR does

Right now the build doesn't work because the SCSS compiler has had some weirdness introduced by the change to pull request out into typescript-common. Just needs package.json jostled a bit

(no changelog, tests etc because it's just a dependency change that fixes a recently introduced issue)


t83714:
Tested on gitlab & worked well~ 🎉

PR #1872, opened 2018-11-23
Added default no-cache headers to admin ui.

### LINKED ISSUE #1864: Cache fix for Admin
(this issue was actually written (probably poorly) by @AlexGilleran :) )

### Problem description
Currently we have a default cache time of 60 seconds for everything. Which is fine, we want it to stick in CDNs - if anything it should probably be longer in some cases. If it's just changed that's fine, because it'll be OK in 60 seconds anyway.

_Except_ when we've just changed a value via the admin ui, because that works by doing the operation to make the change, then immediately refetching the data. If that refetch hits the cache, it looks like the original operation to change things had no effect, even if you refresh the whole page! You've got to wait a minute or disable cache via dev tools.

SO what we should do is make sure every GET request (or every request in general is fine too) made from the admin ui has headers set on it to ensure that the response isn't retrieved from or put into the browser cache.

### Acceptance Criteria
- Changes made via the admin ui can be immediately seen
- GET requests from admin ui have cache control headers set

AlexGilleran:
### What this PR does

Fixes #1864 

Now everything that the admin-ui has no-cache headers set on it

### Checklist


magda-gateway/views/admin.js
```diff
@@ -437,7 +437,10 @@ function imageConfig(body, name) {
 
 function request(method, url, body = null, contentType = undefined) {
     console.log(method, url, body);
-    const headers = {};
+    const headers = {
```
t83714: Find this post:
https://stackoverflow.com/questions/49547/how-to-control-web-page-caching-across-all-browsers#answer-2068407
It seems the following headers can provide best compatibility coverage across different web browser:  
```
Cache-Control: no-cache, no-store, must-revalidate
Pragma: no-cache
Expires: 0
```

AlexGilleran: 👍

t83714:
Had another play on gitlab:
https://issue-1864-dont-cache-in-admin-ui.dev.magda.io/auth/admin
And can't locate the no-cache header.
image
The cache headers probably should be added to https://github.com/magda-io/magda/blob/a983b9b2c4475e068d5b97ef3cd6f2f649fb0cae/magda-gateway/src/createAuthRouter.ts#L77  rather than `magda-gateway/views/admin.js`?

AlexGilleran:
I'm really worried about the actual admin document because that will only change on new deployments, I'm only really worried about the AJAX calls, which will change as a result of what the user does.

I might be missing something - do we need to make sure we never cache the JS?

PR #1873, opened 2018-11-26
Issue/1748 Redirect HTTP urls to HTTPS urls

### LINKED ISSUE #1748: HTTP requests to prod domain don't redirect to HTTPS
### Problem description
When requesting the live domain ( search.data.gov.au ) over `http` the web module is responding with a `200` status code. This should be `301`.

`curl -I http://search.data.gov.au`

image

t83714:
### What this PR does

Fixes #1748 

Change Summary:
- Redirect HTTP urls to HTTPS urls based on `X-Forwarded-Proto` header
- Added `enableHttpsRedirection ` switch (default: false)
- Will not redirect `/v0/healthz`

Created a PR to DGA repo as well to turn on `enableHttpsRedirection` switch:

https://github.com/magda-io/magda-config-data.gov.au/pull/14

### Checklist


magda-gateway/src/createHttpsRedirectionMiddleware.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        const uri = new URI(req.originalUrl);
+        if (uri.pathname() === "/v0/healthz") {
+            next();
+            return;
+        }
+
+        const protocol = req.get("X-Forwarded-Proto");
+        if (protocol && protocol === "http") {
```
AlexGilleran: Does this mean that it won't redirect unless the connection is coming through some kind of proxy that sets `X-Forwarded-Proto`?

Is it possible that someone will want HTTPS redirection but not have the header? I guess it always has to go through the ingress in our current config?

In any case I think http://expressjs.com/en/api.html#req.protocol might be better?

kring: 👍 to `req.protocol`. And set your [trust proxy](https://expressjs.com/en/guide/behind-proxies.html) appropriately so that `req.protocol` and other properties use trustworthy `X-Forwarded-*` headers.

t83714:
That's a good idea~ Thanks Guys @AlexGilleran & @kring 👍 
Updated the code to use `req.protocol` instead of accessing `X-Forwarded-Proto` header directly~
Could you please have another look when you have the chance~

PR #1877, opened 2018-11-28
Resolve timeout for some DAP marine dataset

jevy-wangfei:
### What this PR does

Fixes # timeout for some DAP marine dataset

### Checklist

Dap marine dataset may spend more than 10 seconds to response. After talking with Dap team, they realized that this is a bug, but they will not plan to fix it until there are more reports. 

This update only increases the default `request` timeout (read from operate system).


AlexGilleran:
We're just about to drop everything and start working on https://github.com/magda-io/magda/issues/1878, maybe we can fold this into it so we can test at the same time @t83714 ?

t83714:
Yeah~ It makes sense to test this with #1878 👍

PR #1882, opened 2018-12-02
Issue/1878 Make broken link minion per domain request wait time configurable

### LINKED ISSUE #1878: Broken link minion is slamming the DAP too hard
### Problem description
Just got a call from the DAP people. Apparently now we've fixed the DAP connector, the broken link minion is going nuts hitting the same URLs again and again and again, and downloading entire (massive) files. They're not the only ones with this problem - ACT, for instance, rate limits without dropping 429s so we don't know to slow down, and data.gov.au only just unblocked us. We should be better citizens.

I believe this is down to a few issues:
- Now that we're using `GET` instead of `HEAD`, servers are continuing to send the entire file back to us. I believe we were only getting the first 50 bytes but maybe this got taken out / isn't working properly
- When the broken link minion goes through events (say the name is updated 3 times in a row), the broken link will hit the URL each time. This means that from the data hosts point of view, they see GET after GET after GET from the same source for the same URL.
- The fact of dropping responses after we get the headers back now means that the broken link minion _tears_ through URLs, which creates a lot of load on whatever its hitting. It needs to go slower.

So what I propose is this:
- Allow the gaps between hitting URLs on the same domain to be overridable on a per-domain basis... e.g. the default is 1 second, but we can configure data.gov.au to be one minute, data.act.gov.au to be 30 seconds etc.
- When we test a link I _believe_ we record the date of the test. We should look that up before we hit the URL, and if it's below a configurable threshold, not recheck unless the URL has changed.
- Given that `GET` with a byte range isn't so well supported and neither is `HEAD`, we should do both: the minion should first do a `HEAD`, then if that doesn't work, do a `GET`.

### Acceptance Criteria
- We can slow down the requests on data.csiro.au and data.act.gov.au to acceptable enough levels that they return 200
- The same URL is not hit over and over within a certain time threshold
- The broken link minion uses `HEAD` wherever it can and only falls back to `GET` on failure

t83714:
### What this PR does

Fixes #1878 

Make broken link minion per domain request wait time configurable

Change Summary:
- Will try `head` request and fallback to `GET` if receive `405` method not allowed
  - A `DevNull` writable stream to discard all data for `GET`
- Allow to set wait time per domain
- New test cases
- Added config for `magda-dev.yaml`:
  - data.csiro.au: 60 secs
  - data.gov.au: 60 secs
  - data.act.gov.au: 30 secs

Also created PR for DGA repo: https://github.com/magda-io/magda-config-data.gov.au/pull/16

### Checklist

### DIFF
--- a/deploy/helm/magda-dev.yml
+++ b/deploy/helm/magda-dev.yml
@@ -99,7 +99,10 @@ correspondence-api:
   smtpPort: 2525
 
 minion-broken-link:
-  domainWaitTimeConfig: '{"data.csiro.au":60,"data.gov.au":60,"data.act.gov.au":30}'
+  domainWaitTimeConfig:
+    data.csiro.au: 60
+    data.gov.au: 60
+    data.act.gov.au: 30
 
 connectors:
   config:
--- a/deploy/helm/magda-dev.yml
+++ b/deploy/helm/magda-dev.yml
@@ -98,6 +98,9 @@ correspondence-api:
   smtpHostname: "smtp.mailgun.org"
   smtpPort: 2525
 
+minion-broken-link:
+  domainWaitTimeConfig: '{"data.csiro.au":60,"data.gov.au":60,"data.act.gov.au":30}'
+
 connectors:
   config:
     - image:
--- a/deploy/helm/magda-dev.yml
+++ b/deploy/helm/magda-dev.yml
@@ -100,8 +100,8 @@ correspondence-api:
 
 minion-broken-link:
   domainWaitTimeConfig:
-    data.csiro.au: 60
-    data.gov.au: 60
+    data.csiro.au: 5
+    data.gov.au: 5
     data.act.gov.au: 30
 
 connectors:
--- a/magda-minion-broken-link/src/index.ts
+++ b/magda-minion-broken-link/src/index.ts
@@ -6,12 +6,11 @@ import commonYargs from "@magda/minion-framework/dist/commonYargs";
 const ID = "minion-broken-link";
 
 const coerceJson = (json?: string) => {
-    try {
-        const data = JSON.parse(json);
-        return data ? data : {};
-    } catch (e) {
-        return {};
+    const data = JSON.parse(json);
+    if (!data || typeof data !== "object") {
+        throw new Error("Invalid `domainWaitTimeConfig` parameter.");
     }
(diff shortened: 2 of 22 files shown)


AlexGilleran:
Rebased on 0.0.50...

deploy/helm/magda-dev.yml
```diff
@@ -98,6 +98,9 @@ correspondence-api:
   smtpHostname: "smtp.mailgun.org"
   smtpPort: 2525
 
+minion-broken-link:
+  domainWaitTimeConfig: '{"data.csiro.au":60,"data.gov.au":60,"data.act.gov.au":30}'
```
AlexGilleran: You can actually have these specified as YAML then dynamically JSONify them - see gateway/templates/configmap.yaml :)

magda-minion-broken-link/src/index.ts
```diff
@@ -5,15 +5,35 @@ import commonYargs from "@magda/minion-framework/dist/commonYargs";
 
 const ID = "minion-broken-link";
 
+const coerceJson = (json?: string) => {
```
AlexGilleran: Wouldn't we rather have it crash and make it obvious that something is broken if we pass bad config?

deploy/helm/magda-dev.yml
t83714: Yeah~ that will make config file much nicer 😄 
Have updated~ Please have a look~

magda-minion-broken-link/src/index.ts
t83714: Yeah~ it should make more sense to let it crash~ 👍 
Updated to throw the error~

t83714:
I think it's ready to have a look again~

AlexGilleran:
@t83714 I found that it was stalling when connecting to the DAP so I made some changes, can you take a look?

I also have no idea how to test that it's not downloading the file into memory, guess we'll have to cross our fingers :(.

t83714:
😄 Actually looks quite good for me~ 
Only wondering why we need `req.abort()` around:
```
const req = request[method](url, requestOpts)
        .on("error", err => rejectResponse(err))
        .on("response", (response: http.IncomingMessage) => {
            req.abort();
```
Anyway, will wait until you've pushed new commits

PR #1921, opened 2018-12-07
Issue/1874 Added preemption priority classes

### LINKED ISSUE #1874: Spike: Look into Kubernetes pod priority
### Problem description
As of a recently, Kubernetes pod priority has moved into beta:

https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/

We should look into how we can use this - probably something along the lines of making sure the gateway has the most priority, then elasticsearch/db, then public api pods, then minions/connectors.

Desired result:
Spike & do - figure out how to change the configuration and change it

nahidakbar:
### What this PR does

Fixes #1874 

### Checklist

### DIFF
--- a/deploy/helm/magda/charts/elasticsearch/templates/data-statefulset.yml
+++ b/deploy/helm/magda/charts/elasticsearch/templates/data-statefulset.yml
@@ -18,6 +18,7 @@ spec:
         component: elasticsearch
         role: data
     spec:
+      priorityClassName: magda-9
       terminationGracePeriodSeconds: 10
       affinity:
         {{ include "magda.antiPreemptibleAffinity" . }}
--- a/deploy/helm/magda/charts/indexer/templates/deployment.yaml
+++ b/deploy/helm/magda/charts/indexer/templates/deployment.yaml
@@ -12,6 +12,7 @@ spec:
       labels:
         service: indexer
     spec:
+      priorityClassName: magda-8
       containers:
       - name: indexer
         env:
--- a/deploy/helm/magda/charts/indexer/templates/deployment.yaml
+++ b/deploy/helm/magda/charts/indexer/templates/deployment.yaml
@@ -12,7 +12,7 @@ spec:
       labels:
         service: indexer
     spec:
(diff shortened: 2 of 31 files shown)


deploy/helm/magda/charts/indexer/templates/deployment.yaml
```diff
@@ -12,6 +12,7 @@ spec:
       labels:
         service: indexer
     spec:
+      priorityClassName: magda-8
```
AlexGilleran: The indexer should probably sit a level below `api`s given that it's fine for it to be bumped at any time (it'll resume its hook and catch up when it restarts)

deploy/helm/magda/charts/session-db/templates/statefulset.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       labels:
         service: session-db
     spec:
+      priorityClassName: magda-10
```
AlexGilleran: I don't think the session db needs to be higher than the other dbs - if it goes down most things will (should?) still work, it's just signed in functionality that'll fail from a user's point of view, and stuff within the system like crawling, which can wait.

deploy/helm/magda/charts/web-server/templates/deployment.yaml
```diff
@@ -12,6 +12,7 @@ spec:
       labels:
         service: web
     spec:
+      priorityClassName: magda-7
```
AlexGilleran: I would think this is as important as the APIs?

deploy/helm/magda/charts/elasticsearch/templates/data-statefulset.yml
```diff
@@ -18,6 +18,7 @@ spec:
         component: elasticsearch
         role: data
     spec:
+      priorityClassName: magda-9
```
AlexGilleran: should also add a high priority to es masters 👍

nahidakbar:
adjusted @AlexGilleran

PR #1943, opened 2018-12-17
Issue/1938 Made the events page size and the webhook timeout configurable

### LINKED ISSUE #1915: Error message placement is not optimal/Focus is not managed when a user sends form and there are errors (high)
### Problem description
Intopia issues 32 and 34 - 
Error messages pointing out incomplete or incorrectly completed fields in the form are generated when the user submits the form and are displayed below the field in question. The user is expected to see the fields and return to the incorrect fields to correct the error.
and
When the user submits the form and there are errors, inline error messages are generated to display visually below each relevant field. Focus remains on the Send button.

Test if screen reader announces the error and change placement of error message in suggest a dataset page and all forms.  Also need to inform DTA so they can incorporate in to their design library.

[32] Error message placement is not optimal
Type: Expert observation
Choose a Success Criteria
Location: Form error messages, Suggest a Dataset page
Platform: All
Severity: High
What’s the issue?
intopia.digi tal 38
Error messages pointing out incomplete or incorrectly completed fields in the form are generated when the user submits the form and are displayed below the field in question. The user is expected to see the fields and return to the incorrect fields to correct the error.
Why it matters
Blind users have no way of knowing the error messages have appeared. Even if a blind user returns to the first field, they will not know there is an error message below the field. Screen readers will not announce the error until it is encountered – after the user has moved past the field that contains the error.
Our recommendation
Place the error messages between the field label and the field itself. That way the user will see or have announced the error after the label and the next focusable element will be the field with the error, which the user can then correct and move through the rest of the form, correcting any other errors in sequence.
Note that this recommendation is based on this specific form in context on this page, with due consideration for the original design. Other options that change the design more could be considered, such as listing the field errors at the top of the page as a set of anchor links that jump to the relevant fields. Other options again might work better for longer forms or forms with a different range of input options.
and
[34] Focus is not managed when a user sends form and there are errors
Type: WCAG 2.1 failure
Fails 2.4.3 Focus Order - Level A
Location: Form error messages, Suggest a Dataset page + Feedback widget
Platform: All
Severity: High
What’s the issue?
When the user submits the form and there are errors, inline error messages are generated to display visually below each relevant field. Focus remains on the Send button.
Why it matters
intopia.digi tal 40
Screen reader users are not advised that the form has errors, how many errors there are, where the errors are, or how the user can fix them.
Our recommendation
When the error messages are generated, focus should be taken to the labelled title of the first field that generated an error.

AlexGilleran:
### What this PR does

Fixes #1915 

Right now we've got problems because the broken link minion can take a REALLY LONG TIME to check links if they don't support HEAD and return big files on GET. Right now because webhooks restart every 10 minutes if there's no response, that means that the broken link minion gets the same payload over and over.

This makes the page size configurable (and lower) so that the broken link minion is doing less at a time, and also makes the restart time configurable (and much higher).

### Checklist


t83714:
👍 Merging...

PR #1951, opened 2018-12-20
Made statefulsets end up on non-preemptible node in dev

AlexGilleran:
### What this PR does

Right now despite node affinity and node priority, the statefulsets in dev are still ending up on preemptible nodes and getting shut down and started back up all the time, which is super annoying.

Turns out that this is because preferred affinity isn't enough to actually combined with node priority and preemption - in order to kick less important pods off the single expensive non-preemptible instance in dev, a pod has to be unschedulable. So this changes the affinity to required, meaning that the es and postgres pods won't be scheduled in dev unless they're on the non-preemptible node.

In order to make this easier to customise, I've just allowed additions to the affinity to be added as yaml instead of as preference weight. 

I've also added priorities to `tags.all` because it stops pods from deploying successfully without it.

### Checklist

(no issue, this is more an ops thing)

### DIFF
--- a/deploy/helm/magda-dev.yml
+++ b/deploy/helm/magda-dev.yml
@@ -71,7 +71,12 @@ combined-db:
       fileName: TerriaJS-5e042b649f8a.json
   data:
     storage: 250Gi
-  antiPreemptibleAffinity: 100
+  affinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      nodeSelectorTerms:
+        - matchExpressions:
+          - key: cloud.google.com/gke-preemptible
+            operator: DoesNotExist
 
 elasticsearch:
   data:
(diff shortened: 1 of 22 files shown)


t83714:
Deployed to gitlab (no data preview) and got this error:
```
Release "fix-statefulsets-in-dev" does not exist. Installing it now.
Error: YAML parse error on magda/charts/elasticsearch/templates/data-statefulset.yml: error converting YAML to JSON: yaml: line 25: mapping values are not allowed in this context
ERROR: Job failed: exit code 1
```

deploy/helm/magda-dev.yml
```diff
@@ -80,7 +85,12 @@ elasticsearch:
     googleApplicationCreds:
```
t83714: looks like some indent issue above this line

AlexGilleran:
@t83714 successfully deployed now 😎.

Also fixed up it not deploying successfully on old K8S versions.

PR #1955, opened 2018-12-21
Added nodeAffinity prefix to dev affinity declarations.

AlexGilleran:
### What this PR does
So these failed when they got to dev, turns out I somehow messed up the k8s declaration for node affinity (by leaving out `nodeAffinity` at the last moment and kubernetes didn't throw a proper error for it for some reason (it just resolved to `{}`). This makes it work properly.

### DIFF
--- a/deploy/helm/magda-dev.yml
+++ b/deploy/helm/magda-dev.yml
@@ -72,25 +72,27 @@ combined-db:
   data:
     storage: 250Gi
   affinity:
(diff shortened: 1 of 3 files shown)


AlexGilleran:
That's annoying, `magda-0` is apparently conflicting with itself? :(

Removed the globalDefault for now seeing as it doesn't actually do anything (default is already 0)

PR #1991, opened 2019-01-15
Stopped Zero-Rated Datasets Being Unranked Relevant to Each Other

### LINKED ISSUE #1954: Indexer gives datasets with no quality aspect a quality rating of 0
### Problem description
Following on from #1949.

The indexer currently defaults to a quality rating of `0` when it can't find a `dataset-quality-rating` aspect for a dataset. This is a problem because it causes the score of any datasets to become `0` regardless of the search term, which completely destroys keyword search. So there's two problems to solve:

1. Make the quality rating in elasticsearch `1` when there's no aspect to draw it from (essentially we're giving the dataset the benefit of the doubt)
2. Make sure that we're NEVER EVER setting the quality to `0` even if it's a `0` star dataset, so that ranking between 0 star datasets still works. I'm thinking we should set it to `0.01` or something like that.

- Do a search that only returns 0-star datasets or datasets with no quality aspect
- Notice that they're not ordered properly

AlexGilleran:
### What this PR does

Fixes #1954 

This makes the lowest quality a dataset can have in the index `0.01`, so that when multiplied by relevance for search they don't end up with a score = `0`, and are ranked correctly relevant to each other but still rank below higher-quality datasets.

This also makes the UI round down instead of up for the quality indicator, so that if quality = 0.01 it shows up as 0 stars.

Testing notes:
- The current preview namespace has all the quality ratings set to 0 stars so you can test how they rank relevant to each other.

### Checklist


magda-scala-common/src/main/scala/au/csiro/data61/magda/model/Registry.scala
```diff
@@ -238,11 +238,15 @@ object Registry {
...
+          val score = if (totalWeighting > 0) {
             ratings.map(rating =>
               (rating.score) * (rating.weighting / totalWeighting)).reduce(_ + _)
           } else 0d
-        case _ => 0d
+
+          // Make sure no quality score is set to zero, otherwise it results in relevance being * by 0 which makes
+          // results come back in a random order
+          if (score > 0) score else 0.01
+        case _ => 1d
```
t83714: When no qualityAspect, do we want it to be 1 star or 0 star?
If we want it to be 0 star, here probably should be `case _ => 0.01d`?

AlexGilleran: I was thinking we give aspects with no quality the benefit of the doubt - we assume that they're 100% quality until we know that they're not.

Otherwise
a) we end up ranking datasets that might be perfectly good as 0% quality until a minion catches up and
b) if there's no minion (like on internal instances), all the datasets are 0%. Which I guess is fine as long as it doesn't affect search rankings relevant to each other, but it's still a bit weird IMHO.

t83714: 👍

PR #1995, opened 2019-01-18
Made the query for getting aspects by path able to use indices.

### LINKED ISSUE #1993: DB Usage is super high on data.gov.au
### Problem description
Right now we're having to use a lot of CPU cycles for the production database on data.gov.au. This looks to be because even with indices in the right places, the way that registry looks up the old data.gov.au slugs (by querying `ckan-dataset.name` and others) somehow slips through a crack in postgres and avoids the index no matter how you specify it. So we need to find a different way to do it.

AlexGilleran:
### What this PR does

Fixes #1993 

So what ambushed us on the DGA launch was that there's a lot of traffic that comes in on old CKAN addresses and uses the CKAN direction module in the gateway. That module looks up CKAN dataset/org/resource names and ids in the JSON, which is fine as long as we have an index for those paths in the `data` column.

BUT you need the path to be a `varchar[]` in order for it to use the index. If it's not, it'll still work exactly the same, but won't use the index. Which is the difference between a 6ms call and a 6000ms call, 100% utilisation across 6 CPUs vs 3% utilisation, and (importantly) the database not responding at 11:36 every evening when some external scraping job hits us, or the database performing perfectly at that time.

Casting the path to a `varchar[]` doesn't count for some reason, but `string_to_array` does. 🤷‍♂️ 

I actually cheated and manually put this into prod already because I didn't want to get paged at 11:30 again, this is the result:

image

o_0

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,7 @@ Ops:
 -   Added `cloudsql-db-credentials` to create-secrets tool
 -   Only runtime dependencies will be included by docker image build script
 -   Fixed an file selector error in create-secrets script when current directory & non of its sub directory has \*.json file
+-   Changed SQL for querying aspects by path/value so that it could take advantage of JSONB indices.
 
 Security:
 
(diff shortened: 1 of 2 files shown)


t83714:
Tested url https://issue-1993.dev.magda.io/organization/australianbureauofstatistics-geography 
And works well~
Well done~ 👍

PR #2003, opened 2019-01-22
Increase Travis scala test time limit to 30 mins

t83714:
### What this PR does

This PR Increase Travis scala test time limit to 30 mins (as a quick fix) to avoid failure due to test cases timeout .

Tested with this PR seems work alright.

Now if you check the log before the test completes, you'd see something like:
```
$ javac -J-Xmx32m -version
javac 1.8.0_151
Using Scala 2.11.8
travis_time:start:1380adc3
[0K$ travis_retry travis_wait 30 sbt test


Still running (1 of 30): sbt test
Still running (2 of 30): sbt test
Still running (3 of 30): sbt test
```

But once the test is full completed, the log output will be back to the normal and with all detailed output like now.

### Notes:

Even with the help of `travis_retry` & `travis_wait`, we can't go over travis's hard total running time limit --- which is 50 mins.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
           - REACT_APP_SHA1=$TRAVIS_COMMIT
       addons:
         postgresql: "9.6"
-      script: travis_retry travis_wait 30 sbt test
+      script: travis_wait 30 sbt test
     - language: node_js
       node_js: "8"
       cache:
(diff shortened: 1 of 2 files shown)


AlexGilleran:
image

:(

The problem seems to be that sometimes the search tests just stop and don't output anything for 10 minutes now, not so much the time limit?

AlexGilleran:
We can try, but I'm pretty sure this is happening because the test is genuinely freezing up 😬 . We'll probably just find it failing after 50 minutes instead of 30.

t83714:
Yeah~ Failed twice in a row~
Seems an issue that once happen it will never complete~
Let's find a probably solution then~

PR #2011, opened 2019-01-24
Issue/1917 accessibility fix for search facet buttons 

### LINKED ISSUE #1917: Text associated with search facet buttons is inadequate (med)
### To-Do to fix:
Add alt/title text to each filter:

Filter: [filter text]

Dev note: this would require updating the title text to match the current filter state. I assume this is possible? - @philipgrimmett 

### Problem description
Intopia issue 20 - Screen reader users are not given clear information about the functionality and purpose of the facet buttons.

Giving the alt attribute a null value, alt=”” changes the announcement to “clickable any organisation button”, which is an improvement but still not very clear. Expanding the text to be more informative would likely cause design issues.
Therefore, consider inserting a line or two of visible text above the search facet buttons that tell the user what to do. This would likely benefit all users, and definitely those using screen readers.

[20] Text associated with search facet buttons is inadequate
Type: WCAG 2.1 failure
Fails 3.3.2 Labels or Instructions - Level A
Location: Search facet buttons, Search page
Platform: Desktop
Severity: Medium
What’s the issue?
Screen reader users are not given clear information about the functionality and purpose of the facet buttons.
Why it matters
Each search facet button consists of a button element wrapped around an SVG with text overlaid. The SVG is given an alt attribute. The result is that screen readers announce the first button (for example) as “clickable organisation graphic, clickable any organisation button”.
Our recommendation
Giving the alt attribute a null value, alt=”” changes the announcement to “clickable any organisation button”, which is an improvement but still not very clear. Expanding the text to be more informative would likely cause design issues.
Therefore, consider inserting a line or two of visible text above the search facet buttons that tell the user what to do. This would likely benefit all users, and definitely those using screen readers.

t83714:
### What this PR does

Fixes #1917 

-  Added `aria-label` field to Facet buttons to reflect the current filter selection as following format:
```
Default
Organisation filter: any organisation
Location filter: any location
Date filter: any date
Format filter: any format

Active
Organisation filter: Department of Finance
Location filter: STE New South Wales
Date filter: from Jan 2012 - May 2018
Format filter: CSV, JSON
```
In order to make accessibility announcement work, I put generated text content to `aria-label` field rather than `alt` field.

Test Site:

https://issue-1917.dev.magda.io

### Checklist


t83714:
@philipgrimmett 
Yeah~ It's possible. I'll have a look and let you know once the testing is updated. 👍

t83714:
@philipgrimmett 
I've updated the PR for the remove button.
Please have another look and let me know any other changes are required.

@aneesha09 
I had a look at the options reading order issue that you mentioned this morning.
It seems it's caused by a backend search API issue here:  https://github.com/magda-io/magda/issues/2015
We probably want to investigate separately.

PR #2018, opened 2019-01-30
Updated Magda.io

AlexGilleran:
There are no unit tests for this 😛


AlexGilleran:
Need this for a meeting so I'm gonna be bad and merge it (at least it won't break the build!). Still leave comments if there are any!

PR #2030, opened 2019-02-07
Made url for CKAN authentication configurable.

### LINKED ISSUE #2029: CKAN Authentication is hard-coded to data.gov.au
### Problem description
The CKAN authentication provider started out as an experiment so it was hard coded to data.gov.au... and it never got changed, even though there's already a parameter for it in magda-gateway.

Now that data.gov.au CKAN has moved to data.gov.au/data, we need to update it... and we might as well do it properly by allowing it to be configured.

AlexGilleran:
### What this PR does

Fixes #2029 

The hard-coded old data.gov.au location we had in the CKAN auth provider is now out of date - this makes it configurable.

### Checklist

### DIFF
--- a/deploy/helm/magda-dev.yml
+++ b/deploy/helm/magda-dev.yml
@@ -40,28 +40,11 @@ gateway:
     directives:
       scriptSrc:
       - "''self''"
-      - "''unsafe-inline''" # For VWO, hopefully we can remove this soon
-      - "data:" # Also for VWO
       - browser-update.org
-      - dev.visualwebsiteoptimizer.com
-      - platform.twitter.com
-      - www.googletagmanager.com
-      - www.google-analytics.com
-      - rum-static.pingdom.net
-      - https://cdnjs.cloudflare.com/ajax/libs/rollbar.js/2.4.1/rollbar.min.js
-      - https://tagmanager.google.com/debug
-      - assets.zendesk.com
-      - static.zdassets.com
-      - ekr.zdassets.com
       objectSrc:
       - "''none''"
-      sandbox: # We run the sandbox because we have to have unsafe-inline, when we remove unsafe inline we should remove this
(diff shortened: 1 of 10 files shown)


deploy/helm/preview.yml
```diff
@@ -36,28 +36,11 @@ gateway:
...
       objectSrc:
       - "''none''"
-      sandbox: # We run the sandbox because we have to have unsafe-inline, when we remove unsafe inline we should remove this
-      - allow-scripts
-      - allow-same-origin
-      - allow-popups
-      - allow-forms
-      - allow-popups-to-escape-sandbox
       reportUri: https://sdga.report-uri.com/r/d/csp/enforce
+  ckanAuthenticationUrl: https://data.gov.au/data
```
t83714: This line should probably be moved to `auth` section~

deploy/helm/magda-dev.yml
```diff
@@ -40,28 +40,11 @@ gateway:
     directives:
       scriptSrc:
       - "''self''"
-      - "''unsafe-inline''" # For VWO, hopefully we can remove this soon
```
t83714: We now have the following CSP error in console:
```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' browser-update.org". Either the 'unsafe-inline' keyword, a hash ('sha256-lHjh6+8zaTcKepW6RLPXoOB5686KuJPwukiRVR+0zl4='), or a nonce ('nonce-...') is required to enable inline execution.

react-ga.js:893 Refused to load the script 'https://www.google-analytics.com/analytics.js' because it violates the following Content Security Policy directive: "script-src 'self' browser-update.org". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```
So we don't have anything reply on `unsafe-inline` right?

PR #2094, opened 2019-03-08
New Dataset touch-ups for pfod day

AlexGilleran:
### What this PR does

Puts some delicious icing on #2704, which @nahidakbar already baked 😉 .

This is a few touch-ups to the add dataset page that I made for the PFoD day demo. Mainly:

- Showing a loading spinner while the file is being processed. This works on chrome, but doesn't even show up on Firefox... but that's OK because I figure we'll push it into a webworker pretty soon.
- Formatting the output a bit more nicely - e.g. showing file length as KiB/MiB etc, showing dates without the time, showing spatial extent as a sentence instead of a JSON.

I've also added a now.json file for demos - you can now run `now` from the magda-web-client directory if you want to quickly demo something that's client-side-only without going through CI.

I also added a bit to the code for finding columns - it now checks all rows rather that just the first one for header names.

### Checklist
- [x] There are unit tests to verify my changes are correct or unit tests aren't applicable
- [x] No need for changes, this already falls under what Nahid put in there
- [x] No story.


nahidakbar:
image
Seems to be something wrong in being able to test it

nahidakbar:
Probably related: 
image

AlexGilleran:
@nahidakbar should be ok now.

Issue #2025, opened 2019-02-01
When searching for Register of Native Title Claims, the dataset with the exact title comes 5th


AlexGilleran:
### Problem description
E.g. https://data.gov.au/search?q=Register%20of%20Native%20Title%20Claims

Somehow despite having 3 stars, the dataset that matches the title exactly comes 5th behind some less relevant and worse quality datasets.

Need to investigate why this is, definitely doesn't seem right to me.

gordjw:
@AlexGilleran I'm not sure if this is related, but didn't want to start a new issue without checking:

Looking for this dataset "[Street furniture including](https://data.gov.au/dataset/ds-melbourne-https%3A%2F%2Fdata.melbourne.vic.gov.au%2Fapi%2Fviews%2F8fgn-5q6t/details)"

No results are returned when searching by [exact title](https://data.gov.au/search?organisation=data.melbourne.vic.gov.au&q=%22Street%20furniture%22) (where the terms aren't also in the description)

[Searching by description](https://data.gov.au/dataset/ds-melbourne-https%3A%2F%2Fdata.melbourne.vic.gov.au%2Fapi%2Fviews%2F8fgn-5q6t/details?q=%22City%20of%20Melbourne%20owns%20and%20maintains%20various%20objects%20and%20pieces%22) or [non-exact title](https://data.gov.au/search?organisation=data.melbourne.vic.gov.au&q=Street%20furniture) works though!

AlexGilleran:
@gordjw thanks for that - looks slightly different though, I've created #2037 :)

Issue #2065, opened 2019-02-22
Match-ALL search only set defaultOperator to AND for non nested fields


t83714:
### Problem description

Match-all search only set defaultOperator to AND for non nested fields

This line:
https://github.com/magda-io/magda/blob/01f2a7b1b8acebe4921829d8d00beaeec4102a16/magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/ElasticSearchQueryer.scala#L527

is always 'OR'.

i.e. we actually always query all `distribution fields` in `OR` mode even under match-all mode.

This'd explain why #2064 doesn't make search 
https://data.gov.au/search?q=a%20bus 
return 0 result ( just impact the score)

This could also explain #2025 as title search on dataset will never kick-in.

If remove the `of` from the keywords:
https://data.gov.au/search?q=Register%20Native%20Title%20Claims
then you will have the result on no.1

t83714:
closed via PR: https://github.com/magda-io/magda/pull/2097

PR #2097, opened 2019-03-08
Issue/2033 Made Search API always Match All Input  keywords

### LINKED ISSUE #2033: Partially matched result may look counter-intuitive & Probably need to make sure Filter options are non-negotiable
### Problem description

The following links:
https://data.gov.au/search?format=WMS&organisation=CSIRO%20Oceans%20and%20Atmosphere&q=%2A
shows that:

There are `43` documents match `CSIRO Oceans and Atmosphere`

image

But `5832` documents match `WMS`

image

And `10327` results are returned

image

Those different figures could confuse users, although it is correct behaviour --- when first search attempt can't find result we will ignore some of the filters.

### Proposed Solutions:

- We probably should make sure filter options are never ignored. 
- If we want to offer more results to the user as suggestion, we probably should generate alternative query by altering the query string only (i.e. never cancel or modify filter options).

One example could be what google does:

image

Or what ebay does:

image

- If those options are technically hard in short term, we should consider remove the partially match feature

t83714:
### What this PR does

Fixes #2033 
Also fixed: #2025 & #2065 

Change Summary:

- Removed partial match logic
- Pull location keywords from input text and query as all keywords OR non location keywords plus geoQuery
- Default Operator should be AND for distribution query
- Fixed: LanguageAnalyzerSpec.scala
  - Only randomly pick one distribution to test as now it's AND operator in simple_string_query. All keywords must exist in one distribution title
- Adjusted RegionName Boost test cases logic 
- Ignore inexact test cases
- Added mapping for accrualPeriodicity
- corrected not exist field name `contactPoint.name` -> `contactPoint.identifier`
- moved `NON_LANGUAGE_FIELDS` to a seperate simple_string_query so they won't generated always false query terms

### Test Site:

https://issue-2033.dev.magda.io

### Checklist


magda-int-test/src/test/scala/au/csiro/data61/magda/api/DataSetSearchSpec.scala
```diff
@@ -641,7 +669,7 @@ class DataSetSearchSpec extends BaseSearchApiSpec with RegistryConverters {
         }
       }
 
-      it("inexact") {
+      ignore("inexact") {
```
AlexGilleran: I think we should just delete these if we don't need them, can always get them out of git later if we need them.

.travis.yml
```diff
@@ -1,75 +1,75 @@
 dist: trusty
 sudo: false
 matrix:
-  include:
```
AlexGilleran: Looks like a lot of these yml files have had 4 space formatting applied, can we revert that? If it's prettier's fault we might have to put new rules in prettier rc.

magda-int-test/src/test/scala/au/csiro/data61/magda/api/DataSetSearchSpec.scala
t83714: Deleted~ 👍

.travis.yml
t83714: Yeah~ it's triggered by prettier pre-commit hook~
Overrider it to 2 spaces indent.
It seems pettier enforce an `indentation style for sequence in mapping`. 
i.e. 
```yaml
paths:
- .yarn
- "yarn.lock"
```
Will become:
```yaml
paths:
  - .yarn
  - "yarn.lock"
```
See:
https://github.com/prettier/prettier/issues/4723
If that's not desired style, we probably have to add those yaml files to ignore list.

AlexGilleran:
Just realised this also needs the index version bumped in `common.conf` 😬

.travis.yml
AlexGilleran: Ah right. Yeah I suppose the new style is neater, makes for an ugly diff though 😬

t83714:
> Just realised this also needs the index version bumped in `common.conf` 😬

Thanks a lot for this 👍 Have added a new commit~

PR #2108, opened 2019-03-14
Issue/2044 Simplify Facet Processing

### LINKED ISSUE #2015: Search datasets API returns facets options in wrong order sometimes
### Problem description

Two issues:
- Search API Seems NOT sort facets options correctly sometimes.
  - A options with 0 `hit count` can be returned as the first option sometimes
- Format facets sometimes returned with an invalid identifier:
  - e.g.  returned an identifier appeared to be an organisation id:
```
hitCount: 0
identifier: "org-aodn-Integrated Marine Observing System (IMOS)"
matched: true
value: "png"
```

Send the following request to search API (dev site master branch):

https://dev.magda.io/api/v0/search/datasets?query=*&publisher=Department%20of%20Environment%2C%20Land%2C%20Water%20%26%20Planning&publisher=Geoscience%20Australia&publisher=CSIRO%20Oceans%20and%20Atmosphere%20-%20Information%20and%20Data%20Centre&format=pdf&format=zip&format=xlsx&format=csv&format=wms&format=tiff&format=png&region=STE%3A4&start=0&limit=11

You will got the following response (see screenshot):

image

Problems are:
- Option `Department of Environment, Land, Water & Planning` with 0 hitCount shouldn't returned as the first option
- Option `png` with 0 hitCount shouldn't returned as the first option
- Option `png` shouldn't come with an `identifier` `org-aodn-Integrated Marine Observing System (IMOS)` (looks like an organisation id)

### LINKED ISSUE #2044: Search API: Simplify facet processing
### Acceptance Criteria:

- No per facet option filter aggregation on different levels. 
e.g. aggregation like `format-html-filter` `format-pdf-filter` etc.
- Filters should work properly after the change for AgMag and DGA
- Should not change API signature/spec, only the logic behind it

### Problem description

When searching datasets, the search API needs to return the `hitCount` of all facet options.  We currently aggregate on different levels in order to get this result. 

One sample request:

[codebeautify.json.txt](https://github.com/magda-io/magda/files/2863524/codebeautify.json.txt)

Can we simplify the aggregation logic so that it's easier to extends the logic to support dynamic filter #2014  ?

e.g. 
- Will a global term aggregation (with one facet type removed) sufficient  to produce the result we need? --- this one will probably be same with the next one if we remove the `Partially match`
- Or Much simple term aggregation within current query context?

We probably also want to relook at the followings:
- Partially match make it harder so we may want to have a look this issue first: #2033 
- Should the meaning of the facet option hitCount be:
  - How many document more we can get by select this facet option (global term aggregation (with one facet type removed))
  - How many documents included in the result for each of the facet options (Simple term aggregation within current query context)

t83714:
### What this PR does

Fixes #2044 
Also fixed #2015 

This PR is about Simplify Facet Processing so that:
- only use one top filter agg
- no per facet option aggs
- Selected (Matched) options will always be returned on top and always in descending order

Other changes:
- Updated `yarn dev` script in magda-elastic-search so it's easier to start a local test node for local testing
- Added `countErrorUpperBound ` field to facetOption: If it's larger than 0, it mean still has documents not included in result. Thus, hitCount is not accurate
- Updated prettier.config.js: `**/*.yml` will not match `.travis.yml`. Needs to add `**/.*.yml`

### Test Site:

https://issue-2044.dev.magda.io

### Checklist


magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/FacetDefinition.scala
```diff
@@ -109,6 +113,18 @@ trait FacetDefinition {
...
+  def getInputFacetOptions(query: Query): List[String]
+
+  /**
+    * There is a bug in elastic4s that:
+    * if `include` or `exclude` field of `terms agg` is a Seq with only 1 element, a string instead of an array of string will be generated in JSON request
+    * This will cause an `x_content_parse_exception` as `include` & `exclude` must be both array or string
+    * This function will duplicate the only element in a Seq if its size is 1
+    * Luckily elasticsearch will ignore any duplicate items in `include` or `exclude`
+    */
+  def atLeast2Items(items: Seq[String]): Seq[String] = if(items.size == 1) items ++ items else items
```
AlexGilleran: Can we name this something like `fixArrayBug`?

When it's used elsewhere, `atLeast2Items` reads like you're filtering for facet items that contain at least 2 items, which is a bit confusing :).

t83714: @AlexGilleran Renamed `atLeast2Items` to `fixArrayBug`
Please have another look when you got a chance~

PR #2126, opened 2019-03-26
Provide access to CKAN Data API for enabled resources

maxious:
### What this PR does

Makes CKAN Data API access obvious again with "Access Data API" buttons":

Screen Shot 2019-03-26 at 6 13 13 pm
Screen Shot 2019-03-26 at 6 02 54 pm

Happy to discuss visual design.

### Checklist


magda-web-client/src/Components/DistributionRow.tsx
```diff
@@ -166,7 +167,24 @@ class DistributionRow extends Component<PropType> {
...
+                    )}{" "}
+                    <br />
+                    <br />
+                    {distribution.ckanResource &&
+                        distribution.ckanResource.datastore_active && (
+                            <a
+                                className="download-button au-btn au-btn--secondary"
+                                target="_blank"
+                                rel="noopener noreferrer"
+                                href={distribution.sourceDetails.url.replace(
```
AlexGilleran: This makes me a bit nervous, is there any chance at all that `distribution.sourceDetails` or `distribution.sourceDetails.url` could be null or undefined? I'd be a bit more comfortable if there was a check to avoid a nullpointererror whiting out the whole page 😬

magda-web-client/src/Components/RecordHandler.js
```diff
@@ -257,6 +257,21 @@ class RecordHandler extends React.Component {
...
                             </a>
                         ) : null}{" "}
+                        {this.props.distribution.ckanResource &&
+                            this.props.distribution.ckanResource
+                                .datastore_active && (
+                                <a
+                                    className="download-button au-btn au-btn--secondary"
+                                    target="_blank"
+                                    rel="noopener noreferrer"
+                                    href={this.props.distribution.sourceDetails.url.replace(
```
AlexGilleran: Ditto the above comment re: a null check

magda-web-client/src/Components/DistributionRow.tsx
maxious: lodash _.get() to the rescue! I can use the generic CKAN API docs for the link if this field isn't there.

PR #2171, opened 2019-04-29
Issue/2150 Item 1 Whoami Endpoint Changes

### LINKED ISSUE #2150: End to end basic "Approver" functionality: Item 1 Whoami Endpoint Changes
### This is the break-down task `item 1` for  #2140 

Create simple `Roles` & `Permissions` tables in Database & Make sure that information is available from Auth API `whoami` endpoint

We currently have  `admin` and `non-admin` but they are implemented as a simple `is_admin` field (with value yes / no) and can't support the third role `approver` without DB changes.

Moreover, having `Roles` & `Permissions` in the database will provide the following benefits:
- We can try different access control story without modifying the code. Just update the records on those tables. It will save us a lot of overhead while we are trying different ideas
- This allows us to transit from the current simple role model into a more formal access control model but still be compatible with current functionality. i.e. Make system still have two roles by default (Admins / Authenticated Users) and still keep the current default admin account for API access. 

To minimize the trouble of deployment, we should make sure when `roles` & `permission` tables are empty, there will no impact on current functionality. 

We probbaly can leave the org unit structure to a sperate ticket if we still have concerns regarding its implementation.

We don't need a UI for this task. You can alter role & permission later via Database management tool.

> Item 1 Estimate: 5
> No blockers

t83714:
### What this PR does

Fixes #2150

Item 1 Whoami Endpoint Changes

- Create Initial roles permissions table schema
- more generic authentication
- Whoami API should return more details. Which will be used as input of all OPA request
```json
{
  "id": "",
  "displayName": "Anonymous User",
  "email": "",
  "photoURL": "",
  "source": "",
  "sourceId": "",
  "isAdmin": false,
  "roles": [
    {
      "id": "00000000-0000-0001-0000-000000000000",
      "name": "Anonymous Users",
      "description": "Default role for unauthenticated users",
      "permissionIds": [
        "e5ce2fc4-9f38-4f52-8190-b770ed2074e6"
      ]
    }
  ],
  "permissions": [
    {
      "id": "e5ce2fc4-9f38-4f52-8190-b770ed2074e6",
      "name": "View Published Dataset",
      "resourceId": "951c688c-074d-45f6-807f-15e2587d743b",
      "resourceUri": "object/dataset/published",
      "userOwnershipConstraint": false,
      "orgUnitOwnershipConstraint": false,
      "preAuthorisedConstraint": false,
      "operations": [
        {
          "id": "d0554b94-9904-4c2a-b168-295cf503ed4e",
          "uri": "object/dataset/published/read",
          "name": "Read Publish Dataset"
        }
      ]
    }
  ]
}
```

### Checklist


magda-migrator-authorization-db/sql/V1_4__default_dataset_resource_operations.sql
```diff
@@ -0,0 +1,16 @@
+INSERT INTO "public"."resources" ("id", "name", "uri", "description")
```
AlexGilleran: why split these up?

t83714: v1.3 migration is fundamental things to describe the story --- structures & default values are likely never required to be changed even we decided to change our story.
v1.4 & up are an implementation of the story --- they are all setup initial data --- we may want to change it if we decide to change the story.
I was thinking it might be helpful to split it up if we want to change things while we work on the story. 😄 
Merge those files into one definitely will make DB migration complete quicker.
@AlexGilleran 
Do you think we should merge those file now or wait a bit later?

AlexGilleran: If the data works, might as well merge them :)

t83714: Merged them into one here (on PR for issue/2155):
https://github.com/magda-io/magda/pull/2184/commits/baf54711bcb872741003a4d1a4f5a8ac14aa8352

AlexGilleran: OK I'll close this then :)

PR #2172, opened 2019-04-29
Issue/2151 Item 2 OPA deployment

### LINKED ISSUE #2151: End to end basic "Approver" functionality: Item 2 OPA deployment
### This is the break-down task `item 2` for  #2140 

OPA deployment:  Added OPA (Open Policy Agent) to our cluster with a simple setup of:
  - 1 replicate 
  - deploying policy via `ConfigMap` so we can modify the policy without re-deployment
    - configMap has 1 Mb limitation but should be sufficient for our idea exploration
    - We can change it to a more stable option (i.e. serve a bundle via HTTP endpoint) later 

> Item 2 Estimate: 5
> No blockers

t83714:
### What this PR does

Fixes #2151 

OPA deployment: Added OPA (Open Policy Agent) to our cluster with a simple setup of:

1 replicate
deploying policy via ConfigMap so we can modify the policy without re-deployment
configMap has 1 Mb limitation but should be sufficient for our idea exploration
We can change it to a more stable option (i.e. serve a bundle via HTTP endpoint) later

#### Commits after issue/2150

https://github.com/magda-io/magda/compare/issue/2150...issue/2151

### Checklist


magda-authorization-api/src/createOpaRouter.ts
```diff
@@ -0,0 +1,141 @@
...
+    router.use(bodyParser.json({ type: "application/json" }));
+
+    async function appendUserInfoToInput(
+        req: express.Request,
+        res: express.Response
+    ) {
+        let userInfo: User;
+
+        try {
+            userInfo = await database.getCurrentUserInfo(req, jwtSecret);
```
AlexGilleran: probably need to log this or something if it isn't already? Also won't this mean that the user just gets told they're not allowed to do something even though they are because the database failed somehow?

t83714: This interface mainly used for providing some background data (`input.user`) for OPA evaluation (we have to decide a role to make the story unified & simple 😭 ). The user won't directly interact with the OPA decision here, OPA policy (every policy serves as an API) may decide to return more info through API and UI may decide to tell the user something case by case. 
Put as Anonymous users as it's close to the meaning of unable to authenticate the user.  
@AlexGilleran 
Do you think it's sufficient if we log the database error at backend via a simple `console.log`?

AlexGilleran: > Put as Anonymous users as it's close to the meaning of unable to authenticate the user.

I'm more inclined to say that a database failure here should cause whatever it depends on to fail too - given that it's all about security it'd be good to be as careful as possible. e.g. if down the line we built a version that disabled the anonymous user and an attacker got access by overloading the database until they're defaulted to an anonymous user anyway.

t83714: Yeah~ I probably will just throw an error with 500 status code for database errors~
Will add this to latest PR for issue/2155

t83714: Updated on PR for issue/2155:
https://github.com/magda-io/magda/pull/2184/commits/baf54711bcb872741003a4d1a4f5a8ac14aa8352

AlexGilleran:
Closing as changes made further upstream :)

PR #2199, opened 2019-05-08
Added basic spatial preview to add dataset

### LINKED ISSUE #2191: Add Spatial Preview Map to New Dataset Page
### Problem description
We want to show a spatial preview of the geographic extent of a new dataset via a map on Step 1 of the add dataset page:

image

### Acceptance Criteria
- A map is on the page, with a rectangle on it that reflects the dataset's current spatial extent
- There's a "Show Spatial Preview" button that allows it to be shown/hidden. If the dataset has a spatial extent then it should be "Yes" by default, otherwise "No". So if we've picked up a spatial extent from the file, that should already be on the map when the page loads.
- When the spatial extent is changed via the text boxes in the UI, the map preview should change accordingly
- The map should match the style of the search filter map if possible
- There should NOT be an open in nationalmap button even though the design has it.

### Technical Notes
- This shouldn't require too much technical cleverness - if we could just load up leaflet and tell it to draw a rectangle with the correct coordinates that would be ideal. There's probably no need to load up a full instance of terria and tell it to add a `MagdaCatalogItem`.

nahidakbar:
### What this PR does

Fixes #2191 

Added basic spatial preview to add dataset through leaflet-react

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@ UI:
 -   Refactored web client to group similar things togeather
 -   Implemented basic admin pages in react ui
 -   Started implementing new add dataset flow design changes
+-   Added basic spatial preview to add dataset
 
 ## 0.0.55
 
--- a/magda-web-client/package.json
+++ b/magda-web-client/package.json
@@ -49,6 +49,7 @@
     "react-app-polyfill": "^0.2.1",
     "react-file-drop": "^0.2.7",
     "react-i18next": "^8.1.0",
+    "react-leaflet": "^2.2.1",
     "react-list-drag-and-drop": "^0.9.0",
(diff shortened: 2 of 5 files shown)


t83714:
@nahidakbar 
Tested with this file
[Workbook1.xlsx](https://github.com/magda-io/magda/files/3160465/Workbook1.xlsx)
And the preview looks great with populated spatial extent.

Just one little thing:
For a file without spatial extent data:
e.g. this one: 
[non spatial.xlsx](https://github.com/magda-io/magda/files/3160491/non.spatial.xlsx)
According to the ticket requirement, we probably wanted to set the toggle button to `NO` and hide the map preview~

PR #2216, opened 2019-05-14
User testing fixes

AlexGilleran:
### What this PR does
- Removes CSS modules because they don't work with the SCSS compiler ( :( )
- Pulls the airbnb react-dates CSS into its own file so that the SCSS compiler picks it up (I think it's ignoring it because it starts with `_`.
- Makes the airbnb react-dates CSS work in general

### DIFF
--- a/magda-web-client/src/Components/Common/Choice.scss
+++ b/magda-web-client/src/Components/Common/Choice.scss
@@ -1,27 +1,27 @@
-.buttonHeading {
+.choice-buttonHeading {
     font-size: 0.9em;
     margin: 0 0 1em 0;
     text-align: left;
     font-style: italic;
 }
 
-.choiceIconRow {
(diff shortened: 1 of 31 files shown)


t83714:
👍 Good Job~
Only one tiny thing:

The icons are much smaller now:

image

Probably because of those lines:

https://github.com/magda-io/magda/blob/20f3a15fb37463f84ccae6f1aac429d6bb479ee5/magda-web-client/src/index.scss#L9-L11

We probably want to restrict the scope of the `border-box`

AlexGilleran:
I'd really like to keep the `border-box` thing - it's necessary for the datepicker to work, and also it's pretty normal for modern web development (non border-box sizing is crazy), I'm surprised we only had it set on divs so far.

PR #2218, opened 2019-05-20
[WIP] Add Esri auth provider to magda-gateway

### LINKED ISSUE #2167: Investigate Esri Oauth support
Hi Magda team

For a Terria project we have a requirement to display Esri services that are only available to authenticated users. @kring has suggested that we look at implementing an Esri Auth provider within magda to help facilitate this.

Esri offers an [OAuth2 endpoint](https://developers.arcgis.com/rest/users-groups-and-items/authentication.htm) which is helpful.

When it comes to Magda from what I can tell the existing OAuth providers (eg Google, Facebook) are catered for within the [magda-gateway/src/oauth2/](https://github.com/magda-io/magda/tree/28ce9fd1a7f711f1380b22596d6bfc71f8cb8d8d/magda-gateway/src/oauth2) directory. It looks like these integrations are managed using passport.js, good news is there is an [ArcGIS strategy](http://www.passportjs.org/packages/passport-arcgis/) already. So I'm hopeful that I/we could largely follow the existing patterns that have been established.

The front-end login components appear to be managed within [magda-web-client/src/Components/Account/Login.js](https://github.com/magda-io/magda/blob/eef14055ae7a81fd58cad839e21a2be06874f811/magda-web-client/src/Components/Account/Login.js).

All in all this seems fairly achievable, it feels like a lot of the pieces are already there and patterns have been established within Magda. Is there anything major that I'm missing or need to be aware of?

I'm happy to take a look and see how far I get although I'm sure I'll need a bit of hand holding as I familiarise myself with magda & typescript.

Cheers
Rowan

rowanwins:
### What this PR does
This pull request seeks to implement support for Esri OAuth2 support within `magda-gateway`, and also shows it in use within `magda-web-client`.

Currently this PR only supports authenticating against arcgis.com, however the [passport-arcgis](https://github.com/DavidSpriggs/passport-arcgis)  library does also support [passing an `authorizationURL` option](https://github.com/DavidSpriggs/passport-arcgis/blob/master/lib/passport-arcgis/strategy.js#L48) that would allow people to authenticate against Esri Portals (on-prem deployment of arcgis.com) or instances of ArcGIS Server. I'm thinking it would be worth adding an option along these lines.

Testing locally I'm able to authenticate however I eventually get an error
````
Sign In Failed! Error: Encountered error 500 when POSTing new user to http://authorization-api/v0/private/users
```` 
which looks like it's probably something to do with my dev setup beyond what this PR is doing.

I've not touched tests yet as I wasn't sure what might be required on that front so happy to take some advice on that front.

Let me know if there are any other q's or suggestions

cc @kring 

Fixes #2167

### Checklist

PS the last two commits were accidentally blank. I think prettier (or something) took hold and kept [changing these lines](https://github.com/magda-io/magda/pull/2218/files#diff-dd61fc98007a61d16fbf308c025a6f81R107)


AlexGilleran:
Moved to #2220 for testing :)

PR #2236, opened 2019-05-29
Tweaks for running on Windows.

kring:
* Don't specify a storageclass for the Docker registry, cause `standard` doesn't exist in Docker Desktop and the default should be fine everywhere.
* Add yaml and instructions for setting up a volume for Postgres data on Windows, because Postgres is grumpy about the permissions for a hostpath backed by a Windows share.
* (not Windows specific) bump the version of Java used by `magda-builder-scala`.

### DIFF
--- a/deploy/helm/docker-desktop-windows.yml
+++ b/deploy/helm/docker-desktop-windows.yml
@@ -1,0 +1,3 @@
+combined-db:
+    data:
+        storageClass: local-storage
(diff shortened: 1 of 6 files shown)


mwu2018:
I cherry picked the commits and wonder how long it will take to build the combined-db-0 container. It is still in ContainerCreating stage after 7 min.

kring:
We solved @mwu2018's problem offline. He needed to delete the PVC that the existing combined-db PVC and pod to force the PVC to be recreated on the right PV.

t83714:
🎉 
Thanks, guys~ 👍 
This pr will definitely help a lot of windows users

PR #2241, opened 2019-06-03
Readded exclusion for lexicon js to speed up build times.

AlexGilleran:
### What this PR does

This still makes the difference between `yarn run dev` taking 30 minutes vs a couple of minutes for me, so I'd like to re-add it.

### DIFF
--- a/magda-web-client/craco.config.js
+++ b/magda-web-client/craco.config.js
@@ -18,6 +18,37 @@ module.exports = {
             );
             webpackConfig.module.rules = updatedRules;
 
+            // For some reason the unusual structure of lexicon.js causes babel-web-server to take 20 minutes to compile.
+            // Exclude it
+            webpackConfig.module.rules = webpackConfig.module.rules.map(
+                rule => {
+                    if (rule.oneOf) {
+                        rule.oneOf.forEach(rule => {
+                            if (
+                                rule.options &&
+                                rule.options.presets &&
+                                rule.options.presets.length &&
+                                rule.options.presets[0].some &&
+                                rule.options.presets[0].some(
+                                    preset =>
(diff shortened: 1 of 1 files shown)


kring:
I actually tried it before, and it was fine, before I removed it as part of my last PR. But I removed it because it didn't make things any faster on my system. And because the main part of the change already brought the build times from ~infinity to 2 minutes, I thought this wouldn't be necessary on Mac either. Sorry about that.

mwu2018:
This is what I have if I run "lerna run build --include-filtered-dependencies"
"lerna info run Ran npm script 'build' in '@magda/web-client' in 204.7s:" 
Then it says:
yarn run v1.15.2
$ pancake && craco build --react-scripts ../node_modules/react-scripts
...
The bundle size is significantly larger than recommended.
...
Done in 203.35s.

Finally it stays there forever.

kring:
Try adding a `--stream` to the end of your lerna command line so you can see where it's hung. Sounds like some other project, other than `magda-web-client` is hanging for you.

mwu2018:
I ran "yarn build" within `magda-web-client` without this PR, it hangs forever too. The output looks like this:
```
magda-web-client$ yarn build
yarn run v1.15.2
warning package.json: "dependencies" has dependency "@craco/craco" with range "^5.0.2" that collides with a dependency in "devDependencies" of the same name with version "^3.6.0"
$ pancake && craco build --react-scripts ../node_modules/react-scripts
...
The following changes are being made to your tsconfig.json file:
  - compilerOptions.baseUrl must not be set (absolute imports are not supported (yet))

Creating an optimized production build...
```

mwu2018:
It was running. It finished after 840 seconds.

PR #2268, opened 2019-06-26
Added vanguard to dev deploy and hopefully fixed in master

AlexGilleran:
### What this PR does

Fixes what tripped Mike up today - the Vanguard stuff is merged and working but our config in preview and in dev isn't quite right. 

This adds the creation of the vanguard cert secret to CI script and adds the vanguard config to dev.


deploy/helm/magda-dev.yml
```diff
@@ -33,6 +33,8 @@ gateway:
     arcgisClientId: "d0MgVUbbg5Z6vmWo"
     googleClientId: "275237095477-f7ej2gsvbl2alb8bcqcn7r5jk0ur719p.apps.googleusercontent.com"
     ckanAuthenticationUrl: https://data.gov.au/data
+    vanguardWsFedIdpUrl: https://thirdparty.authentication.business.gov.au/fas/v2/wsfed12/authenticate
+    vanguardWsFedRealm: https://environment.magda.io/integration-test-2
```
t83714: This url https://environment.magda.io/integration-test-2 should be protected by basic auth.
Will that be a problem if used for general config.

By the way, what's the easiest way to test it?

Do we need to use any preset username & password?

AlexGilleran: Nah it's just a identification uri.

You should just be able to click on the vanguard link and log in.

PR #2271, opened 2019-07-01
Issue/2262 New Multi-select / tags input

### LINKED ISSUE #2262: Details And Contents Page - Item 2 New Multi-Select Design
Applied new design to all dropdown box on `Details And Contents Page`.

See screenshot below:

image

- Estimate: ~1-3~  5 

28th Jun: Just realize we need to make Vocabulary suggestion work with this as well. Would be more complicated

t83714:
### What this PR does

Fixes #2262 

- New Multi-select / tags input
- Adjusted Vocabulary Suggestion Input accordingly

image

### test site: 

https://issue-2262.dev.magda.io

### Checklist


AlexGilleran:
Nice work

Just one thing annoying me: If I type and hit the down/up arrows, the text in the box changes and I can select with ENTER ok, but there's no style on the selected element (they all have a white background), which is a bit hard to use. Can we give the currently selected one a dark background?

t83714:
@AlexGilleran 
I made the changes so that the selected item (when you press down arrow key) should get the dark background now. 
Test url see above

PR #2325, opened 2019-07-09
DRAFT: Implementing Auth in the Registry

AlexGilleran:
DOES NOT FIX ANYTHING (or even pass CI) (yet)

Just dumping this into a draft PR so I can get some feedback - don't feel the need to read all 3000 lines, the code is there mainly so you can read the write-up below and then go look up the code that makes it happen if you want more detail.

So essentially this is a proof of concept of how I think we should harmonise the power of the registry with the power of OPA without compromising either. It's been through a few iterations and I think it'll have to go through a few more, so thoughts are very welcome.

### Foundations
- Each recordaspect row now has a column for `authpolicyread`, `-update`, and `-delete`
- The values in these columns match the package names of policies in OPA
- The registry looks up these 

### Potentially controversial things
- I've completely removed the concept of a "Resource" because I think it's a bit confusing to have both resources and policies. Happy to discuss/change this.
- I want to make it so that auth around owner, owning org unit and ad hoc permissions are handled via an `access-control` aspect rather than by creating extra permissions in the auth API - I just think that's a lot simpler. So there'd be no more need to create permissions in the auth-db in order to share a dataset with someone, you'd just concatenate their ID to `access-control.adhocViewUserIds`. You can see what I'm trying to get at in the rego files
- This doesn't actually line up with the search api at all, and I don't think it can or should, just because the search index has a whole different schema. I think the way to align search and registry is to have search policies and registry policies that have the same name but different .rego files.

#### When a record aspect is requested (simplest case)
- A lookup is done to find out what policy that record/aspect combination has
- A call is made to through the auth api to OPA to `/compile`, passing through the user's JWT. 

E.g. imagine that the policy for this aspect says that only users who own the record and have permission to perform the `registry/record/read` operation can access it. So the registry asks the auth api "this user has this JWT and wants to do this, what do I have to know to allow access", then the auth api looks up the user's details and forwards the request to OPA, asking it "this is user `123`, who has permission to perform operations `registry/record/read` (and other details), what do I need to know in order to allow access". OPA then sees that the user has the right operation permission, and sees that their id is `123`, so it comes back and says "this user is allowed access if `record.aspects["access-control"].ownerId = "123"`". 

- The above comes back in OPA's (pretty ugly) AST, which is translated into a strongly-typed and a bit simpler intermediate representation in Scala.
- This intermediate representation is passed into `RecordPersistence`, which knows how to translate it into SQL. Record aspect policies are written to expect a JSON object full of aspects at `input.object`. So a policy can allow/deny access based on any value in any aspect.
- `RecordPersistence.scala` ends up making some kind of request like `SELECT * FROM recordaspects WHERE recordid='<record id>' AND aspectid='<aspect id>' AND ((authpolicyread='<auth policy name>' AND "data->aspect->access-control->ownerId"='123'))`
- If the user has access, the SQL will return a value, otherwise it'll 404, which is by design - we don't want people to be able to know about the existence of records/aspects unless they have the right to view them (otherwise you could probe for stuff like `records/ds-nuclear-launch-codes`)

#### When a record (with multiple aspects) is requested
- A lookup is made to find all the policies of all the aspects that the record has
- All these policies are looked up in OPA and turned into intermediate OPA language
- The lookup happens in `RecordPersistence` - this time each aspect query has `(authpolicyread = 'policyA' AND <policy details>)` attached to it, so if the user has access to one aspect and not another on the same record, it'll only return aspects that the user is allowed to see.
- This means that `optionalAspect` and `aspect` parameters work pretty much how they did before - if you put an `aspect` query for an aspect you're not allowed access to the request will 404, if you put it as an `optionalAspect` then the request will return but that aspect won't come back.

#### When a request for multiple records/aspects is requested
- A lookup is made to find all the policies for _any_ of the aspect ids that have been requested (see potential problem 2)
- Requests are made to OPA as above, but each record aspect subquery has 1-M `(authpolicyread = 'policyA' AND <policy details>)`. 
- So as above, `optionalAspect`/`aspect` work as expected - a request for `/records?aspect=aspectid` will only return records where the user is allowed to see that record's `aspectid` aspect, with `optionalAspect` it'll also return records where `aspectid` exists and the user doesn't have access, but it'll look like that aspect doesn't exist for that record.

#### Update/Delete
Similar to the above but using the `authpolicyupdate` and `authpolicydelete` rows. OPA will be consulted, translated to SQL, the SQL `DELETE` or `UPDATE` statement will be executed with a `WHERE` statement that has the appropriate filters. Currently this means that if a user tries to delete or update something that they don't have access to, the service will react exactly as if the aspect doesn't exist - which isn't ideal (it should probably return a 403 if they can see the aspect but a 404 if they can't) but it's OK for now I think.

#### Create
I haven't started to implement this yet but I think it's pretty simple - each `aspect` row should have a policy attached to it that spells out what users are allowed to create instances of this aspect and with what values.

#### Getting the record (not aspect) metadata i.e. record name, id
Haven't done this yet - I'm thinking it should let you see the record name/id if you have access to any aspect on that record, otherwise 404.

#### Creating a record (separately from aspects)
Also not done - I think we should allow the admin to pick a policy for this. By default it should allow anyone with the "Add a Record" permission.

#### What happens if the OPA compile can't find _any_ way that this user should be allowed to read this aspect?
The SQL query simply resolves to `false` - e.g. `WHERE (authpolicyread = 'policyA' AND false)`, which always matches nothing. Possibly this could be more optimised but I don't think it matters that much.
#### The inverse - what happens if the OPA compile says that this user is always allowed to read this aspect?
`WHERE (authpolicyread = 'policyA')` - i.e. match any aspect that has this policy, because this user is allowed to access all aspects that have this policy.
#### Should update policy also take into account the _new_ values of an aspect - i.e. should I be able to write a policy that prevents certain people from changing certain keys to certain values?
Phew, maybe. I think we can leave this for later.
#### How do I actually set a policy on an aspect?
Right now by manually editing the database. There definitely needs to be an API for this, and also some kind of auth policy around who gets to set auth policies (maybe this policy should be customisable too further down the track? Short term I think it should just be some kind of `registry/set-auth-policy` permission).
#### What about calls that don't come from a user (e.g. webhooks)
I guess it just uses user 0 like existing minions do, although maybe it doesn't need any auth control at all.

### Potential problems
1. Potentially this opens up an attack vector where if an aspect is used by auth but not actually available to be read by an attacker, possibly they could use multiple requests to narrow down what its value was, but I think this is the responsibility of whoever writes the policy
2. Using the `/records` involves getting any policy that applies to any of the requested aspects in the entire recordaspects table, which could mean a lot of requests. However there's probably not going to be _that_ many unique policies, and they'll change extremely seldom, so I think we should be able to cache our way out of trouble if we have to.
3. The SQL is doing a bit more so it's a bit slower... but in my testing found that while queries were 2-3 times more work, it's not like they're _exponentially_ slower, and this could be optimised further.

### TODO (not necessarily all as part of this PR)
- (lots of) unit tests for the scala
- Tests for the rego code
- Rego to support drafts permissions in the registry
- Aligning search and registry permission systems
- Having a proper JSON schema for the access-control aspect and making the registry create it at startup


AlexGilleran:
Superceded by @mwu2018 's prod-quality version :)

PR #2336, opened 2019-07-11
WIP - Esri portal connector

rowanwins:
### What this PR does
This PR adds a new connector for harvesting data from an Esri Portal via [their Rest API](https://developers.arcgis.com/rest/users-groups-and-items/working-with-users-groups-and-items.htm).

### Checklist

### DIFF
--- a/magda-esri-portal-connector/src/EsriPortalUrlBuilder.ts
+++ b/magda-esri-portal-connector/src/EsriPortalUrlBuilder.ts
@@ -11,12 +11,22 @@ export default class EsriPortalUrlBuilder {
     public readonly name: string;
     public readonly baseUrl: uri.URI;
     public readonly apiBaseUrl: uri.URI;
+    public token: string;
 
     constructor(options: EsriPortalUrlBuilderOptions) {
         this.id = options.id;
         this.name = options.name || options.id;
         this.baseUrl = new URI(options.baseUrl);
         this.apiBaseUrl = this.baseUrl.clone().segment("sharing/rest");
+        this.token = null;
+    }
+
+    // https://someportal/arcgis/sharing/rest/generateToken
+    public getTokenUrl(): string {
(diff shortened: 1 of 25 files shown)


magda-esri-portal-connector/src/EsriPortalUrlBuilder.ts
```diff
@@ -0,0 +1,67 @@
...
+        return this.apiBaseUrl
+            .clone()
+            .segment("portals/self")
+            .addSearch({ f: "pjson" })
+            .toString();
+    }
+
+    // https://someportal/arcgis/sharing/rest/content/items/ea532dd8876b4646aa2ab49533e458aa?f=pjson
+    public getContentItemUrl(id: string): string {
+        // TODO Sometimes getting an id of undefined...
```
rowanwins: This is where seems to be throwing the error when trying to create a distribution source

rowanwins:
@kring I think this is now working reasonably well

In terms of the initial crawl of data I'm restricting the initial harvesting of data to `Map Service` and `Feature Service` data types. If you do a general query you get back a squillion irrelevant things like Apps, Tools etc etc.

When setting up the data and distributions I'm crawling the individual layers and adding them as distributions. For example if an item in portal is listed as `/some/MapService` then that becomes the dataset, and I then create distributions for `/some/MapService`, `/some/MapService/0`, `/some/MapService/1` etc etc 

DatasetWithDistributions

Some outstanding questions in my mind...
- how do we pass auth through to a connector (eg so we can access the private content), as it stands the search is only retrieving public content.
- Do we want to add to the data type search `Scene Service` so they get pulled into magda as well even through we don't quite support them in Terria at the moment?

rowanwins:
travis-ci/push is unhappy because there are no tests at the moment

mwu2018:
Superseded by PR #2483

PR #2386, opened 2019-08-02
Update pagination

sukhrajghuman:
### What this PR does

- Pagination **links** should use [anchors, not buttons.](https://marcysutton.com/links-vs-buttons-in-modern-web-applications/)
- Have made pagination links bigger as per [2.5.5](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html)
- Removed the svg images for next and previous links and used '>' and '<' with appropriate aria labels just to simplify.
- Added [aria current](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/current/) to active pagination link
- The previous gray color for pagination summary was not accessible on a white background. Have changed it slightly

TODO:

Right now the anchor tags have blank hrefs...will fix this soon

### Checklist


AlexGilleran:
Good to see you again @sukhrajghuman! Yeah changing them to `<Link>`s would be great, you should be able to just use

```
<Link to={`/search?page=${i}`} />
```

where `i` is the page to go to.

sukhrajghuman:
Hey @AlexGilleran miss ya man.

I had to add the current path so the search query remains in the url. 

Let me know if this is okay?

magda-web-client/src/Components/Common/Pagination.js
```diff
@@ -1,41 +1,56 @@
...
         this.renderPrevButton = this.renderPrevButton.bind(this);
+        this.generateRoute = this.generateRoute.bind(this);
     }
 
-    onClick(page) {
-        this.props.onPageChange(page);
+    // generates the href for each link-list-item in pagination list.
+    generateRoute(page) {
+        return (
+            `search?` +
```
AlexGilleran: Might want to put a leading `/` in front of this, just in case it ever gets invoked from something like `/search/otherPath?etc` because that'll cause it to go to `/search/search` (I think?).

sukhrajghuman: yes good point! I've changed it so it uses `this.props.location.pathname` (which is basically the same as '/search' )

sukhrajghuman:
@AlexGilleran I put  blank alt text since this is a decorative image, and they should be hidden from screen readers.

See reference:

https://www.w3.org/WAI/tutorials/images/decorative/

AlexGilleran:
:shipit:

PR #2394, opened 2019-08-07
Issue/2362

### LINKED ISSUE #2362: Add authorization to getting a record's top-level metadata (name/id)
### Problem description
Protect the [/records/{id}](https://dev.magda.io/api/v0/apidocs/#api-Registry_Record_Service-GetV0RegistryRecordsId) endpoint with authorization - don't worry about protecting the aspects (yet), this is simply about whether a call to get a certain record should return `200` or `404.

However to get this working requires a fair bit of fundamentals to be done - setting up a policy in OPA, and setting up the plumbing to make the registry know how to talk to OPA.

Record authentication wasn't done at all in [the PoC PR](https://github.com/magda-io/magda/pull/2325), so this diverges a bit from that, but you can use a lot of what's done for record aspects.

### Acceptance Criteria
- All records should have some kind of read policy - this should be either a policy that allows all users with a view record permission to view it, or one that only allows access from an owning org unit (see technical notes below)
- Existing records should be accessible just as they are now (so data.gov.au doesn't break)
- If a record can be read, it should return the name, id and any aspects requested (no need for auth on aspects yet)
- If a record cannot be read, it should return 404

### Technical Notes
- A new `access-control` aspect needs to be created that contains a `orgUnitOwnerId` string.
- New policies need to be written for OPA - one that allows any user with the `object/registry/record/read` to view, and one that requires them to both have that permission AND to have the `orgUnitOwnerId` in the `access-control` aspect match the user's `managingOrgUnitIds` (see [these](https://github.com/magda-io/magda/tree/issues/2247-registry-auth/deploy/helm/magda/charts/opa/policies/object/registry/aspect))
- A new column needs to be added to `records` in the registry db - `authpolicyread` - which will contain a reference to the OPA policy to be queried. Make sure that existing rows end up with a read value that allows them to be read just as they would have before.
- Functionality needs to be added to query OPA via the Auth API. For this use case you can either use partial evaluation like the PoC PR does (set the record's details to "unknown", ask OPA what you need to know about a record to allow/deny access and then turn that into SQL), or you could retrieve all a record's aspects and pass them directly into OPA, which will save having to implement an OPA parser. Don't forget that a policy can be set which uses aspects that aren't being requested (i.e. the `dcat-dataset-strings` aspect could have a policy that requires the `access-control` aspect to be read in order to allow/deny access), so you can't only pass the aspect that's being requested into OPA.

### Out of Scope
- Any aspect-specific authorisation (although that will come later) - just return everything for now whether its authorised or not
- Definitely any authorisation around dereferenced aspects
- Getting a single record's details via other endpoints (e.g. history, summary etc)
- Reading multiple record aspects or multiple records
- Anything to do with update/create/delete of anything
- Anything to do with search
- Any kind of admin for any of this
- Actually creating or registering the `access-control` aspect

This one small step, giant leaps can come later ;)

mwu2018:
### What this PR does

Fixes #2362 

### Checklist


AlexGilleran:
@mwu2018 is this superceded by https://github.com/magda-io/magda/pull/2411 ?

mwu2018:
@AlexGilleran Sorry. Too busy at writing test code. Yes, this PR can be deleted.

PR #2415, opened 2019-08-16
Cleaned up dcat-dataset-strings

### LINKED ISSUE #2393: Move provenance metadata out of dcat dataset strings
## Description
For some reason the provenance metadata that environment provide as part of their data inventory made it into dcat-dataset-strings despite having nothing to do with dcat. It should really be in its own aspect.

AlexGilleran:
*** WARNING: DO NOT DEPLOY VIA GITLAB ***
I've made a test site at https://issues-2393-move-provenance-metadata.dev.magda.io/ but it has a password on it because it's loaded the ENV CSV. Ask me for the password, and don't deploy via gitlab or it'll take the password off!

### What this PR does

Fixes #2393 

- Moved provenance info out of `dcat-dataset-strings`, into a `provenance` aspect
- Moved information security out of `dcat-dataset-strings`, into a `information-security` aspect
- Removed some stuff in `dcat-dataset-strings` that didn't match DCAT and we weren't using anyway.
- Changed the UI / CSV connector to work
- Did a few clean ups in the UI, esp. re: state when picking an associated org.
- Improved checking for aspects a bit - now it just `PUT`s a new aspect instead of checking for its existence, the same way as the connectors / minions do. Not ideal but it's better than the aspect being out of date.

### Checklist


magda-scala-common/src/main/scala/au/csiro/data61/magda/model/Misc.scala
```diff
@@ -92,7 +91,7 @@ package misc {
       quality: Double,
       hasQuality: Boolean = false,
       source: Option[DataSouce] = None,
-      creation: Option[DcatCreation] = None,
+      creation: Option[Provenance] = None,
```
t83714: If the key `creation` still stay unchanged, then search API will response `dataset` object with key `creation` rather then `provenance`. Then, in `DatasetSummary` we probably can't reference `dataset.provenance.isOpenData`

t83714:
Probably also need to update here to include new aspects in indexer webhook:
https://github.com/magda-io/magda/blob/437ae51a44ab27ce9016a967dfaff21dd09b2015/magda-scala-common/src/main/scala/au/csiro/data61/magda/model/Registry.scala#L688

AlexGilleran:
@t83714 good points, I've just changed everything to `provenance` now - please take another look :)

AlexGilleran:
Thanks @t83714 , both those should be fixed in the latest commit :)

magda-web-client/src/Components/Dataset/Add/Pages/DatasetAddAccessAndUsePage/index.tsx
```diff
@@ -22,23 +20,29 @@ import { getFormatIcon } from "../../../View/DistributionIcon";
...
-        _licenseLevel
+        informationSecurity
     } = props.stateData;
 
     const editDatasetPublishing = props.edit("datasetPublishing");
     const editDatasetAccess = props.edit("datasetAccess");
-    const editDatasetUsage = props.edit("datasetUsage");
+    const editInformationSecurity = props.edit("informationSecurity");
+
+    console.log(licenseLevel);
```
t83714: We probably want to remove this~

t83714:
Tested in dev cluster and license ctrl works great!
I think we only need to remove the single line `console` before merge~

PR #2429, opened 2019-08-22
Issue/2424 Capture publishing errors and show errors to users

### LINKED ISSUE #2424: Add Dataset Page 5, once `Publish` button is click, Any error should be displayed in an Error Alert Box
## Description

Add Dataset Page 5, once the `Publish` button is clicked, Any error should be displayed in an Error Alert Box.

Currently, any errors will cause `Publish` action failed but the user won't get any feedback.

There are three main scenarios we need to cover:
- Server / Connectivity Error (`500` or timeout) - we should show "Magda has encountered an error when submitting your dataset. Everything is still saved locally - please try again in 5 minutes"
- Validation Error - "Magda has encountered the following error: ${error}. We've logged this internally, please contact ${emailAddress} for help." - where emailAddress is the default magda email addres that's used for dataset requests.
- Authorisation Error (`403`) - display "You've been logged out - please <a>Sign In</a> again to submit the dataset - everything has been saved locally".

## Acceptance Criteria

-  Any error thrown within `Publish` button event handler should be captured.
    - Including errors thrown in any promise created in the handler 
- The error message should use Au Gov design system [`error alert` component](https://designsystem.gov.au/components/page-alerts/)
- The error message will be displayed above the `Publish` button (and below the last textarea question)

## Out of Scope
- Form validation - this should only cover errors thrown by the publish ajax calls

t83714:
### What this PR does

Fixes #2424 :  Capture publishing errors and show errors to users
- Make sure all promises are handled in a way that errors can be captured at higher-level try/catch
- Add an error state to Redux global state store instead of in component state
  - Because progress Meter (not a chid component of the metadata page) needs to clear the error message when the user moves to a different step
- Added default contact email config to web-server pod
  - Also added a global `.Values.global.defaultContactEmail` values
    - When it presents, both correspondence-api & web-client will receive this value
    - you still can set different values at the `correspondence-api` or `web-server` chart level
- Make sure API code is using `registryAuthApiUrl` & `registryApiUrl` value from the config rather than ad-hocly string concatenation

### Test Site:

https://issue-2424.dev.magda.io

### Checklist


magda-web-client/src/Components/Dataset/Add/ErrorMessageBox.tsx
```diff
@@ -0,0 +1,115 @@
...
+                                    }`}
+                                >
+                                    {config.defaultContactEmail}
+                                </a>{" "}
+                                for help.
+                            </span>
+                        </div>
+                    </div>
+                );
+            case 403:
```
AlexGilleran: Looks like this needs to cover `401` too:

image

image

magda-web-client/src/api-clients/RegistryApis.ts
```diff
@@ -27,15 +27,11 @@ export function fetchOrganization(publisherId: string): Promise<Publisher> {
...
-    await request(
-        "PUT",
-        `${config.baseUrl}api/v0/registry-auth/aspects/${id}`,
-        {
-            id,
-            name: jsonSchema.title,
-            jsonSchema
-        }
-    );
+    await request("PUT", `${config.registryAuthApiUrl}aspects/${id}`, {
```
AlexGilleran: For some reason this is going to `/registry` and not `/registry-auth`

magda-web-client/src/Components/Dataset/Add/ErrorMessageBox.tsx
```diff
@@ -0,0 +1,115 @@
...
+                            </span>
+                        </div>
+                    </div>
+                );
+            case 400:
+                return (
+                    <div className="au-body au-page-alerts au-page-alerts--error">
+                        <div>
+                            <span>
+                                Magda has encountered the following error:
```
AlexGilleran: This needs a space before the error, and we should get rid of the full-stop at the end too:

image

magda-web-client/src/Components/Dataset/Add/ErrorMessageBox.tsx
t83714: This probably caused by the other problem below.
But we could hit into a 401 authorisation error in future --- do we need a different error message?

magda-web-client/src/api-clients/RegistryApis.ts
t83714: It seems this will be fixed by your commit in this PR:
https://github.com/magda-io/magda/pull/2411
I will probably merge that once it goes into the master to avoid creating more reference points of the legacy `/registry-auth`

magda-web-client/src/Components/Dataset/Add/ErrorMessageBox.tsx
t83714: added a space between error: and actual error message and remove the extra `.`

magda-web-client/src/Components/Dataset/Add/ErrorMessageBox.tsx
AlexGilleran: Nah I think it should show the 403 message for 401 too :)

AlexGilleran: ... although the message should probably be something along the lines of 

> You don't have permission to publish the dataset - this is most likely because you've been logged out - please Sign In again to submit the dataset. Everything has been saved locally.

magda-web-client/src/api-clients/RegistryApis.ts
AlexGilleran: I think it's probably best to just fix this and let that one hit a merge conflict... it might be a while 'til that one's ready for merge because it's pretty huge and pretty complex whereas this one's quite small and simple.

Issue #2403, opened 2019-08-08
Add Dataset Page 4: Overall page adjustment


t83714:
Adjust the layout & spacing around all elements on the page 4

Estimate: 3

## Acceptance Criteria
- Layout and spacing matches designs in [Zeplin](https://app.zeplin.io/project/5cff0c2b6308535da4f2dfe8/screen/5d0094fea239555dc29561d6)

t83714:
closed via https://github.com/magda-io/magda/pull/2434

PR #2434, opened 2019-08-27
Issue/2402 Updated security classification & sensitivity questions according to the new design

### LINKED ISSUE #2402: Add Dataset Page 4: What data sensitivity and access restrictions should be applied to this dataset?
Adjust style to match the design & added alert box that will be shown according to the selected option.

Estimate: 3

Update2nd Sep 2019:

- Multiple values should be allowed for Sensitivity markers

Update 26th Aug 2019:

Latest design notes:

[sharpmfd@csiro.au_20190826_153532.pdf](https://github.com/magda-io/magda/files/3539543/sharpmfd%40csiro.au_20190826_153532.pdf)

image

Summary:
- The first tooltip design see the new design notes
- The tooltip will contain a link that points to a detail info page. This page contains the table below (from https://www.protectivesecurity.gov.au/information/sensitive-classified-information/Pages/default.aspx):

image

Screen Shot 2019-08-26 at 2 18 55 pm

- The second question will only be shown if the first one select `Senstive`
- Tooltip should use this one:

image

#### Additional information:

Original Invasion design screenshot:

Screen Shot 2019-08-26 at 2 18 55 pm

## Acceptance Criteria
- There are two tooltips in the design - the first one for security classifications, rather than having a big bunch of text like in the design, should link to a page (like the one we have for data quality) which contains the text inside the tooltip that's in the design. It should keep the first paragraph, but then say "For more information visit <link>" where <link> is a Magda page with details about security classifications (e.g. https://x.magda.io/pages/security-classifications). The second tooltip (for sensitivity) is all good.
- If "protected" is selected, it should show the protected warning (see design)
- if "secret" or "top secret" is selected, it should show the "secret" warning (see design)

t83714:
### What this PR does

Fixes #2402
- Updated security classification & sensitivity questions according to the new design
- Also made sure the overall layout are the same as new design #2403 
- Updated CSV connector `information-security` aspect mapping logic based on new question rules:
  - `disseminationLimits` only available when `classification` === "OFFICIAL:SENSITIVE"

### Testing
- Tested locally & in Cluster

Test Site: 

https://issue-2402.dev.magda.io

### Checklist


AlexGilleran:
A few problems:

#### Page Goes Blank After Submitting
I'm not sure why, but immediately after submitting the screen goes blank (everything under `<div class="root">` is gone. This doesn't happen on dev.magda.io, there's nothing in the console though :/.


#### Should it be possible to have > 1 sensitivity markers?

image

This only allows 1 at a time :(.

#### Also the link in the sensitivity marker tooltip goes bold, which causes the whole tooltip to expand which is a bit weird:
image
image

#### Conversely, the link in this tooltip doesn't have any hover styling at all... I think the cursor should at least change to indicate a clickable state?
image

t83714:
@AlexGilleran --- I think it's ready to have a look again:

1> I updated the sensitive markers question so it's a multi-select question now.

2> Blank screen was caused by the edit page --- it still expects multi-values sensitive markers (string[]). 
We won't have the problem now as the question value is saved as `string[]` now

3> I can re-produce the tooltip style on firefox, Safari or Chrome.
For me, when hover, you see:
- underline of the link is removed
- the cursor changes to a hand cursor
I merged the master before looked into this issue. Probably the master fixed it 😅 
If it still happens to you, please let me know~

PR #2444, opened 2019-08-30
Added esri groups policies.

### LINKED ISSUE #2436: Add group-based registry access control policies
An Esri portal has group-based access control to each data item. Magda authz needs to replicate the same access control on harvested items.

A data item may allow  multiple groups access. A user may be in multiple groups. The policies will grant user access to a data item if the user's groups intercepts with the item's groups, something like
```
esri {
    input.object.registry.record["esri-access-control"].groups[_] == input.user.groups[_]
}
```

mwu2018:
### Add esri group-based opa policies

Fixes #2436 

With this group-based opa read policies, access control on registry records can be group-based.
This PR is built on top of PR #2432 

### Checklist


deploy/helm/magda/charts/opa/policies/object/registry/record/groups.rego
```diff
@@ -0,0 +1,5 @@
+package object.registry.record
```
AlexGilleran: Can we name this `esri-groups.rego` or something like that so it's obvious that it's going to use the `esri-access-groups` aspect? :)

mwu2018: done

PR #2454, opened 2019-09-06
Issue/2427 Add Custodian field (ie Division) to the 'People and Production' page

### LINKED ISSUE #2427: Add Custodian field (ie Division) to the 'People and Production' page
DAWR have identified the Custodian (aka Division responsible) is missing from Magda. This field is also used by the ONDC so is a transferable concept across organisations.

This is a field we could incorporate on the 'People and Production' page, ideally
- when a user selects the Team responsible for the dataset, the system automatically pre-populates the high level Division below in a field labelled 'Custodian' (pulling this from the org directory information)
- The Custodian should be an editable field, using similar style to organisation field above

IMG_3495

## Acceptance Criteria
- The custodian should be a dropdown that only allows preset values (no free text input a la "Organisation"
- The values to select from in the dropdown should be OrgUnits from the auth API at a certain level, which is configurable. E.g. if top level is "Root", second level is "Dept Agriculture", third level is "Division A", "Divison B", you can supply "3" to only show divisions, or supply no value to display all org units.
- The default value in the custodian dropdown should change based on what team is selected as the team responsible (i.e. a team will sit below a division, so go up the tree to get the division and pre-populate that)
- If a custodian has already been selected, it should not change based on the team being selected
- The value should be pre-populated based on what division the user is in if possible
- "Display data custodian area name" should be another option for showing contact point info (see @alyce's sketch)

## Technical Notes
- There should already be functionality in the auth api to both get the path between org units and to find org units at a certain level (good job @t83714!)

t83714:
### What this PR does

Fixes #2427 
Add Custodian field (ie Division) to the 'People and Production' page
- Added helm configure option `custodianOrgLevel` to magda-web-server chart
- New custodian question
  - Make sure before a user selects an option, a default option is selected based on `team` question
- Save `custodianOrgUnitId` to dataset access control aspect
- Updated `contactPointDisplay` options as per designed

### Checklist


magda-web-client/src/Components/Dataset/Add/Pages/People/CustodianDropdown.tsx
```diff
@@ -0,0 +1,102 @@
...
+                >
+                    Try Again
+                </button>
+            </div>
+        );
+    } else {
+        const selectedValue = orgUnitId && find(result, option => option.id === orgUnitId);
+
+        if ((!selectedValue || !hasUserSelected) && teamOrgUnitId) {
+            const relatedOrgUnit = find(
```
AlexGilleran: You could do the cast to `OrgUnitWithRelationship` here to avoid doing it twice later :)

magda-web-client/src/Components/Dataset/Add/Pages/People/CustodianDropdown.tsx
```diff
@@ -0,0 +1,102 @@
...
+                option =>
+                    option.relationship && option.relationship !== "unrelated"
+            );
+            if (
+                relatedOrgUnit &&
+                (relatedOrgUnit as OrgUnitWithRelationship).id !== orgUnitId
+            ) {
+                // --- we do need to send out onChangeCallback to update upper level state to make sure it's consistent with the screen value
+                // --- only update when we find a different value here otherwise we will trigger a infinite loop
+                onChangeCallback(
```
AlexGilleran: Is this because we only know what the default should be after we've called the API to find out what custodian org unit the user has a relationship with? Could it be done with `useEffect(..., [results])` ?

magda-web-client/src/api-clients/OrgUnitApis.ts
```diff
@@ -27,3 +27,35 @@ export async function listOrgUnits({
...
+ * Possible Value: 'ancestor', 'descendant', 'equal', 'unrelated'.
+ *
+ * @export
+ * @param {number} orgLevel - The level number (starts from 1) where org Units of the tree are taken horizontally.
+ * @param {string} [testOrgUnitId] - Optional; The org unit id that is used to test the relationship with each of returned orgUnit item.
+ * @returns {Promise<OrgUnitWithRelationship[]>}
+ */
+export async function listOrgUnitsAtLevel(
+    orgLevel: number,
+    testOrgUnitId?: string
```
AlexGilleran: Same thing as above re calling it `testOrgUnitId` :)

magda-authorization-api/src/createApiRouter.ts
```diff
@@ -453,6 +453,73 @@ export default function createApiRouter(options: ApiRouterOptions) {
         }
     });
 
+    /**
+     * @apiGroup Auth
+     * @api {put} /public/orgunits/listOrgUnitsAtLevel/:custodianOrgLevel/:testOrgUnitId List OrgUnits at certain org tree level
```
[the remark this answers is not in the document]
t83714: no problem~ Will update accordingly 👍

magda-web-client/src/Components/Dataset/Add/Pages/People/CustodianDropdown.tsx
t83714: Thanks 👍Will have a try~
Previously try to cast it to a (relatedOrgUnit | undefined) but seems typescript not smart enough to let me use `OrgUnitWithRelationship.id`

magda-web-client/src/Components/Dataset/Add/Pages/People/CustodianDropdown.tsx
t83714: 1> It's for the following scenario:
If we set a default value on screen and user is ok with the value, he probably won't touch the dropdown before click `next page` ---- for this case, the default value never saved to metaDataPage state.
2> What's the reason that we want to use a useEffect hook? 
You can tell this logic is under a conditional block. Thus, probably not good use case for hooks?

magda-web-client/src/api-clients/OrgUnitApis.ts
t83714: Will update accordingly 👍

t83714:
@AlexGilleran 
For some reason, typescript does not like:
```typescript
const selectedValue = orgUnitId && find(result, option => option.id === orgUnitId);
```
and only prefer:
```typescript
const selectedValue = orgUnitId && find(result, option => option.id === orgUnitId);
	if (orgUnitId && option.id === orgUnitId) return true;	
	else return false;	
});
```
Otherwise, it reports:
```
TypeScript error: Type '0 | { label: any; value: any; } | undefined' is not assignable to type 'ValueType<{ label: string; value: string; }>'.
  Type '0' is not assignable to type 'ValueType<{ label: string; value: string; }>'.  TS2322

    86 |                 }}
    87 |                 styles={ReactSelectStyles}
  > 88 |                 value={
       |                 ^
    89 |                     selectedValue && {
    90 |                         label: selectedValue.name,
    91 |                         value: selectedValue.id
```
Also see [gitlab log](https://gitlab.com/magda-data/magda/-/jobs/290431319).

Unless you have a better way of putting it, I am gonna change it back.

magda-web-client/src/Components/Dataset/Add/Pages/People/CustodianDropdown.tsx
AlexGilleran: The thing is it's equivalent to using `setState` in a render, which should [never](https://github.com/facebook/react/issues/5591#issuecomment-161678219) happen 😬 .

I was thinking the `useEffect` would sit at the top level of the render function and watch for changes to results - I think it'd also remove the need for `hasUserSelected`. Something like

```javascript
useEffect(() => {
    if (!orgUnitId) {
        onChangeCallback(getDefaultCustodianId(results, teamOrgUnitId))
    }
}, [results])
```

Does that make any sense? The idea being that because it's in a hook it wouldn't run on render, so there's no mixing the purpose of the render fn, and no (less?) danger of causing an infinite loop. The other alternative might be to do it as a `then` on the promise?

AlexGilleran:
> @AlexGilleran
> For some reason, typescript does not like:
> 
> ```ts
> const selectedValue = orgUnitId && find(result, option => option.id === orgUnitId);
> ```
> 
> and only prefer:
> 
> ```ts
> const selectedValue = orgUnitId && find(result, option => option.id === orgUnitId);
> 	if (orgUnitId && option.id === orgUnitId) return true;	
> 	else return false;	
> });
> ```
> 
> Otherwise, it reports:
> 
> ```
> TypeScript error: Type '0 | { label: any; value: any; } | undefined' is not assignable to type 'ValueType<{ label: string; value: string; }>'.
>   Type '0' is not assignable to type 'ValueType<{ label: string; value: string; }>'.  TS2322
> 
>     86 |                 }}
>     87 |                 styles={ReactSelectStyles}
>   > 88 |                 value={
>        |                 ^
>     89 |                     selectedValue && {
>     90 |                         label: selectedValue.name,
>     91 |                         value: selectedValue.id
> ```
> 
> Also see [gitlab log](https://gitlab.com/magda-data/magda/-/jobs/290431319).
> 
> Unless you have a better way of putting it, I am gonna change it back.

Oh sorry, I thought that was so simple I'd just do it rather than nag you to do it instead, but I guess not :(. I think changing the condition to `typeof orgUnitId !== "undefined"` will work, it's complaining because `orgUnitId` might be `0` which is falsy.

AlexGilleran:
(fixed)

PR #2483, opened 2019-09-13
With esri connector

### LINKED ISSUE #2440: Add group access aspect to harvested data items.
When data items are harvested from Esri portal, magda should query the Esri portal for each item's group access info then add "esri-access-control" aspect to the item.
If an item is public, no need to add any access control aspect.

Aspect "esri-access-control"  definition is in `magda-registry-aspects`, something like
```
{
    "$schema": "http://json-schema.org/hyper-schema#",
    "title": "Access Control information of an Esri data item",
    "description": "Access Control information of an Esri data item. Including: group IDs that are allow to read the data item.",
    "type": "object",
    "properties": {
        "groups": {
            "title": "Allowed access group IDs.",
            "type": "array",
            "items": {
                "title": "A group ID",
                "type": "string",
                "ids": [
                    {
                        "id": "{$}",
                        "rel": "item"
                    }
                ]
            }
        }
    }
}
```
The implementation might be on top of Esri connector https://github.com/magda-io/magda/pull/2336.

The usage of `esri-access-control` can be found in test class of RegistryAspectsSpec in `magda-registry-api/registry`.

Here is the workflow.
1. Create a record representing a dataset. This record probably does not need `esri-access-control` aspect so it is `public`. (We may need further discussion on if access control is needed here.)
2. Create distribution records and add `esri-access-control` aspect to each of them. The groups info is obtained from the esri portal.
3. Add distributions aspect to dataset record created in step 1.

mwu2018:
### What this PR does

Fixes #2440 
This PR is superseded by PR #2508 .
This is built on top of PR #2336 .
1. Harvested data will have access control aspect if required.
2. No unit tests.
3. If deployed to connect to nsw esri portal, search api and registry history api should not be 
   exposed because no access control have been implemented for those APIs.
4. No organization-details aspect because Esri Portal doesnt really capture 
   organisation in any meaningful way.

### Checklist


magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -4,6 +4,10 @@ import formatServiceError from "@magda/typescript-common/dist/formatServiceError
 import { ConnectorSource } from "@magda/typescript-common/dist/JsonConnector";
 import retry from "@magda/typescript-common/dist/retry";
 import request from "@magda/typescript-common/dist/request";
+import {
+    ESRI_NSW_ORG,
+    ESRI_NSW_PORTAL
```
rowanwins: These should be made generic (eg not NSW)

magda-typescript-common/src/session/SessionConsts.ts
```diff
@@ -0,0 +1,2 @@
+export const ESRI_NSW_ORG = "esri nsw authenticated user group";
+export const ESRI_NSW_PORTAL = "https://beta.portal.data.nsw.gov.au/arcgis";
```
rowanwins: Can we think of a different way to do this?

rowanwins:
superceded by #2508

Issue #2448, opened 2019-09-04
Large data preview - script timeout


gordjw:
**Describe the bug**
Previews of large data causes the page to freeze, making all other interactions impossible.

**To Reproduce**

1. Go to 'https://data.gov.au/dataset/ds-dga-487c604c-c93e-409e-90b9-6d1461c780ac/details?q=APS%20survey%202018'
2. Click on any download links (or wait until the Preview script kicks in, then click download)
3. Expect: nothing will be clickable, may have to 'Exit page'

**Expected behavior**
To be able to interact with components that are normally working, not having one script lock up the page.

**Screenshots**

**Desktop (please complete the following information):**

-   OS: MacOS
-   Browser: Chrome
-   Version: 75

**Additional context**
Maybe disabling preview for large datasets that won't load in a "reasonable" amount of time could help, could be a size based limit for simplicity.
Could also investigate ways to run Preview script in a separate thread.

Related to #1934

AlexGilleran:
Yeah this is actually a massive deal, it'd be really good to at least mitigate it.

> Maybe disabling preview for large datasets that won't load in a "reasonable" amount of time could help, could be a size based limit for simplicity.

This seems like a good idea, maybe we could do a HEAD request and pull it off the Content-Length header or something 🤔

t83714:
It seems it locks up at Papa Parse parsing staging. Probably, a simple fix could be turning on `Papa Parse`'s [worker mode](https://www.papaparse.com/#worker) so it will be done in a separate thread?

AlexGilleran:
Investigating solutions in #2455

AlexGilleran:
No longer locks up in dev (https://dev.magda.io/dataset/ds-dga-487c604c-c93e-409e-90b9-6d1461c780ac/distribution/dist-dga-b89fa786-77e7-4aee-abd8-70fc233dfd42/details?q=APS%20survey%202018), closing this.

PR #2489, opened 2019-09-18
Issue/2455 Visualisation Data file download & Processing Improvement

### LINKED ISSUE #2455: Improvement Visualization Frontend Performance
## Description
There are a few quick frontend visualization improvements we probably want to implement:
1> Turn on `papa parser` [worker](https://www.papaparse.com/faq#workers) option to enable data downloading & parsing to be done in a separate thread. Thus, it won't slow down the web user interface.

This only requires set worker option to `true` (and upgrade papa parse to 5.0)

2> As table & chart preview was developed separately, they actually load/parse data independently. 
i.e. if switch from chart view to table view, the data file will be loaded again.
We should lift the data loading & Parse code to higher level so that the data will be shared between two preview component.

3>  We should use `chunk` option of papa parse to stream the remote file and only partially download the file until `certain limit`. This doesn't rely on HTTP `head` request.

Those improvements above might help #2448 

## To Investigate
- Are any/all of these quick wins?
- If so we should do them, otherwise write a ticket describing the proper way to fix it and figure out a dodgier solution (e.g. not showing for big files)

t83714:
### What this PR does

Fixes #2455 
- Use worker to download & process CSV file (Thus, processed in a separate thread(s))
  - See below: requests with `gear` icon are triggered by web workers
image
  - CSV data parsing can utilise more CPU cores without freezing web UI
  - This will also help to also #2448 
  - CSP settings are updated to support this.
    - Please note: We only need `blob:` to make data load & parse worker work. But since we set `worker-src`, we also need 'self' for `cesiumWorker`. Also, added `blob:` to `script-src` as `safari` doesn't support `worker-src` yet (but will support soon). 
- Download & Data parsing is processed by chunks. (By default 2MB chunk size. Can be configured via `helm parameters` below.). 
  - i.e. When the server supports `Range` HTTP header, we will download & process data piece by piece, instead of downloading the whole file before process it.
  - This allows us can choose to download a portion (e.g. the first 200 rows) of the file if the file is too big (can be configured via `helm parameters` below)
  - `Content-Range` & `X-Content-Range` response headers are added to helmet cors config to support this
  - It seems `data.gov.au` server doesn't support the `Range` header correctly. 
    - e.g. for URL: https://data.gov.au/data/dataset/fa0b0d71-b8b8-4af8-bc59-0b000ce0d5e4/resource/35953a01-a9a8-4609-8566-c9fa7de465d3/download/credit_lic_201909.csv
    - Set Header: `Range: bytes=20-40` will get response header `Content-Range: bytes 20-40/8256106`. But still response whole file content (you need to disable cache to reproduce this in web browser). See screenshot:

image

- Because of this unexpected server behaviour, I have to use a forked version papaparse (`magda-io/PapaParse#magda`). My PR to the `papaparse` project has already been merged: [see here](https://github.com/mholt/PapaParse/pull/709). Once the new version is released (no set date yet), you can change back to the official version.

- Introduce the following Helm configuration Parameters:
  - `csvLoaderChunkSize`: download & parse chunk size. in bytes. Default: 2MB
  - `maxChartProcessingRows`: max. data rows that chart preview modules will try to render. If over this limit, chart preview will be disabled (instead of present the partial data) to avoid incomplete data presentation/confusion.
  - `maxTableProcessingRows`: max. data rows that table preview will show. Table preview will always show the no. of data rows beblow this limit. Excess data will be discarded.
- Chart & table previews modules used to load & parse the data separately. i.e. the same file will always be downloaded & processed again when you switch between chart & table. This PR refactored the code and make the two modules share the same data source.

### Test Site:

https://issue-2455.dev.magda.io/dataset/ds-dga-fa0b0d71-b8b8-4af8-bc59-0b000ce0d5e4/details?q=*

Tested the above URL on chrome, safari, firefox & IE11 (using [chrome IE tab](https://chrome.google.com/webstore/detail/ie-tab/hehijbfgiekmjfkfjpbkbammjbdenadd))

### Checklist


magda-web-client/src/Components/Common/DataPreviewChart.tsx
```diff
@@ -55,122 +74,100 @@ class DataPreviewChart extends Component {
...
-            throw e; //--- not capture here; only for debug
+        this.chartDatasetEncoder.processData(
+            this.props.distribution,
+            this.props.dataLoadingResult
+        );
+
+        let chartType = this.state.chartType;
+        if (!this.state.xAxis || !this.state.yAxis) {
+            this.chartDatasetEncoder.setDefaultParameters();
+            //if(this.chartDatasetEncoder.yAxis.isAggr) chartType = "pie";
```
AlexGilleran: Does this need to stay in?

magda-web-client/src/Components/Common/DataPreviewChart.tsx
```diff
@@ -199,26 +196,46 @@ class DataPreviewChart extends Component {
                     </button>
                 </AUpageAlert>
             );
-        if (this.state.isLoading) return <Spinner height="420px" />;
+        if (this.props.isLodaing) return <Spinner height="420px" />;
```
AlexGilleran: `isLoading` 🤔

magda-web-client/src/Components/Common/DataPreviewChart.tsx
t83714: removed 👍

magda-web-client/src/Components/Common/DataPreviewChart.tsx
t83714: Thank you~ updated 👍

t83714:
@AlexGilleran 
I think it's good to have a look again~

PR #2490, opened 2019-09-18
Issues/2460 colour coordinated datepicker

### LINKED ISSUE #2460: Datepicker in add dataset should be colour coordinated
**Describe the bug**
In the add dataset flow we have a date picker that works quite nicely, but is still using the default teal colour. We should make it match the main colour used by the rest of the system (purple for DGA, sand for Agriculture, etc).

**To Reproduce**

1. Go to /dataset/add
2. Continue through the flow until you get to "Dates and Updates"
3. Note the colour on the datepicker used for "When was the dataset first issued" and "When was the dataset most recently modified".

**Expected behavior**
I expect the teal to be replaced with the main colour in the rest of the UI

**Screenshots**
image

AlexGilleran:
### What this PR does

Fixes #2460 

This makes the react-dates datepicker uses Magda colours, as set through the SCSS compiler config.

As part of this I made the default config colours a bit more organised - you should now be able to set the whole colour scheme by just specifying a primary and secondary colour.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -95,6 +95,7 @@ UI:
 -   Fixed edit file panel text input layout
 -   Fixed a blank screen issue on dataset page
 -   Fixed "NOT SET" appears on the dataset page for non-admins
+-   Made the datepicker for add dataset use the correct colours.
 
 Gateway:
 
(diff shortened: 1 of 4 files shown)


AlexGilleran:
o_0 what have I done, it's not supposed to be blue!

I think I've got it down the files I originally tried to commit now 😂

t83714:
Well done~ Work great now 👍

PR #2508, opened 2019-09-23
Improve access policy

### LINKED ISSUE #2440: Add group access aspect to harvested data items.

### LINKED ISSUE #2495: Let esri item owner see the item
In an esri portal, if an esri item (record) has access attribute of "private", the item can only be viewed by the owner, that is, not view-able by any groups. However, the current magda esri access control policy is group-based, which implies that the private item's owner will not have access to the item via magda registry api.

We need to add esri owner-based access policy to magda. The technical steps may include
1. Add new policy to opa.
2. In `esri-access-control` aspect, in addition to `groups` property,  add property `owner`.
3. When harvesting a private item from the esri portal, add `esri-access-control` aspect with `owner`. The value of `owner` 
4. When an esri user login, save the `profile.username` or `profile.email` (TBD. See note) as `esriUser` in the session, in addition to `esriGroups`.

Note:
Query for esri item  (with service account):
```
https://beta.portal.data.nsw.gov.au/arcgis/sharing/rest/content/items/<item-id>?f=json
```
which will get `owner` property of value `user-id`. Assuming `user-id` is unique, it can be assigned to `owner` property of `esri-access-control` aspect for the item.

Alternatively  we can further query for esri user
```
https://beta.portal.data.nsw.gov.au/arcgis/sharing/rest/community/users/<user-id>?f=json
```
which will get `email` property. An `email-address` is unique. Assign `email-address`  to `owner` property of `esri-access-control` aspect for the item.

mwu2018:
### What this PR does

Fixes #2440, #2495, #2517 
This PR also supersedes PR #2483 and #2336 .
1. Add `esri_owner` policy. An esri record always has property `{"owner": <ownerId>}` in `esri-access-control` aspect.  Owner always has access to his own records. 
2. Add `esri_public` policy. An esri public record always has property ` {"access": "public"}` in `esri-access-control` aspect. Both anonymous and authenticated users will have access to esri public records.
3. Add `admin_role` policy. A user with magda admin role will has access to any records.
4. The esri access policy id is changed from `object.registry.record.esri_groups` to `object.registry.record.esri_owner_groups`.
5. Both `owner_orgunit` and  `esri_owner_groups` policies include `admin_role` policy.
6. An esri item's access attribute is one of the following `public`, `private`, `org` and `shared`. In magda, an `org` or `shared` esri record must have `groups` property in `esri-access-control` aspect; a `private` esri record must have `owner` property in `esri-access-control` aspect; a `public` esri record must have 'access' property of value 'public' in `esri-access-control` aspect.
7. Any records without access control aspect are for public access.
8. Esri login session will have `esriUser` property in addition to `esriGroups`.
9. Fixed bug when reindexing (maybe also minions). Because of 5), a system request (tenant id == -1) having admin role credentials will have access to all records.
10. In order to use esri connector, the gateway and esri connector must be configured to set `esriOrgGroup` to a string without spaces. See `deploy/minikube-esir-dev.yml` for example. 
### Checklist


magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -0,0 +1,563 @@
...
+                                        item,
+                                        distUri
+                                    );
+                                } catch (err) {
+                                    console.log(
+                                        `Broke on item url: ${
+                                            item.url
+                                        }, dist uri: ${distUri}`
+                                    );
+                                    console.log(err);
```
AlexGilleran: This seems like it should be `console.error` ?

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -0,0 +1,563 @@
...
+                                            distUri
+                                        );
+                                    }
+                                } catch (err) {
+                                    console.log(
+                                        `Broke on item url: ${
+                                            item.url
+                                        }, dist uri: ${distUri}`
+                                    );
+                                    console.log(err);
```
AlexGilleran: This should be a `console.error` too.

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    private requestDataSearchPage(
+        url: uri.URI,
+        startIndex: number
+    ): Promise<EsriPortalDataSearchResponse> {
+        const pageUrl = url.clone();
+        pageUrl.addSearch("start", startIndex);
+        pageUrl.addSearch("num", this.pageSize);
+        const that = this;
```
AlexGilleran: We shouldn't have to do this with arrow functions (`this` is maintained)

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -0,0 +1,563 @@
...
+        const operation = async () =>
+            new Promise<EsriPortalDataSearchResponse>((resolve, reject) => {
```
AlexGilleran: Is this extra `Promise` really needed given that it's an async function? You should be able to just `return` the value and it'll still end up as a promise.

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -0,0 +1,563 @@
...
+                const requestUrl = pageUrl.toString();
+                console.log(
+                    `Requesting start = ${startIndex}, num = ${this.pageSize}`
+                );
+
+                request(
```
AlexGilleran: Rather than another layer of callback, could this be an `await`? I think there's a promisified `request` library that we're using somewhere else, or we could use node-fetch?

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -0,0 +1,563 @@
...
+                            return;
+                        }
+
+                        // A single portal item only has one distribution in the form of a url.
+                        // That url however may represent a single layer, or a map (containing multiple layers)
+                        // The bulk of the distribution information needs to be retrieved
+                        // from the map or feature service endpoint
+                        // An individual item may result in one or many additional requests
+                        for (let i = 0; i < body.results.length; ++i) {
+                            const item = body.results[i];
```
AlexGilleran: This is maybe a question of style, but this is a really long loop and I find it a bit hard to follow the flow. Is there any chance of breaking this up into a few more functions and getting rid of the `continue`s?

So it's more like
```javascript
if (isShared(item)) {
    return processSharedItem(item);
} else if (isPublic(item)) {
    return processPublicItem(item);
} else if (isPrivate(item)) {
   return processPrivateItem(item);
} else {
 // etc
}
```

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -0,0 +1,563 @@
+import AsyncPage from "@magda/typescript-common/dist/AsyncPage";
+import EsriPortalUrlBuilder from "./EsriPortalUrlBuilder";
+import formatServiceError from "@magda/typescript-common/dist/formatServiceError";
+import { ConnectorSource } from "@magda/typescript-common/dist/JsonConnector";
+import retry from "@magda/typescript-common/dist/retry";
+import request from "@magda/typescript-common/dist/request";
+import {
+    ESRI_NSW_ORG,
+    ESRI_NSW_PORTAL
+} from "@magda/typescript-common/dist/session/SessionConsts";
```
AlexGilleran: Can we pull this out into something a bit more generic that could be applied to more than just NSW?

If this still needs client-specific stuff hard-coded into it then it's not ready for the core Magda codebase yet, it should probably stay in terriajs/magda.

mwu2018: I will try to make this configurable.

mwu2018: done

magda-esri-portal-connector/src/EsriPortal.ts
rowanwins: Have brought in node-fetch

magda-esri-portal-connector/src/EsriPortal.ts
rowanwins: Promise removed

magda-esri-portal-connector/src/missing.d.ts
```diff
@@ -0,0 +1 @@
+declare module "node-fetch";
```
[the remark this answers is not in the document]
rowanwins: Ha didn't think of that (because I know next to nothing about ts)

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -474,44 +474,48 @@ export default class EsriPortal implements ConnectorSource {
...
+    //     } catch (err) {
+    //         console.error(
+    //             `Broke on item url: ${
+    //                 item.url
+    //             }, dist uri: ${distUri}, layer id: ${lyr.id}`
+    //         );
+    //         console.error(err);
+    //     }
+    // }
+
```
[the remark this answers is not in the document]
rowanwins: I left this here because the logic of the connector changed and we still may want to do this at some stage, basically it's collecting additional distributions associated with a dataset...

Happy to cull it though if that's your preference. Or I could at least add some comments as to why it's there.

magda-typescript-common/src/JsonTransformer.ts
```diff
@@ -8,7 +8,7 @@ import createServiceError from "./createServiceError";
  * A transformer takes source data and transforms it to registry records and aspects.
  */
 export default abstract class JsonTransformer {
-    private sourceId: string;
+    public sourceId: string;
```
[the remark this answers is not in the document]
rowanwins: again blame the ts novice

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala
```diff
@@ -1636,55 +1636,74 @@ object DefaultRecordPersistence
       """
   }
 
+  /**
+    * When tenantId equals SYSTEM_TENANT_ID, if the request also provides JWT that can prove
+    * that the requester has admin role, this function will return List(sqls"true"). In that
+    * case, the requester will have access to all records.
+    */
```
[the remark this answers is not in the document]
mwu2018: This has been removed in the next PR #2521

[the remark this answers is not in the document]
rowanwins: This one is for you @mwu2018

magda-esri-portal-connector/src/EsriPortal.ts
```diff
@@ -407,6 +300,111 @@ export default class EsriPortal implements ConnectorSource {
...
+                    return;
+                }
+
+                await this.processRootMapServiceAsDistribution(
+                    item,
+                    distInfo,
+                    distUri
+                );
+
+                // const layersLength = distInfo.layers
```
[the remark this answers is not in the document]
rowanwins: Ah so this is related to your comment here
https://github.com/magda-io/magda/pull/2508#discussion_r333321863

This was the loop which called the function.

mwu2018:
> A few comments left in and such - also looks like the registry tests are actually failing no matter how many times I restart them 😬

My new PR #2521 includes everything from this PR and has passed all tests. Don't know why.

magda-esri-portal-connector/src/EsriPortal.ts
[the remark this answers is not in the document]
AlexGilleran: Yeah happy to leave it in as long as there's a comment :)

PR #2516, opened 2019-09-25
Made indexer look for tenant 0 instead of tenant -1 when crawling.

AlexGilleran:
### What this PR does

This fixes the problem we're seeing where the indexer reindexing results in 0 items in the search index. Currently all the datasets get put into the registry under tenantid = 0, but indexer looks for them in tenant id = -1. This puts both on `0`. I'm not sure if that's how it's supposed to be, but it works.

### Checklist

-   [x] No changes, this makes the indexer do what it already says it does in changes.
-   No zenhub issue.


AlexGilleran:
Closing, this is apparently just papering over the actual bug.

PR #2521, opened 2019-10-09
Support more complicated OPA

### LINKED ISSUE #2515: Add esri timestamp policy
Magda mirrors `point-in-time` esri portal access control policy. We need to add an `esri-time` policy so that an esri record (public or not) will not be accessible if its `esri-access-control` aspect has expired.

mwu2018:
### What this PR does

Fixes #2515 
This PR is built on top of PR #2508 . Because it contains substantial code changes, I'd suggest that it should be reviewed after PR #2508 is merged to the master.

1. Re-implement record opa parser/queryer to support more complicated OPA policy, which is required by esri expiration policy.
2. Move all OPA related code from RecordPersistence class to a helper class, which makes the code clean and easy to maintain.
3. Most of unused code are removed.
4. Esri access policy also checks mirrored authz info expiration time.

### Checklist


AlexGilleran:
Looks OK from the code, just need to give it a test.

PR #2523, opened 2019-10-09
Issue/2491 Querying formats in upper case causes two formats to be selected

### LINKED ISSUE #2491: Querying formats in upper case causes two formats to be selected
**Describe the bug**
If one uses either the URL or the format dropdown to put in a format in UPPER CASE, the result is that two "CSV" values get selected in the formats dropdown. I'm guessing that this is because it's searching for both "CSV" and "csv" somehow, and putting both in the list, then they're all getting converted back to lower case for display.

**To Reproduce**
Either just go to https://dev.magda.io/search?format=CSV, OR to get there from scratch, go to the search, go to the format dropdown and manually type in "CSV" (all upper case) and hit enter, WITHOUT selecting it from the dropdown.

**Expected behavior**
It should display "CSV" because it's only searching for one format.

**Screenshots**
image

**Desktop (please complete the following information):**

-   OS: MacOS
-   Browser Firefox 69

t83714:
### What this PR does

Fixes #2491 

Querying formats in upper case causes two formats to be selected

#### Cause
- When merging aggregation result with input options, the option value comparison should be case insensitive (as the value from es will be lower case)

### Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -107,7 +107,7 @@ buildtest:search-with-index-cache:
     - docker-compose up -d
     - cd ..
     - "ls $CI_PROJECT_DIR/sbt-cache/ || cp -R /sbt-cache/* $CI_PROJECT_DIR/sbt-cache/"
-    - sbt -DelasticSearch.serverUrl=elasticsearch://docker:9200 "intTest/testOnly au.csiro.data61.magda.api.*Spec"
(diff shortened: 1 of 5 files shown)


t83714:
Test cases are added~ Now is ready to have a look again~

t83714:
The test cases previously had some bugs and have been fixed now~

PR #2541, opened 2019-10-16
Fixed: Requesting records for all tenants returns no records

### LINKED ISSUE #2359: A system tenant should get record summaries of all tenants
### Problem description
The current RecordPersistence.getSummaries should return all record summaries of all tenants if the requester is a system tenant.

### LINKED ISSUE #2360: A system tenant should get count of all tenants
### Problem description
If a system tenant gets /v0/records/count, the result will be correct.
However, if a system tenant gets /v0/records/count?aspect=$aspectId, the result will always be 0.

### LINKED ISSUE #2517: Requesting records on tenant id -1 results in an empty response
**Describe the bug**
Currently if a reindex is requested/automatically happens in the indexer, it'll end up wiping every record out of the search index.

After some ~messing about~ investigation, turns out that this is because the requester requests records from tenant id -1. This _should_ result in all records being returned regardless of tenant, however it results in no records being returned - exactly the same as if the registry was completely empty.

**To Reproduce**
Do a request to `v0/records` on the registry API with `X-Magda-Tenant-Id` set to `-1`. You can confirm that it's not working if doing the same request with tenant id `0` actually returns stuff.

**Expected behavior**
All records on all tenants should be returned.

AlexGilleran:
### What this PR does

Fixes #2517 
Fixes #2359 - Because the system tenant now gets all summaries
Fixes #2360 - Because the system tenant now gets all counts

On its surface this fixes the problem we're seeing where the indexer sees 0 datasets when it goes to reindex, but I went further and refactored the whole way that multi-tenancy is picked up in the registry API and passed through to the database.

Currently, requests to the registry either pass a tenant id or -1, which means to return results for all tenants. In general this is tested for (like `if tenantId != -1 then filter sql`), but there's a few cases where this has slipped through, and `WHERE tenantId = -1` is making it through to the SQL statements, which will always result in ean empty result set.

So I've tried to make it super super super hard to make that mistake. Now in the APIs there's a directive that turns the incoming tenant id header into an instance of a `TenantId` sealed trait which can either be `SpecifiedTenantId` or `AllTenantsId`. I can't find a way to make it break on compilation if this is ever interpolated into SQL, but it will fail at runtime and log an error complaining that it doesn't know how to interpolate it.

Also:
- Converting from a TenantId to SQL is now done via a function in `SQLUtil`, which either turns it into `Table.tenantId = ${tenantId}` or `true` depending on whether the tenant id is specified or not
- Completed some TODOs around tenant ids (e.g. for summaries)
- Wrote a test specifically testing for GET /records with tenant id = -1
- Did a bunch of formatting (sorry!)
- Moved RegistryExternalInterface into magda-indexer seeing as it's got indexer-specific classes in it anyway

### Checklist


magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala
```diff
@@ -393,7 +394,11 @@ object DefaultRecordPersistence
           from RecordAspects
           inner join Aspects using (aspectId, tenantId)
           inner join Records using (recordId, tenantId)
-          where (RecordAspects.recordId, RecordAspects.aspectId, RecordAspects.tenantId)=($recordId, $aspectId, $tenantId)
+where (RecordAspects.recordId, RecordAspects.aspectId)=($recordId, $aspectId) AND ${SQLUtil
+      .tenantIdToWhereClause(
+        tenantId,
+        Some(SQLSyntax.createUnsafely("RecordAspects"))
```
t83714: Why pass `SQLSyntax.createUnsafely("RecordAspects")` instead of "RecordAspects"?
In `tenantIdToWhereClause`, It seems we pass the content through `SQLSyntax.createUnsafely` as well?

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RegistryApp.scala
```diff
@@ -60,7 +60,7 @@ object RegistryApp extends App {
   logger.info("Starting MAGDA Registry")
 
   GlobalSettings.loggingSQLAndTime = LoggingSQLAndTimeSettings(
-    enabled = false,
+    enabled = true,
```
t83714: Will this print SQL for every requests? Do we want this for production?

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala
AlexGilleran: There's a bit of explanation in the scaladoc - essentially because if it just accepted a String, the person calling the function might not realise that the content of that string is going to be passed unsafely into SQL. By requiring an SQL statement, they have to use `createUnsafely` which means they'll be aware to make sure user input doesn't get passed in (hopefully).

The other alternative is to accept an enum or something, but then the enum needs to be updated whenever a table name changes.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RegistryApp.scala
AlexGilleran: Oops

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala
t83714: 👍

PR #2542, opened 2019-10-16
Issue/2539 Fixed CSV loader didn't retry the different line ending correctly

### LINKED ISSUE #2539: SoE dataset CSVs can't be displayed as charts anymore
**Describe the bug**
Since the new chart changes (which are great in general!), there's at least one chart that worked before that doesn't now:

On dev.magda.io:
image

On data.gov.au:
image

The error in the console says:

```console
retry new newline... backend.js:6:2315
    r backend.js:6
    chunk CsvDataLoader.ts:171
    _ papaparse.min.js:744
```

**To Reproduce**
Go to https://dev.magda.io/dataset/ds-dga-5076fbb3-4cbb-47f4-9121-8e8207744f4b/details?q= and scroll down to where the chart should be.

Compare to https://data.gov.au/dataset/ds-dga-5076fbb3-4cbb-47f4-9121-8e8207744f4b/details?q=

**Expected behavior**
Even though it's not a great chart, we should still be able to display one as per the old code on data.gov.au.

t83714:
### What this PR does

Fixes #2539 

Fixed CSV loader didn't retry the different line ending correctly
- Cause & Fix: 
  - When retry parsing with manual line ending setting, we called `papa.abort()` to end possible future work.
  - `papa.abort()` will trigger `complete` event as well which will cause our promise to return an empty result, instead of waiting for next round parsing result.
  - We set a new field to allow `complete` to be skipped for this situation.
- Also, upgrade papaParse to new release 5.1.0 (which has included our PR for the previous fix)

### Checklist


t83714:
@AlexGilleran 
This issue is actually ready to review~
The stuck checking is Snyk 😢

magda-web-client/src/helpers/CsvDataLoader.ts
```diff
@@ -111,6 +121,7 @@ class CsvDataLoader {
         this.isLoading = false;
         this.toBeAbort = false;
         this.isPartialData = false;
+        this.skipComplete = false;
```
AlexGilleran: isn't this line not needed because you've already initialized it above?

magda-web-client/src/helpers/CsvDataLoader.ts
```diff
@@ -192,6 +206,10 @@ class CsvDataLoader {
                 }).bind(this),
                 complete: (() => {
                     try {
+                        if (this.skipComplete) {
+                            this.skipComplete = true;
```
AlexGilleran: Why set it to true if it's already true?

magda-web-client/src/helpers/CsvDataLoader.ts
```diff
@@ -71,6 +71,16 @@ class CsvDataLoader {
      */
     private toBeAbort: boolean = false;
 
+    /**
+     * We set this property to skip the next `complete` call.
+     * This would be usuful when we want to retry the data downloading.
```
AlexGilleran: I still don't quite get why we skip the next complete call - is it because otherwise `complete` will get called with the wrong data?

t83714: What happened was:
- We determined that, somehow, PapaParser's auto line ending detection doesn't work for the file (Around line 182) and planned to re-process the file with fixed line-ending setting "\n".
- We firstly, called `parser.abort();` and then call `resolve(retryLater(...))` to restart the whole process
  - Unfortunately, `parser.abort();` will make PapaParser trigger `onComplete` event and call my `complete` method as well (I thought it won't previously)
  -  in the `complete` method, the `resolve(result)` is called and make the promise be resolved earlier while the second retry still processes the file.

magda-web-client/src/helpers/CsvDataLoader.ts
t83714: I just want to be safe as `this.skipComplete` should be false at the beginning of the loading process.
The loading process could be triggered more than once for different reasons (e.g. retry line ending)

magda-web-client/src/helpers/CsvDataLoader.ts
t83714: Good pick-up, it should be `this.skipComplete = true;`
It still works as it's been reset in `resetDownloadData` as well (i.e. I was lucky 😄 ). 
I will correct this~

magda-web-client/src/helpers/CsvDataLoader.ts
AlexGilleran: Oh right sorry, I thought this was in the constructor 😬

PR #2546, opened 2019-10-18
Publisher id saving

### LINKED ISSUE #2505: Add dataset page Does not Save Publisher ID
**Describe the bug**

Add dataset page Does not Save Publisher ID

**How to reproduce it**

Create a new dataset and select the `responsible organization`.

Once submitted, check the registry api response (as the screenshot below):

image

#### Possible Cause:

Line: https://github.com/magda-io/magda/blob/689469085d940853ba70ab262891e7c3937113aa/magda-web-client/src/Components/Dataset/Add/DatasetAddMetadataPage.tsx#L285

Seems an `await` keyword is missing before `getOrgIdFromAutocompleteChoice`

sajidanower23:
### What this PR does

Fixes #2505 

Await for resolution of publisher id.

### Checklist


magda-web-client/src/config.ts
```diff
@@ -17,7 +17,7 @@ const fallbackApiHost = "https://dev.magda.io/";
 
 const DEV_FEATURE_FLAGS = {
     cataloguing: true,
-    previewAddDataset: true
```
t83714: probably don't change the default value here

sajidanower23: Done

PR #2547, opened 2019-10-18
Add hover text to tooltip for dataset date

### LINKED ISSUE #2465: (?) tooltip icon next to Add Dataset first issued field does nothing
**Describe the bug**
Next to "When was the dataset first issued?", there's a tooltip icon, but when hovered over nothing comes up. Presumably there should be some information about what a first issued date does, otherwise we should remove it.

**To Reproduce**
Go to /dataset/add and click through to the first page.

**Expected behavior**
A tooltip comes up that displays "Use the date the dataset was first created or issued for release".

**Screenshots**
image

**Technical Notes**
See other examples of tooltips in the frontend.

sajidanower23:
### What this PR does

Fixes #2465 

Added a hover text for the date tooltip.

### Checklist

dataset_date_hovertext

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -114,7 +114,8 @@ UI:
 -   Fixed mobile views incorrect min. width
 -   Drafts should be ordered by date on Drafts list page
 -   Changed text to reflect state/territory/country accordingly
--   Save publisher id in the database
+-   Saved publisher id in the database
+-   Added hover text for the tooltip beside the date-picker in the Add Dataset page
 
 Gateway:
 
--- a/magda-web-client/src/Components/Dataset/Add/Pages/DetailsAndContents/index.scss
+++ b/magda-web-client/src/Components/Dataset/Add/Pages/DetailsAndContents/index.scss
@@ -67,6 +67,10 @@
                 color: #484848;
                 position: relative;
             }
+            .inner {
+                width: 400px;
+                left: 15px;
+            }
         }
 
         .question-recent-modify-date {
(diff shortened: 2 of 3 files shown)


alyce:
I'm not sure about the placement of this tooltip, it seems a little disjointed from the tooltip icon with the placement, and the plain rectangle (without an arrow tip) makes this seem even more so. 

This might be more general feedback that may need to be applied more generally across all screens, paging @AlexGilleran to check. 

I would suggest:
- When we implement tooltips, ideally they should appear directly above or below the icon
- Tooltips should have a little arrow lip on them pointing to the tooltip icon so they're clearly associated with the icon
- Can we also please make sure they are keyboard accessible, otherwise we'll fail our accessibility requirements.

See here for some examples and tips: [https://www.nngroup.com/articles/tooltip-guidelines/](https://www.nngroup.com/articles/tooltip-guidelines/)

t83714:
@sajidanower23 We have an example of a more general tooltip on page 4 (see screenshot below):
image
@alyce Can you help to check whether that one is ok in terms of a11s?
You can find it on this demo site as well (no login is required): https://demo.dev.magda.io/

alyce:
That purple tooltip looks good, I'm happy with how that looks so we can implement it everywhere, BUT it is not keyboard accessible :(

I just did a quick google, there is plenty of stuff online about accessible tooltips, such as [this page](https://van11y.net/accessible-simple-tooltip/)

I suspect we'll need to make the ? icon focusable so we can let it have keyboard accessibility

sajidanower23:
@alyce  @t83714 This is how the tooltip text looks now. Thoughts?
tooltip

alyce:
Looks great @sajidanower23 😁

t83714:
The new a11s ticket for the tooltip is here:
https://github.com/magda-io/magda/issues/2548

PR #2550, opened 2019-10-22
Issue/2001 Added the feature of validating aspect data against JSON Schema to Registry API

### LINKED ISSUE #2001: Aspects aren't being validated against schemas strictly enough
### Problem description
Related to #2000 - the data.json is writing "keyword" instead of "keywords" to `dcat-dataset-strings`, but this should be disallowed by the schema validation, as the schema only allows "keywords".

### Acceptance Criteria
- `PUT`ing or `POST`ing a valid aspect to the registry via posting to `/records` or `records/x/aspects` should continue to work as-is
- `PUT`ing or `POST`ing an _invalid_ aspect to the registry (according to its current schema) via posting to `/records` or `records/x/aspects` should return `400`.
- When `PUT`ing or `POST`ing an entire record made up of multiple aspects, a single invalid aspect should cause the entire request to fail and have no effect on the state of the system. Even the valid aspects should not be saved.
- `PUT`ing or `POST`ing a valid record / record aspect should work as it does now.
- This should be able to be turned on or off, so that it doesn't end up blocking us if we find that every single connector, minion etc is posting invalid data (which is _relatively_ likely :( )

### Technical Notes
- I'm not sure to what extent this was attempted in the past, it might mean creating the whole mechanism from scratch.

### Out of scope
- Figuring out what's broken about the existing records, we can worry about that later.

t83714:
### What this PR does

Fixes #2001 

Added the feature of validating aspect data against JSON Schema to Registry API
- Default to off.  Turn it on by setting `validateJsonSchema` = true

~Just realise Synk doesn't like the older version JSON schema lib I used. To upgrade to the latest version, I need to upgrade a lot of other libs to the newer version as well (seeing conflict version error). I guess the similar work has been done in PR https://github.com/magda-io/magda/pull/2543. Will probably wait till that is merged~

Merged PR https://github.com/magda-io/magda/pull/2556 to fix the dependencies issue. Probably should merge that PR first~

### Checklist


magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/AspectValidator.scala
```diff
@@ -0,0 +1,126 @@
...
+              validateWithAspectPatch(JsonPatch(Add(rest, aValue)), recordId, aspectId, tenantId)(session, config)
+            }
+          case (
+            Some(aspectId),
+            List(Remove("aspects" / (_ / rest), old))
+            ) =>
+            if(rest != Pointer.Empty) {
+              validateWithAspectPatch(JsonPatch(Remove(rest, old)), recordId, aspectId, tenantId)(session, config)
+            }
+          // We patch in all other scenarios.
```
AlexGilleran: Can we have a comment explaining what scenarios we don't patch in? I'm not sure how the logic works here

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala
```diff
@@ -777,7 +820,7 @@ object DefaultRecordPersistence
         // we don't end up with an extraneous record creation event in the database.
         case None =>
           DB.localTx { nested =>
-            createRecordAspect(nested, tenantId, recordId, aspectId, newAspect)
+            createRecordAspect(nested, tenantId, recordId, aspectId, newAspect, config, true)
```
AlexGilleran: Why force skip validation here?

t83714: We've done the validation earlier (a few lines above) and try to avoid re-do the validation.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/AspectValidator.scala
t83714: This logic is copied from here (including the comment 😆):
https://github.com/magda-io/magda/blob/64b58ff8db062cd5a260674514cf9d8cef82da2e/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala#L652

Reason: 
As the validation needs to be done before all patches are processed (therefore, we won't half do the data if one of the patches creates invalid data): 
- we need to apply the full patch to create a patched JSON object from original data. 
- And Then validate on this patched version.
See here: https://github.com/magda-io/magda/blob/64b58ff8db062cd5a260674514cf9d8cef82da2e/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala#L543

I probably should consolidate this code into a separate file and share with RecordPersistence.scala?

t83714:
> Had a read through, just a few comments.
> 
> In addition, is it possible to add some testing of the actual API to make sure that they're using the validator if it's turned on?

The test cases are here:
https://github.com/magda-io/magda/blob/issue/2001/magda-registry-api/src/test/scala/au/csiro/data61/magda/registry/JsonSchemaValidationSpec.scala

The validation is turned on here:

https://github.com/magda-io/magda/blob/64b58ff8db062cd5a260674514cf9d8cef82da2e/magda-registry-api/src/test/scala/au/csiro/data61/magda/registry/JsonSchemaValidationSpec.scala#L23

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/AspectValidator.scala
AlexGilleran: Oh right - yeah that'd be great :)

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala
AlexGilleran: Ah right - that could use a comment :)

AlexGilleran:
> The test cases are here:
> https://github.com/magda-io/magda/blob/issue/2001/magda-registry-api/src/test/scala/au/csiro/data61/magda/registry/JsonSchemaValidationSpec.scala

Oh right... for some reason I thought that was testing the validator directly but that looks good, sorry!

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/AspectValidator.scala
t83714: I've refactored the code around the patch procedure~

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala
t83714: Added more comments~

t83714:
@AlexGilleran 
I've updated the PR accordingly. It should be good to have another look now~

t83714:
Created a separate meta JSON schema update PR (for ease of reviewing) here:
https://github.com/magda-io/magda/pull/2600

t83714:
oops~ it seems some existing test cases don't generate valid testing data... 😢 
Will look into that for a fix.

t83714:
@AlexGilleran 
All test cases passed~
It's ready to have another look~

PR #2573, opened 2019-11-04
Size tooltip

### LINKED ISSUE #2558: Add Tooltip Explaining Size Abbreviations
## Description

Currently for files that are drag-dropped onto the Add Dataset flow, we display size as KiB/MiB etc - this is more precise (see https://www.majordifferences.com/2018/03/differences-between-megabyte-and.html), but to the uninitiated looks like a typo. We should add a help tip that explains this.

## Technical Notes

There are already existing tips - they look like (!) - in the UI. We should just reuse whatever we've done for them in this task. If there's not already a reusable component for these we should create one.

## Acceptance Criteria

- Every potential instance of MiB / KiB / etc has a tooltip nearby that explains the megabyte / mebibyte difference.

#### Updated: on 1st Nov
- the tooltip should be a mouseover one
- We only need to update `add dataset page`. Don't need to worry about Admin-ui for this ticket

sajidanower23:
### What this PR does

Fixes #2558 

Adds a mouse-over tooltip explaining the difference between MB and MiB, KB and KiB, etc.

### Checklist

### Screenshot

tooltip_human_file_size


t83714:
Tested in gitlab cluster and works well 👍  

@alyce Are you happy with the tooltip text? it now reads as:

> Note: 1 KiB = 1024 Bytes, 1 MiB = 1024 KiB

Screenshot:

image

If you want to have a play, here is the test site:

https://2558-size-tooltip.dev.magda.io/dataset/add

t83714:
@sajidanower23 
Had a chat with @alyce 
> is it possible to make sure the line break is after the comma, e.g.
> Note: 1 KiB = 1024 Bytes,
> 1 MiB = 1024 KiB
> or make the tooltip a bit wider so it's all on one line?

At the moment, the `1 MiB` was cut off after 1 in tooltip

sajidanower23:
@t83714  that works fine for me in Chrome. Were you using firefox?
tooltip_size

t83714:
> @t83714 that works fine for me in both Chrome and Firefox. What version of which browser were you using?
> tooltip_size

I am using chrome on a Mac. Probably you can wrap that line with a block element to guarantee a new line?

PR #2589, opened 2019-11-11
Memory leak: Explicitly set _isMounted

### LINKED ISSUE #2583: Memory leak when adding a file to the add dataset flow
**Describe the bug**
A warning on a possible memory leak appears on the browser console when a file is added to the add dataset flow.

**To Reproduce**

1. Go to '/dataset/add'
2. Add a file
3. Go to the next page, `Details and Contents`
4. See the warning on the browser console

**Screenshots**
memory_leak

**Additional context**
The warning doesn't appear if you do not upload a file.

sajidanower23:
### What this PR does

Fixes #2583 

Explicitly set `_isMounted` so that a mounted component is marked as such.
Thanks to the comment [here](https://github.com/material-components/material-components-web-react/issues/434#issuecomment-449561024) for the suggestion.

### Checklist

### DIFF
--- a/magda-web-client/src/Components/Header/Header.js
+++ b/magda-web-client/src/Components/Header/Header.js
@@ -17,18 +17,15 @@ class Header extends Component {
     }
 
     componentDidMount() {
-        this._isMounted = true;
-        this.context.router.history.listen(() => {
-            if (this._isMounted) {
-                this.setState({
-                    isMobileMenuOpen: false
-                });
-            }
+        this.unListen = this.context.router.history.listen(() => {
+            this.setState({
+                isMobileMenuOpen: false
+            });
         });
     }
 
     componentWillUnmount() {
-        this._isMounted = false;
+        this.unListen();
     }
 
     toggleMenu() {
--- a/magda-web-client/src/Components/Header/Header.js
+++ b/magda-web-client/src/Components/Header/Header.js
@@ -19,9 +19,11 @@ class Header extends Component {
     componentDidMount() {
         this._isMounted = true;
         this.context.router.history.listen(() => {
-            this.setState({
-                isMobileMenuOpen: false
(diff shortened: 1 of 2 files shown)


magda-web-client/src/Components/Header/Header.js
```diff
@@ -17,13 +17,18 @@ class Header extends Component {
     }
 
     componentDidMount() {
+        this._isMounted = true;
```
t83714: I don't think you interact with this property later?
Sometimes, people define a local `_isMounted ` state to maintain the mount/unmount state.
But  you need to check this property before the problematic `setState` somewhere?

t83714:
I guess you tried to follow this approach but didn't fully implement.
Here are some details:
https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

magda-web-client/src/Components/Header/Header.js
sajidanower23: Good catch! Added the check

magda-web-client/src/Components/Header/Header.js
```diff
@@ -17,17 +17,26 @@ class Header extends Component {
...
         this.context.router.history.listen(() => {
-            this.setState({
-                isMobileMenuOpen: false
-            });
+            if (this._isMounted) {
```
t83714: If we are sure the `setState` here is the one who causes the problem, we probably should solve it by un-register the listener when the component unmount (`componentWillUnmount `). Otherwise, we just muted the warning (as we will not call `setState` on an unmounted component anymore) but the memory leaks (listener) are still there.
You can de-register the listener by:
```
this.unlisten = this.context.router.history.listen(...)
```

Then call `this.unlisten` when component is unmounted.

PR #2600, opened 2019-11-14
Replaced unversioned meta-schema URIs with draft version 7 (Targets PR for issue#2001)

t83714:
### What this PR does

This PR was created only for ease of review PR https://github.com/magda-io/magda/pull/2550 and targets that PR.

Should merge this before merge https://github.com/magda-io/magda/pull/2550

Summary:
- Replace all meta schema URI `http://json-schema.org/schema#` with `https://json-schema.org/draft-07/schema#`
  - The unversioned meta-schema URIs should no longer be used.
  - Https should no longer be used
  - The Json schema Validation lib we used now support up to draft 7. See [here](https://github.com/everit-org/json-schema#user-content-draft-4-draft-6-or-draft-7). 
    - Please note draft 7 is the current latest one (released on 2019-09). 
    - All other supported versions available are:
      - draft 4: `https://json-schema.org/draft-04/schema#`
      - draft 6: `https://json-schema.org/draft-06/schema#`
      - there is no draft 5 identifier (use draft 4) see [all specs](http://json-schema.org/specification-links.html#table-of-all-versions-of-everything)

### DIFF
--- a/magda-registry-aspects/access.schema.json
+++ b/magda-registry-aspects/access.schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://json-schema.org/schema#",
+    "$schema": "https://json-schema.org/draft-07/schema#",
     "title": "Dataset Access Instructions",
     "description": "Provides information that users need to access the dataset.",
     "type": "object",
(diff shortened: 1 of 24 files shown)


t83714:
Close as we're gonna auto map unsupported meta schema URI to draft 7

PR #2604, opened 2019-11-18
Unique key for props in homepage

### LINKED ISSUE #2602: React Warning: Props in homepage need unique key
**Describe the bug**
In similar vein to #2578 , the props in `/` should have unique key and one of them don't.

**To Reproduce**

1. Go to `/` of your Magda instance
1. See error in the console

**Expected behavior**
Each child in a list should have a unique key.

**Screenshots**
react_warning_homepage

sajidanower23:
### What this PR does

Fixes #2602 

Adds unique keys to `div`s that didn't have unique keys.

### Checklist

-  [x] ~There are unit tests to verify my changes are correct~ Unit tests aren't applicable
-  [x] I've updated CHANGES.md with what I changed.
-  [x] I've linked this PR to an issue in ZenHub (core dev team only)


magda-web-client/src/Components/Home/Stories.js
```diff
@@ -108,12 +108,14 @@ class Stories extends Component {
                                         case 2:
                                         case 3:
                                             return (
-                                                <div className="stories-container">
+                                                <div
+                                                    className="stories-container"
+                                                    key={r}
```
t83714: `r` seems the row number. thus appropriate to be the key (as you could have two elements with the same key)?

sajidanower23: The `div` above has the key `r`. This is another list of `div`s within that row. If the key is `r` then all of the `div`s in this row will have the same key.

t83714: I had not looked into the code but if we are sure the row number is only used on row level div then it should be OK.
Also:
image
seems still some divs have no key?
Moreover, I suspect the no `key props` warning should be generated for the mobile view ---- not the code block you modify here.

PR #2619, opened 2019-11-22
Fixed typescript error around session cookie options.

AlexGilleran:
The build is currently breaking because some typescript dependency changed, this fixes it.

### DIFF
--- a/magda-gateway/src/Authenticator.ts
+++ b/magda-gateway/src/Authenticator.ts
@@ -5,14 +5,27 @@ import * as passport from "passport";
 import * as URI from "urijs";
 import * as signature from "cookie-signature";
 
+/** This is present in the express-session types but not actually exported properly, so it needs to be copy-pasted here */
+type SessionCookieOptions = {
+    maxAge?: number;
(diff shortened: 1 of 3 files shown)


magda-gateway/src/Authenticator.ts
```diff
@@ -5,14 +5,27 @@ import * as passport from "passport";
...
 export interface AuthenticatorOptions {
     sessionSecret: string;
     dbPool: pg.Pool;
-    cookieOptions?: express.CookieOptions;
+    cookieOptions?: SessionCookieOptions;
 }
 
 export const DEFAULT_SESSION_COOKIE_NAME: string = "connect.sid";
-export let DEFAULT_SESSION_COOKIE_OPTIONS: express.CookieOptions = {
+export let DEFAULT_SESSION_COOKIE_OPTIONS = {
```
t83714: Should here be `DEFAULT_SESSION_COOKIE_OPTIONS : SessionCookieOptions` so that if we decide to change the default session cookie options, it's still covered by type checking?

PR #2639, opened 2019-11-27
Improve keywords generation logic for Spreadsheet

### LINKED ISSUE #2635: No or few keywords are derived from sparse CSVs
**Describe the bug**
Some CSV/Excel files have very little in the way of text in them, but that text is important. E.g. this SoE CSV:

https://data.gov.au/dataset/ds-dga-ebde6bad-9aef-44b0-916c-8d645ab966e5/details?q=

Looks like this:

image

"Year" is not a good keyword, but "passenger", "car" and "ownership" are all reasonable. We need to figure out how to get those keywords out when there's not a lot of text to determine what's a keyword and what isn't.

Another example is this one:
https://data.gov.au/dataset/ds-dga-40c921f5-4ee8-4299-aadb-93e9d441c591/details?q=
image

This includes "other arthropods" as a keyword for some reason, but it should really be including every row header as a keyword, given that there's so few rows.

**Expected behavior**
- "Passenger", "Car", "Ownership" are all generated from the CSV in https://data.gov.au/dataset/ds-dga-ebde6bad-9aef-44b0-916c-8d645ab966e5/details?q= when it's drag-dropped onto Add Dataset.
- Every row header is included as a keyword when the CSV from https://data.gov.au/dataset/ds-dga-40c921f5-4ee8-4299-aadb-93e9d441c591/details?q= is drag-dropped on the add dataset flow

t83714:
### What this PR does

Fixes #2635 
- Make sure headers are always captured as keywords (except contents without [a-zA-Z] or none text field)
  - If cell content too long (> 100 char or 10 words), still leave for NLP (Only when not enough keywords produced)
- If not enough keywords are produced, non header cells will be used to created keywords as well

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -142,6 +142,7 @@ UI:
 -   Added tooltips to the `Production` section of the `People and Production` page
 -   Reworded the user access options
 -   Removed help icons without content
+-   Improve keywords generation logic for Spreadsheet
 
 Gateway:
 
--- a/magda-web-client/src/Components/Dataset/MetadataExtraction/extractKeywords.ts
+++ b/magda-web-client/src/Components/Dataset/MetadataExtraction/extractKeywords.ts
@@ -19,25 +19,16 @@ export async function extractKeywords(
     },
     output: { keywords: string[] }
 ) {
-    if (input.keywords && input.keywords.length) {
-        // --- if extractText has product keywords from headers, copy from input
-        output.keywords = input.keywords;
-        // --- if no cells are skipped due to large content, no need to NLP
-        if (input.skippedCellForKeywords === false) {
-            return;
-        }
-    }
+    let keywords = [] as string[];
 
-    if (output.keywords && output.keywords.length >= MAX_KEYWORDS) {
-        return;
-    }
+    console.log(input.skippedCellForKeywords);
 
-    if (input.text) {
+    if (input.text && input.skippedCellForKeywords !== false) {
         // Only take up to a certain length - anything longer results in massive delays and the browser
         // prompting with a "Should I stop this script?" warning.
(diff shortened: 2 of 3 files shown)


AlexGilleran:
This works great on the SoE ones but it's a step backwards for others. E.g. ASIC used to give us quite nice keywords but now results in this 😬 :

image

Maybe we need to be breaking on `_` and other punctuation?

[credit_lic_201906.xlsx](https://github.com/magda-io/magda/files/3899775/credit_lic_201906.xlsx)

t83714:
> This works great on the SoE ones but it's a step backwards for others. E.g. ASIC used to give us quite nice keywords but now results in this 😬 :
> 
> image
> 
> Maybe we need to be breaking on `_` and other punctuation?
> 
> [credit_lic_201906.xlsx](https://github.com/magda-io/magda/files/3899775/credit_lic_201906.xlsx)

If we just break the `_` it will produce something like `CRED LIC STATUS`...
The previous result is created by text-based NLP --- the sample file has a `CRED_LIC_AUTHORISATIONS` columns which provide sufficient text content for analysis.

Probably skip the new logic (i.e. use header for keywords) if a big chunk of text is found from the file?

Currently, NLP only kicks in if headers can't produce sufficient (10)  keywords

AlexGilleran:
Yeah maybe flip it around so if NLP can get enough keywords use that, otherwise use the header names 👍

t83714:
Updates:
- Only generate text for NLP if large text content detected in a cell
- NLP keywords result get higher priority to be returned
- Make sure all cells are checked till a large text content cell is found
@AlexGilleran  I think it's ready to have another play~

AlexGilleran:
👌

PR #2641, opened 2019-11-29
Issue/2593 Upgrade Typescript Project to Use Project References

### LINKED ISSUE #2593: Use TypeScript Project References to make the build faster and less fragile
TypeScript introduced Project References in v3.0:
https://www.typescriptlang.org/docs/handbook/project-references.html

In 3.7, they finally became able to work with code that isn't 100% TypeScript (e.g. that uses `allowJs: true`). So it should be possible to use them in Magda now.

Theoretical benefits:
* No need to maintain dual tsconfig files (`tsconfig.json` for editors, `tsconfig-build.json` for actually compiling). 
* VSCode auto-import would import modules from the right places. No more replacing `src` with `dist` after an auto-import.
* Faster and simpler builds. Just build the microservice you're interested in, and it'll build `typescript-common`, etc. only if necessary.
* Enforced dependencies. One microservice can't accidentally depend on another, for example.

### Acceptance Criteria
Either make a change that does the following or figure out why we can't do it:
- Both `yarn run dev` and `yarn build` both still work for every typescript module
- There are no more `tsconfig-build.json` files, `tsconfig.json` is all that's needed
- All imports from `dist` are changed to `src` for typescript
- @kring to review the PR to confirm it suits Terria

t83714:
### What this PR does

Fixes #2593 

Upgraded the followings:
- `typescript`: to 3.7.2
- `ts-node`: to 8.5.2
- `tsconfig-paths`: to 3.9.0
- `webpack`: to 4.41.2 
  - for support of new version `ts-loader` which support `project references`
  - It's for connectors only (to build browser version)
  - preview-map, web-client are still on their current versions
- added `tsconfig-paths-webpack-plugin` for better `project references` support as well

Changes Summary:
- Update `tsconfig-global.json`:
  - Added new options introduced by new typescript version
  - Merged with current settings
  - "composite": true: enable compile by projects
  - Turn off "noEmitOnError": 
    - In project compile mode, new typescript version will stop emitting `by default` (i.e. when this option off) when meeting any errors. However, if set this option to `true`, ts will stop emit for dependent projects and leave you a `no emit` error.
- Code Changes:
  - replace any `@magda/typescript-common/dist` with `magda-typescript-common/src`
  - replace any `@magda/minion-framework/dist` with `magda-minion-framework/src`
  - `--esModuleInterop` is turned on by default for new version TS. so replace import commonJs moulde `import * as foo from "foo"` with import foo from foo (it will work even the module has no default export). [see here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop)
    - It doesn't mean you should never use `import * as` --- es6 module with no default export can still import all export in that way.
    - It's only about commonJS modules (esp. module export a function eg. `nock` --- it's against ECMAScript spec as a namespace import can't be a callable)
  - Other code changes due to strict type checking & bugs fixes:
    - `magda-typescript-common/src/express/status.ts`: 
      - Defines a function on one if branch is not allowed. 
      - Fix a logic problem that miss checking `nextState` is undefined 
    - `magda-typescript-common/src/tenant-api/AuthorizedTenantClient.ts`:
      - `getTenants`: make exception not all muted so it matches other code.

### Limitation on typescript & Other consideration:

Typescript compiler won't allow you to have two different `compile-time module path` & `runtime module path`, which make it difficult to deploy compiled commonJS module code.

e.g. if you put `import x from "magda-typescipt-common/src"`; in your code, the `tsc` build result will output `require("magda-typescipt-common/src")` in the built code. 

And typescript won't allow you to map it to a runtime path (e.g. `@magda/typescript-common/dist` or `magda-typescipt-common/dist` like the feature that babel, webpack & rollup all provides). I think this is not gonna be changed in short-term as Microsoft team believe mapping the runtime module path alters `import` statement's behaviour. Thus it's against typescript's [design goal 7](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).  See [here](https://github.com/microsoft/TypeScript/issues/31643#issuecomment-496995677).

[project references](https://www.typescriptlang.org/docs/handbook/project-references.html) feature won't help this either.

There are solutions like [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) available. But it's for ts-node (works well for our test cases with project references) and our runtime env in docker is quite different (can't be figured out from `tsconfig.json`).

Thus, my current solution is to use Babel's `babel-plugin-module-resolver` module to replace module alias in TS build result - The current build script is `tsc -b && babel dist -d dist`;
- I've deployed the built result to `https://issue-2593.dev.magda.io/`

@AlexGilleran @kring 
You probably won't like it.  Thus, create a draft PR to collect feedback, ideas and decide where we go 😄 

Other solutions I can think of:
- Use bundle tools to compile the build. e.g. ts-loader works well with the webpack for all connectors setups
- Write a simple bootstrap code (like `tsconfig-paths/register`) to override `require`. We will need to alter the way we run our code. e.g. `node -r ourModuleAlias.register dist/index.js`. Will need to update our helm charts.
- Re-organise `magda-typescript-common` & `magda-minion-framework`:
  - Make all module import through `magda-typescript-common/` or `magda-minion-framework/` (no `src`)
    - e.g. `import addJwtSecretFromEnvVar from "magda-typescript-common/session/addJwtSecretFromEnvVar";`
  - Build as AMD modules and output as a signle file.
  - Use project reference `prepend option` to include `magda-typescript-common` & `magda-minion-framework` in build file. e.g.  `"references": [{ "path": "../magda-typescript-common", "prepend": true }]` will include AMD module build output at beginning of the built file.
  - However, node doesn't support AMD so we still need a runtime loader.

@AlexGilleran @kring 

Please let me know your idea~

### Test Site:

https://issue-2593.dev.magda.io/

### Checklist

-   [ ] There are unit tests to verify my changes are correct


AlexGilleran:
I think I'm OK with this as long as it solves @kring 's problems 👍

t83714:
@AlexGilleran 
Re-create a new PR from this branch to remove the no. of the moving pieces required.
New PR here:
https://github.com/magda-io/magda/pull/2650

PR #2647, opened 2019-12-05
Basic storage service

### LINKED ISSUE #2608: Add Basic Storage
## Description

Magda has long had a gap in its feature-set around storing files - we've always linked back to the source of data, but never actually stored it ourselves. For most of our use cases we don't actually need it, but it's becoming increasingly important, so we should do it.

## Acceptance Criteria

- A new API exists that allows for files to be:
    - Uploaded
    - Downloaded
    - Replaced
- Only admins can upload files
- Downloads are unauthenticated for now
- Files can be stored with:
    - Google Cloud Storage
    - AWS S3
    - Locally
- In particular, it must work on minikube or equivalent without needing to communicate with some kind of cloud service
- Files are stored in a non-volatile manner... i.e. if the container restarts, files should not disappear
- The API supports upload from a web browser, as tested through something like postman or curl (the ui part doesn't need to be done yet)
- The rest of the system should be able to work with this turned off

## Technical Notes

- We'd previously discussed using minio for this, as it allows for us to program against one API that can be backed in a number of ways (GCS, S3, Local).

## Out of Scope

- Policy-based Authorisation - for now admins can add/update any file, and anyone can read them
- UI to actually upload files
- APIfication

sajidanower23:
### What this PR does

Fixes #2608 

Adds a storage service API to the magda ecosystem.

### Checklist

-   [x] There are unit tests for the service
-   [x] There is a README for the service
-   [x] Upload functionality works
-   [x] Download functionality works
-   [x] Add this service to the CI pipeline
-   [x] Make uploading a PUT request
-   [x] Handle empty objects coming through (To reproduce, send a curl request without specifying the header or data)


magda-storage-api/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```console
+$ curl -X GET localhost:6120/v0/myFavouriteCSV
+column1,column2
+A,1234
+B,4321
+C,2007
+```
+
+### POST /:recordid
```
AlexGilleran: nitpick, but it's probably better to call it `fileid` or something like that to show that it's orthogonal to a record in the registry.

magda-storage-api/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@magda/datastore-api",
+  "version": "0.0.56-terria-0",
```
AlexGilleran: Should probably take the terria out

magda-storage-api/src/MagdaMinioClient.ts
```diff
@@ -0,0 +1,123 @@
...
+                        ready: false,
+                        error: "Bucket does not exist."
+                    });
+                }
+            }
+        );
+    };
+
+    getFile(fileName: string): ObjectFromStore {
+        const self = this;
```
AlexGilleran: This is OK but if you switch to arrow functions you don't have to do it, `this` will stay as `this` :)

magda-storage-api/src/GoogleCloudStorageClient.ts
```diff
@@ -0,0 +1,54 @@
+import { Bucket, Storage } from "@google-cloud/storage";
```
AlexGilleran: If we've got minio do we still need a separate Google Cloud client?

deploy/helm/magda/charts/storage-api/templates/init-scss-compiler.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: batch/v1
```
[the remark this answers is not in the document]
sajidanower23: Removed the file :+1:

deploy/helm/magda/charts/storage-api/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+{{ toYaml .Values.resources | indent 10 }}
+        env:
+        - name: NODE_ENV
+          value: production
+        - name: JWT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: auth-secrets
+              key: jwt-secret
+{{- if .Values.global.noDbAuth }}
```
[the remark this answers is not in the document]
sajidanower23: Removed

deploy/helm/magda/charts/storage-api/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+{{- end }}
+      containers:
+      - name: storage-api
+        image: {{ template "dockerimage" . }}
+        imagePullPolicy: {{ .Values.image.pullPolicy | default .Values.global.image.pullPolicy }}
+        command: [
+            "node",
+            "/usr/src/app/component/dist/index.js",
+            "--authApiUrl", "http://authorization-api/v0",
+            "--opaUrl", "http://authorization-api/v0/opa/",
```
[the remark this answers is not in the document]
sajidanower23: The command line args were all wrong. Used the correct ones :+1:

deploy/helm/magda/charts/storage-api/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+            "--listenPort", "80",
+            "--dbHost", "content-db",
```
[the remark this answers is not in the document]
sajidanower23: Nope. Removed.

deploy/helm/magda/charts/storage-api/templates/configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "scss-compiler-config"
```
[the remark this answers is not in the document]
sajidanower23: Nope. Removed. Should I get rid of the file?

magda-storage-api/README.md
sajidanower23: I agree. Changed to `fieldId`

magda-storage-api/src/MagdaMinioClient.ts
sajidanower23: True. Switched to using arrow functions.

magda-storage-api/src/GoogleCloudStorageClient.ts
sajidanower23: I don't think so. Removed Google Cloud client.

t83714:
I could be wrong but I think this ticket is about creating a `MinIO` instance that works under:
- default mode: local storage (in the cluster) probably created as a statefulset
- use s3 storage: Run MinIO in S3 Gateway mode. The MinIO instance pass-through API calls to s3
- use GCS storage: Run MinIO in S3 Gateway mode. The MinIO instance pass-through API calls to google cloud
Probably looks like:
                                              |--> MinIO S3 gateway mode
-->MinIO S3 Like API --->|--> MinIO Default Mode (Local Storage)
                                              |--> MinIO GCS gateway mode

deploy/helm/magda/charts/storage-api/templates/configmap.yaml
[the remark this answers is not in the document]
AlexGilleran: 👍

AlexGilleran:
I think we can leave S3 for now but definitely GCE and local yeah

deploy/helm/magda/charts/storage-api/templates/configmap.yaml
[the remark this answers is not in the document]
sajidanower23: Removed

.gitlab-ci.yml
```diff
@@ -324,6 +324,47 @@ buildtest:typescript-apis:
...
+    key: $CI_JOB_NAME-$CACHE_VERSION
+    paths:
+      - "$CI_PROJECT_DIR/pip-cache"
+  services:
+    - docker:dind
+  variables:
+    PIP_CACHE_DIR: "$CI_PROJECT_DIR/pip-cache"
+    MINIO_HOST: "docker"
+    MINIO_PORT: "9000"
+    # OPA_URL: "http://docker:8181/"
```
AlexGilleran: Can we delete this line?

.vscode/launch.json
```diff
@@ -456,6 +456,30 @@
...
+            "name": "Launch Storage API",
+            "runtimeExecutable": "${workspaceRoot}/scripts/node_modules/.bin/ts-node",
+            "cwd": "${workspaceRoot}/magda-storage-api",
+            "args": ["src/index.ts"]
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "protocol": "inspector",
+            "name": "Content API Test Cases",
```
AlexGilleran:
```suggestion
            "name": "Storage API Test Cases",
```

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+# CHANGELOG
+
+## 0.0.57
```
AlexGilleran: Two ## 0.0.57s here :)

deploy/helm/magda/charts/storage-api/templates/deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+{{- end }}
+      containers:
+      - name: storage-api
+        image: {{ template "dockerimage" . }}
+        imagePullPolicy: {{ .Values.image.pullPolicy | default .Values.global.image.pullPolicy }}
+        command: [
+            "node",
+            "/usr/src/app/component/dist/index.js",
+            "--listenPort", "80",
+            "--bucket", "", # fill this in
```
AlexGilleran: Should these come through helm config / secrets?

magda-storage-api/src/MagdaMinioClient.ts
```diff
@@ -0,0 +1,124 @@
+import { Probe } from "magda-typescript-common/src/express/status";
+import ObjectFromStore from "./ObjectFromStore";
+import ObjectStoreClient from "./ObjectStoreClient";
+import { Stream, Readable } from "stream";
+
+const Minio = require("minio");
```
AlexGilleran: Looks like you can `import` this if you install https://www.npmjs.com/package/@types/minio ?

magda-storage-api/src/test/storage.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                .set("Accept", "application/json")
+                .set("Content-Type", "text/csv")
+                .send("Testing download")
+                .expect(200)
+                .then(_res => {
+                    return request(app)
+                        .get("/v0/download-test-file-1")
+                        .set("Accept", "application/json")
+                        .set("Accept", "text/plain")
+                        .expect(200)
```
AlexGilleran: This is good but it'd be good to test that downloading tests the correct content type. Probably having one test for binary and one test for text would be good too :).

magda-storage-api/yarn.lock
```diff
@@ -0,0 +1,3655 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
AlexGilleran: What happened here? There should just be one yarn.lock 🤔

sajidanower23: Removed

CHANGES.md
sajidanower23: Consolidated the two :+1:

.gitlab-ci.yml
sajidanower23: Done

deploy/helm/magda/charts/storage-api/templates/deployment.yaml
sajidanower23: I think it's done, but do confirm :)

mwu2018:
I am investigating if this service can be used in `Digtal Twin`. Here are my 2 cents.
When running test, it would be good if we can simply run command
`yarn test`. However, in the current form it does not work. I have to run
`docker-compose up` first. Then edit `launch.json` to add
```
"env": {
                "MINIO_HOST": "localhost",
                "MINIO_PORT": "9000",
                "MINIO_ACCESS_KEY": "minio",
                "MINIO_SECRET_KEY": "minio123"
}
```
to "Storage API Test Cases".

magda-storage-api/src/test/storage.spec.ts
sajidanower23: Added a test for binary files :+1:

sajidanower23:
Thanks @mwu2018 . Added env variables to launch.json :+1:

deploy/helm/magda/charts/storage-api/templates/deployment.yaml
```diff
@@ -24,13 +24,35 @@ spec:
...
             "--listenPort", "80",
-            "--bucket", "", # fill this in
-            "--minioAccessKey", "", # fill this in
-            "--minioSecretKey", "", # fill this in
             "--minioEnableSSL", "true",
-            "--minioServerHost", "", # fill this in
-            "--minioServerPort", ""  # fill this in
         ]
+        env:
+        - name: MINIO_HOST
```
AlexGilleran: Realise I'm getting really annoying now, but only the SECRET_KEY and ACCESS_KEY should really be secrets, everything else belongs as a value that comes through from `values.yml`. Just makes it a bit easier to configure, because secrets are annoying (but necessary for keeping things secret)

sajidanower23: Fair point. Added values in the `deployment.yaml` file.

deploy/helm/magda/charts/registry-api/Chart.yaml
```diff
@@ -2,3 +2,7 @@ apiVersion: v1
 description: A Helm chart for Kubernetes
 name: registry-api
 
+dependencies:
```
AlexGilleran: Shouldn't this be a dep of storage api?

deploy/helm/magda/charts/storage-api/values.yaml
```diff
@@ -0,0 +1,12 @@
+image: {}
+resources:
+  requests:
+    cpu: 10m
+    memory: 30Mi
+  limits:
+    cpu: 50m
+scssVars: '{}'
```
AlexGilleran: Don't need SCSS vars for this :)

deploy/helm/magda/charts/minio/.helmignore
```diff
@@ -0,0 +1,21 @@
+# Patterns to ignore when building packages.
```
AlexGilleran: You might have told me this before the holidays and I've forgotten, but what's the advantage of copy-pasting the chart in vs. just referring to its URL as a dependency?

deploy/helm/magda/charts/storage-api/templates/deployment.yaml
```diff
@@ -0,0 +1,58 @@
...
+              name: storage-secrets
+              key: secretkey
+        - name: MINIO_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              name: storage-secrets
+              key: accesskey
+
+{{- if .Values.global.enableLivenessProbes }}
+        livenessProbe:
```
AlexGilleran: Are you sure about these? Your `docker-compose` has it as `http://localhost:9000/minio/health/live`

magda-storage-api/src/MagdaMinioClient.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    constructor({ endPoint, port, useSSL, accessKey, secretKey }: any) {
+        this.client = new Minio.Client({
+            endPoint,
+            port,
+            useSSL,
+            accessKey,
+            secretKey
+        });
+        this.client.makeBucket(this.bucketName, "us-east-1", (err: Error) => {
```
AlexGilleran: - What's the point of making a bucket on startup now that the caller of the API specified a bucket?
- Given that the caller of the API specifies the bucket, how do they make sure it's created before they put something in it?
- Should `"us-east-1"` be configurable?

magda-web-client/src/Components/Dataset/Add/DatasetAddFilesPage.tsx
```diff
@@ -90,6 +90,7 @@ class DatasetAddFilesPage extends React.Component<
                 // --- skip the directory item
                 continue;
             }
+            if (!thisFile) continue;
```
AlexGilleran: What's this for? Shouldn't it be unreachable because of the if statement above?

sajidanower23: Yeah good catch. It was a remnant from a previous PR and a result of merge somewhere down the line, I think :man_shrugging:

deploy/helm/magda/charts/storage-api/values.yaml
sajidanower23: Removed

deploy/helm/magda/charts/registry-api/Chart.yaml
sajidanower23: It should be :man_facepalming:

deploy/helm/magda/charts/minio/.helmignore
sajidanower23: Minio here is not just a dependency, it's also a magda service in a way (since storage needs a minio server running for it to work). Is it possible to achieve that just by referring to the chart via a URL?

magda-storage-api/src/MagdaMinioClient.ts
sajidanower23: > What's the point of making a bucket on startup now that the caller of the API specified a bucket?

What we discussed was that as a proof of concept, we will start with a generic bucket where everything is stored, and then move on to a bucket for each service. This is the generic bucket in question. The caller of the API needs to use the generic bucket's name for the moment in their request until the storage service starts keeping track of who owns which bucket.

> Given that the caller of the API specifies the bucket, how do they make sure it's created before they put something in it?

As things are, they would get a relevant error.

> Should "us-east-1" be configurable?
Yes. I just made it so :+1:

deploy/helm/magda/charts/minio/.helmignore
AlexGilleran: It _should_ be but we haven't tried it in Magda yet - ideally Helm should work a bit like NPM  but for K8S - you refer to a chart which can refer to other charts that it needs, and so on.

magda-storage-api/src/MagdaMinioClient.ts
AlexGilleran: OK 👍

PR #2659, opened 2019-12-15
Reflect move from terriajs to magda-io in docs and code

sajidanower23:
Minor tweaks, mostly `s/TerriaJS\/magda/magda-io\/magda/g`.


magda-migrator-content-db/sql/V1_12__update_linked_data_rating.sql
```diff
@@ -1,3 +1,3 @@
 DELETE FROM "public"."content" WHERE "id" = 'page/dataset-quality';
-INSERT INTO "public"."content" ("id", "type", "content") VALUES
```
AlexGilleran: This isn't actually how flyway works, sadly - if you change an existing migration, it changes the hash and the next time the migrator jobs run they'll end up failing until you run `flyway repair`. Probably better for this one if it was a new migration that changes the contents of the page only if they match the default text set here?

sajidanower23: That was an auto-formatting fix. Reverted the file.

PR #2674, opened 2019-12-20
Fixed: if featureFlags are not set, edit buttons are always shown on Dataset Page

### LINKED ISSUE #2671: User shouldn't see edit buttons on view dataset page unless login as admin
**Describe the bug**
User shouldn't see edit buttons on view dataset page unless login as admin.

see screehshot below, without login the button is shown up:

image

t83714:
…dataset page

### What this PR does

Fixes #2671 
Cause:
- Default value for `featureFlags` in helm chart is `{}`
- If `featureFlags` is not set when deployed, `config.featureFlags.cataloguing` will be `undefined`
- `hasEditPermissions` property (`DatasetPage`) is calculated as:
```js
const hasEditPermissions =
        config.featureFlags.cataloguing &&
        state.userManagement &&
        state.userManagement.user &&
        state.userManagement.user.isAdmin;
```

Its value will **always** be `undefined` rather than `false`

- The way of `ToggleEditor` figure out whether the edit mode is enable or not was:
```js
enabled = enabled === undefined || enabled;
```
This make it `true` when `enabled` = `undefined`

### Checklist


t83714:
Will create another one against release/0.0.56

PR #2676, opened 2019-12-20
Fixed: if featureFlags are not set, edit buttons are always shown on dataset page

### LINKED ISSUE #2671: User shouldn't see edit buttons on view dataset page unless login as admin

t83714:
### What this PR does

Fixes #2671 
Cause:
- Default value for `featureFlags` in helm chart is `{}`
- If `featureFlags` is not set when deployed, `config.featureFlags.cataloguing` will be `undefined`
- `hasEditPermissions` property (`DatasetPage`) is calculated as:
```js
const hasEditPermissions =
        config.featureFlags.cataloguing &&
        state.userManagement &&
        state.userManagement.user &&
        state.userManagement.user.isAdmin;
```

Its value will **always** be `undefined` rather than `false`

- The way of `ToggleEditor` figure out whether the edit mode is enable or not was:
```js
enabled = enabled === undefined || enabled;
```
This make it `true` when `enabled` = `undefined`

### Checklist

### DIFF
--- a/magda-web-client/src/Components/Editing/ToggleEditor.tsx
+++ b/magda-web-client/src/Components/Editing/ToggleEditor.tsx
@@ -6,12 +6,12 @@ interface ToggleEditorProps<V> {
     value: any;
     onChange: Function;
     editor: Editor<V>;
-    enabled?: boolean;
+    editable: boolean;
 }
 
 /**
  * Will toggle between editing and viewing with an edit button.
- * Will always show if enabled is false.
+ * Will always show if editable is false.
  * Can specify custom viewer by specifying children which will be rendered instead of viewer of the editor.
  * Interchangable with AlwaysEditor.
  */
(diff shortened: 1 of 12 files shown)


magda-web-client/src/Components/Editing/ToggleEditor.tsx
```diff
@@ -56,7 +56,7 @@ export class ToggleEditor<V> extends React.Component<ToggleEditorProps<V>> {
             value = this.props.value;
         }
 
-        enabled = enabled === undefined || enabled;
+        enabled = enabled ? true : false;
```
AlexGilleran: Tbh I think there's a few fundamental problems with how this was done originally.

1. That "enabled" isn't a great prop name (the field is "enabled" no matter what, what enabled actually does is toggle whether it can be edited.
2. That it allows for the prop to not be set.

How do you feel about renaming `enabled` to `editable` and making it mandatory (i.e. it always has to be set to `true` or `false`) so that this mistake is harder to make?

t83714: Yeah~ I probably can rename it and make it throw an exception if its value is not a boolean~

t83714:
@AlexGilleran  
I have renamed the prop name and it's ready to have another look~

Issue #2667, opened 2019-12-19
Minion won't set webhook to active when restart


t83714:
**Describe the bug**
Minion won't set webhook to active when restart.
Registry may from time to time mark minion as inactive in DB (e.g. registry can't connect to minion)
When minion restart, minion should reset this mark, otherwise, the webhook will never resume.

Issue #2660, opened 2019-12-16
Registry API Web Hooks Should NOT notify webhook if webhook aspects config doesn't match


t83714:
**Describe the bug**
Registry API Web Hooks Should NOT notify webhook if webhook aspects config doesn't match.

Currently, the registry will notify webhook with empty records list if webhook interested aspects doesn't match:
```
payload:  { action: 'records.changed',
  deferredResponseUrl: 'http://localhost:6101/v0/hooks/minion-ckan-sync/ack',
  lastEventId: 10228,
  records: [] }
payload:  { action: 'records.changed',
  deferredResponseUrl: 'http://localhost:6101/v0/hooks/minion-ckan-sync/ack',
  lastEventId: 10268,
  records: [] }
payload:  { action: 'records.changed',
  deferredResponseUrl: 'http://localhost:6101/v0/hooks/minion-ckan-sync/ack',
  lastEventId: 10308,
  records: [] }
payload:  { action: 'records.changed',
  deferredResponseUrl: 'http://localhost:6101/v0/hooks/minion-ckan-sync/ack',
  lastEventId: 10348,
  records: [] }
```

This doesn't break any function (as [] will be ignored by minion-framework). But it does create a log unnecessary requests between registry and minion.
  
**To Reproduce**

- Modify the following lines of `format` minion index.ts to:
```ts
function sleuthLayerer() {
    return minion({
        argv,
        id: ID,
        aspects: ["currency"], //--- make it watch a uncommon aspect
        optionalAspects: [],
        async: false,
        writeAspectDefs: [formatAspectDef],
        onRecordFound
    });
}
```
- add `console.log(payload);` to this line: https://github.com/magda-io/magda/blob/01381eea959c8893309ce14b70ad65a78a47bd35/magda-minion-framework/src/setupWebhookEndpoint.ts#L21
- re-build format minion
- Run format minion (reset lastEventId to 1)
- Watch the format minion log. You can see it keeps receving requests (with empty payload) from registry. e.g.
```
{ action: 'records.changed',
  deferredResponseUrl: 'http://localhost:6101/v0/hooks/minion-ckan-sync/ack',
  lastEventId: 10348,
  records: [] }
```

**Expected behavior**
The registry should not notify minion for irrelevant events

PR #2677, opened 2019-12-23
Issue/2625 Publish to Ckan (prototype) 

### LINKED ISSUE #2625: Spike: Publishing to CKAN
## Description

We need to figure out the first step for getting stuff from a Magda instance onto data.gov.au. The easiest way to make that happen is going to be directly publishing to the data.gov.au CKAN instance and then letting the data.gov.au Magda pick up the dataset.

This is a spike to figure out how that should happen.

## Acceptance Criteria

-   A design has been written on Github for how this should happen for everyone to discuss
-   There's some kind of working PoC for pushing stuff to CKAN and seeing it there

## Technical Notes

- This will involve using the CKAN API to directly push metadata and files to CKAN
- Most likely this will involve using a minion

We need to consider:
- Authorising to CKAN
- How to make sure everything happens reliably

t83714:
### What this PR does

Fixes #2625

- This branch implements a prototype to demonstrate a potential Magda-Ckan publishing solution
- Although it comes with some frontend code, the solution is meant to provide backend sync capability without enforcing any business workflow. 

### How it works:

To publish a dataset to ckan, the solution implements a minion [magda-minion-ckan](https://github.com/magda-io/magda/tree/issue/2625/magda-minion-ckan-sync) to create, update & delete ckan `dataset` (`package`) through Ckan API depends on the datset's sync settings.

A Magda dataset's sync setting is stored in new aspect `ckan-sync-aspect`:

https://github.com/magda-io/magda/blob/22e716055a0fb94c500a43f95913a6c8727156fa/magda-registry-aspects/ckan-sync.schema.json#L1-L53

The minion only monitors this aspect. i.e. dataset changes don't have to always trigger a sync action. We can control the data sync timing (in order to implement different workflows ) by updating aspect `ckan-sync-aspect` when it's appropriate.

This is done mainly through two fields of `ckan-sync-aspect`:
- status: `retain` or `withdraw`
  - `retain`: indicate the minion should make sure a ckan dataset is created& sync with magda dataset
  - `withdraw`: indicate the minion should make sure if a ckan dataset is created. It should be removed from ckan system
  - If this field doesn't exist or other value, minion should assume nothing needs to be done
  - Set this field will not immediately trigger minion's sync action. Minion will only take action when `syncRequired` is true and set `syncRequired` back to false once the sync is successfully complete. 
- `syncRequired`: true or false.
  - Minion will try to perform the sync action (create / update / delete) when this field is true
  - Minion will set this field back to `false` when a sync is complete sucessfully  

#### Ckan API key 

The minion should determine the API key to use by the following logic (not fully implement one this branch):
- If `ckan-sync-aspect` / `syncUserId` is a valid magda user id, try to locate the API key store for that 
user in magda system
- If doesn't exist, use the default API key configured as part of minion config

### How to run the demo

Modify `minikube-dev.yml` to add:
```
tags:
  all: true
  minion-ckan-sync: true
```
deploy magda using `minikube-dev.yml`

#### Possible issues:

You might be impacted by the following two registry issues:

https://github.com/magda-io/magda/issues/2660
- If you have a larger dataset, you might have to wait for 30 mins + before minion receives first `ckan-sync-aspect` related notification.  Although minion framework won't call `onRecordFound`, but all event records are passing through minion silently (via http requests). To fix that, manually move the lastEventId to jump all irrelevant events.

https://github.com/magda-io/magda/issues/2667
- make sure sync minion is active in the dataset (registry may turn it off for various reasons e.g. you temporarily turn the minion off).  Restart the minion won't turn this field back on.

### Checklist


AlexGilleran:
Looks good - closing this PR but we'll use the branch later :)

PR #2683, opened 2020-01-06
Date filter

### LINKED ISSUE #2672: Date filter only lets you select a date relevant to the data available
**Describe the bug**
When selecting your date range, it is only possible to select a range based on the data available.  It is less confusing to allow the user to select their own chosen date range.

**To Reproduce**

1. Go to 'any date'
2. Click on 'the years from and to'
3. Enter in dates up to the current date

**Expected behavior**
A user should be able to enter a date range of their choice, regardless of the data available.

**Screenshots**
image

sajidanower23:
### What this PR does

Fixes #2672 

Only check for validity while setting temporal filters in the search dataset page.

### Checklist


magda-web-client/src/Components/Dataset/Search/Facets/FacetTemporal.js
```diff
@@ -156,25 +156,19 @@ class FacetTemporal extends Component {
...
-
-        const yearUpper = temporalRangeEnd.getUTCFullYear();
-        const monthUpper = temporalRangeEnd.getUTCMonth();
-
+        if (
+            this.state.startYear > this.state.endYear ||
+            (this.state.startYear === this.state.endYear &&
+                this.state.startMonth > this.state.endMonth)
+        ) {
+            alert("End date is earlier than start date.");
```
AlexGilleran: As discussed, it'd be good if this could be presented the same way as the `"Please enter a valid year."` message above :).

sajidanower23: The prompt style is similar to the "Please enter a valid year" now.
prompt
Although, I noticed that the years/months don't get updated when typing is stopped :thinking: 
For example, if I finished typing 1987 (previous year was 2016), the end year is still 2016. But when I change 1987 to something else, the end year becomes 1987, so it's always late to register, if that makes sense :man_shrugging: 
But clicking the 'Apply' button does get the latest year. It just means that the prompt is late to come and late to go.

AlexGilleran: I think the delay is ok (there's a slight debounce) but the error messages look different:

image

sajidanower23: Made it looks the same
painful_css

AlexGilleran: Tested just now and I'm able to go from an later year to an earlier year without that coming up:

image

or https://2672-date-filter.dev.magda.io/search?dateFrom=3000-01-31T13%3A00%3A00.000Z&dateTo=2995-09-30T14%3A00%3A00.000Z&q=

sajidanower23: > Although, I noticed that the years/months don't get updated when typing is stopped thinking
For example, if I finished typing 1987 (previous year was 2016), the end year is still 2016. But when I change 1987 to something else, the end year becomes 1987, so it's always late to register, if that makes sense man_shrugging
But clicking the 'Apply' button does get the latest year. It just means that the prompt is late to come and late to go.

So what I meant here is not a time delay, but an actual delay in reading of the year (meaning, the props get the second-latest value of the years and months, not the very latest one). So when you stopped typing 1400, it didn't register 1400, it registered the year before 1400. If you change it again, 1400 will register and you should see the error. I am not sure what to do about this though :thinking:

magda-web-client/src/Components/Dataset/Search/Facets/FacetTemporal.js
```diff
@@ -131,50 +136,90 @@ class FacetTemporal extends Component {
...
+            this.setPrompt("End date is earlier than start date.");
+        }
         this.setState({
             startMonth
         });
     }
 
     selectEndMonth(endMonth) {
+        const startBeforeEnd = this.checkStartAndEndDate();
+        if (!startBeforeEnd) {
```
AlexGilleran: Why not just do this bit in `checkStartAndEndDate`?

sajidanower23: Because it might be good to keep this a purely "is the start date-end date combo valid?" function. I suppose we can add a wrapper to this function (to add the warning) to deal with the annoying code duplication.

magda-web-client/src/Components/Dataset/Search/Facets/FacetTemporal.js
AlexGilleran: 😂 oh right. Hmmm...

magda-web-client/src/Components/Dataset/Search/Facets/FacetTemporal.js
sajidanower23: Done in 950b1d5

AlexGilleran:
@sajidanower23 ok so I ended up starting with that problem where the error didn't come up in sync with the user's actions and kept shaving yaks until I'd rewritten most of the date picker, as well as refactored a bunch of SCSS :grimacing: .

So date picker changes:
- Date picker logic redone with react hooks
- Date picker now works a bit different - rather than being prepopulated with years, it shows the temporal range of the dataset as placeholders but you have to put in a year
    - Which enables this: when you click outside the filter having not  changed anything, it doesn't apply the filter and reload the result, it just does nothing
- Any year moment thinks is valid is valid (so 200 is valid, 20000000 is not valid)
- The date filter is disabled if there's no dates to filter by, instead of disappearing
- Fiddled with the outlines to make it a bit easier to see what you're doing when tabbing through... which lead to:

SCSS Changes:
- Discovered when playing with outlines that the way we've been importing the design systems scss has been wrong the whole time. Essentially we import the scss from `pancake.scss`, override the colours as expected, then _reimport_ a bunch of it when we import design systems components as react components, except these imported ones don't have the colours overridden properly, which has meant that we've had a bunch of awful global styling that breaks both our stuff and breaks some of the built-in design systems styling.
- To fix this, I've made it so that we don't import _any_ react components from the design systems, everything is done using good old markup and classes. This should result in more predictable css, and a smaller bundle (no more css being imported repeatedly)
- As a result of this, I deleted a bunch of global styles from index.scss

Can you please give what I did a review and a test? I've done a full deploy onto https://2672-date-filter.dev.magda.io/search :pray: .

PR #2696, opened 2020-01-14
Added an opt-in for previews when files are large

### LINKED ISSUE #2687: Error loading CSV files smaller than 2MB
**Describe the bug**
Requesting a range for bytes on a file smaller than that range causes an HTTP 416 error:
```
$ curl 'https://data.gov.au/preview-map/proxy/_0d/https://data.gov.au/data/dataset/a03ae98f-da63-4669-9abf-fef9947b666a/resource/467fcec1-fd07-407a-ae19-790262d2d6c7/download/yarra-dumped-waste.csv'  -H 'range: bytes=0-2097151'
<html>
  <head><title>Request Range Not Satisfiable</title></head>
  <body>
    <h1>Request Range Not Satisfiable</h1>
    <p>The Range requested is not available.
<br/>Range request was made beyond the end of the content,
which is 1276263 long.
  Range: bytes=0-2097151
<!--  --></p>
    <hr noshade>
    <div align="right">WSGI Server</div>
  </body>
</html>
```

**To Reproduce**

1. Go to https://data.gov.au/dataset/ds-dga-a03ae98f-da63-4669-9abf-fef9947b666a/distribution/dist-dga-467fcec1-fd07-407a-ae19-790262d2d6c7/details

**Expected behavior**
CSV file is loaded in the table

**Actual behaviour**
Console error:
```
Failed to process CSV data file:  Error
    at u._chunkError (blob:https://data.gov.au/17d6eba8-eb8a-49ad-8519-770327b554ed:1)
    at u._chunkLoaded (blob:https://data.gov.au/17d6eba8-eb8a-49ad-8519-770327b554ed:1)
    at u._nextChunk (blob:https://data.gov.au/17d6eba8-eb8a-49ad-8519-770327b554ed:1)
    at u.stream (blob:https://data.gov.au/17d6eba8-eb8a-49ad-8519-770327b554ed:1)
    at Object.parse (blob:https://data.gov.au/17d6eba8-eb8a-49ad-8519-770327b554ed:1)
    at t.onmessage (blob:https://data.gov.au/17d6eba8-eb8a-49ad-8519-770327b554ed:1)
```

**Additional context**
https://github.com/magda-io/magda/pull/2489
The 2MB range seems to come from https://github.com/magda-io/magda/blob/d43e9783e22afa5e707147715070af569b96476a/magda-web-client/src/config.ts#L253

AlexGilleran:
### What this PR does

Fixes #2687 (sort-of)

So a while ago Jacky added some clever CSV parsing logic that gets a CSV a chunk at a time and parses it on the fly. Unfortunately:
- The way that PapaParse implements this is that if the chunk size is 1mb and the file is 2.5mb, it'll request 0-1mb, 1-2mb, 2-3mb... but the last request is out of range, and while some servers will respond with whatever they can, others will return `416`.
- If you ask the version of CKAN that data.gov.au/data runs for a partial range, it'll return a response that says it's returning a partial file in the headers, but is in fact the full file

Potentially you could modify Papaparse to work around the first bullet point, but the second is pretty hard to crack - Jacky actually tried before by reading the total length of the response, but that's since been reverted because it didn't quite work.

So with this PR I've essentially waved the white flag and compromised. The _original_ problem was that the dataset view page would lock up when trying to parse big CSV files, and this solves that by:
- Retaining the work Jacky did to make it do CSV _parsing_ in a webworker
- Rather than having the webworker fetch the CSV file as chunks, it fetches the whole file, but before fetching it tries to determine the size of the file with a `HEAD` request. If the file is too big (over a threshold specified in config), or it doesn't get a good enough HEAD response to figure that out, it displays a warning to the user before previewing, and allows them to override and preview anyway - kind of like what a browser does if it thinks there's a security risk.

### Checklist


magda-web-client/src/Components/Common/DataPreviewMap.tsx
```diff
@@ -0,0 +1,325 @@
+import React, { Component, useEffect, useState } from "react";
```
AlexGilleran: Sorry about the full-page diff here - the key changes are that introduced `singleFile` to the array of data sources - if this is set to true, the file will be checked for length before it's allowed to be previewed. If it's false, it's assumed that the dataset is a service that can't be reasonably checked for data volume, and it's just allowed to be previewed.

PR #2697, opened 2020-01-14
Storage API docs

### LINKED ISSUE #2692: API Documentation for Storage API
## Description

The storage api is currently missing from the [API Docs](https://dev.magda.io/api/v0/apidocs/) - documentation for each API endpoint needs to be added, just as it is for the other services.

## Acceptance Criteria

-   All api calls in the storage api are documented

## Technical Notes
You can see other examples of how this is done in the other node.js apis.

Correspondence-api's createApiRouter should have an example.

sajidanower23:
### What this PR does

Fixes #2692 

Adds documentation for the Storage API.

### Checklist

### DIFF
--- a/magda-storage-api/src/createApiRouter.ts
+++ b/magda-storage-api/src/createApiRouter.ts
@@ -41,7 +41,8 @@ export default function createApiRouter(options: ApiRouterOptions) {
      *
      * @apiError {string} status NOT FOUND
      *
-     * @apiErrorExample {string} No such object with fileId {fileid} in bucket {bucket}
+     * @apiErrorExample {string} 404
+     *      No such object with fileId {fileid} in bucket {bucket}
      */
     router.get("/:bucket/:fileid", async function(req, res) {
         const fileId = req.params.fileid;
--- a/magda-storage-api/src/createApiRouter.ts
+++ b/magda-storage-api/src/createApiRouter.ts
@@ -27,7 +27,11 @@ export default function createApiRouter(options: ApiRouterOptions) {
     });
 
     /**
-     * @apiGroup Storage API
+     * @apiDefine Storage Storage API
+     */
+
+    /**
+     * @apiGroup Storage
      *
      * @api {get} /v0/{bucket}/{fieldid} Request to download an object in {bucket} with name {fieldid}
      *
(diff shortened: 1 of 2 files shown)


magda-storage-api/src/createApiRouter.ts
```diff
@@ -26,7 +26,27 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+     * @apiGroup Storage API
+     *
+     * @api {get} /v0/{bucket}/{fieldid} Request to download an object in {bucket} with name {fieldid}
+     *
+     * @apiDescription Downloads an object
+     *
+     * @apiParam (Request body) {string} bucket The name of the bucket under which the requested object is
+     * @apiParam (Request body) {string} fieldid The name of the object being requested
+     *
+     * @apiSuccess {string} status OK
```
AlexGilleran: This comes out like

image

Which implies that `status` is a `string` returned in the body of the response.

I think what you want to do is something more like what authorization-api does - leave out the `@apiSuccess` bit and just keep the `@apiSuccessExample`.

magda-storage-api/src/createApiRouter.ts
```diff
@@ -98,7 +134,29 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+     *
+     * @apiParam (Request body) {string} bucket The name of the bucket where the object resides
+     * @apiParam (Request body) {string} fieldid The name of the object to be deleted
+     *
+     * @apiSuccessExample {json} 200
+     *    {
+     *        "message":"File deleted successfully",
+     *    }
+     *
+     * @apiError {string} Internal Server Error
```
AlexGilleran: You'll need to take out these too, they result in weirdness like this:

image

magda-storage-api/src/createApiRouter.ts
```diff
@@ -26,7 +26,23 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+     *
+     * @apiParam (Request body) {string} bucket The name of the bucket under which the requested object is
+     * @apiParam (Request body) {string} fieldid The name of the object being requested
+     *
+     * @apiSuccessExample {text} 200
+     *    Contents of a file
+     *
+     * @apiError {string} status NOT FOUND
+     *
+     * @apiErrorExample {string} No such object with fileId {fileid} in bucket {bucket}
```
AlexGilleran: This is breaking your build.
```suggestion
     * @apiErrorExample {string} 404
     *      No such object with fileId {fileid} in bucket {bucket}
```

You can test without pushing by running `yarn run build` in `magda-apidocs` and running an http server in `magda-apidocs/build` :)

PR #2700, opened 2020-01-14
Add mustBeAdmin hook to PUT endpoint

### LINKED ISSUE #2694: Storage API: Restrict file upload to admins
## Description

Currently any user can upload a file to Magda - this will probably require a more sophisticated authorisation system in the future, but we should start by making so that only an admin can upload a file.

## Acceptance Criteria

- A valid `PUT` API call to upload a file to the storage api, when logged in as an admin, continues to work
- A valid `PUT` API call as a non-admin (logged out or logged in as non-admin) returns [403](https://http.cat/403)
- `GET` continues to work for as before.

## Technical Notes
There's a middleware in typescript common called `mustBeAdmin` that can be used to protect an express route.

e.g.
```typescript
router.get(mustBeAdmin, '/blah', //etc
```

sajidanower23:
### What this PR does

Fixes #2694 

Restricts `PUT`ing of file to admins of the Magda instance only.

### Checklist


deploy/helm/magda/charts/storage-api/templates/deployment.yaml
```diff
@@ -39,6 +39,11 @@ spec:
             secretKeyRef:
               name: storage-secrets
               key: accesskey
+        - name: JWT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: storage-secrets
+              key: jwt-secret
```
AlexGilleran: The JWT secret is already in `auth-secrets`, can we just use that one?

sajidanower23: Yep, sure can.

sajidanower23: Yep

PR #2703, opened 2020-01-16
Update org branding

sajidanower23:
### What this PR does

Update organisation branding from CSIRO Data61 to CSIRO's Data61.


AlexGilleran:
CSIRO|Data61 5eva.

PR #2706, opened 2020-01-21
Endpoint to create buckets

### LINKED ISSUE #2693: Storage API: Mechanism to create buckets
## Description

Currently the storage api requires PUTs and GETs to specify a bucket id, but there's no way to make a bucket. We should add one!

## Acceptance Criteria

-   An API call can be made to the storage api to create a new bucket.
-   If the bucket already exists, then return `200`
-   Only admin accounts can create buckets

## Out of scope

-   Each service should have their own bucket
-   There should be different permission policies that you can set on buckets

sajidanower23:
### What this PR does

Fixes #2693 

Adds an endpoint to create buckets

### Checklist


magda-storage-api/src/MagdaMinioClient.ts
```diff
@@ -17,26 +25,41 @@ export default class MagdaMinioClient implements ObjectStoreClient {
...
+                        err.message ===
+                        "Your previous request to create the named bucket succeeded and you already own it."
+                    ) {
+                        return resolve({
+                            message: "Bucket " + bucket + " already exists 👍",
+                            success: false
+                        });
+                    } else {
+                        console.log("😢 Error creating bucket: ", err);
+                        return resolve({
```
AlexGilleran: Why not just reject in the case of an error? It seems like the `message` being passed back here doesn't actually get used by the API anyway?

magda-storage-api/src/MagdaMinioClient.ts
```diff
@@ -17,26 +25,41 @@ export default class MagdaMinioClient implements ObjectStoreClient {
...
-                    return console.log(
-                        "Bucket " + this.bucketName + " already exists 👍"
-                    );
-                } else {
-                    return console.log("😢 Error creating bucket: ", err);
+        this.region = region;
+        this.createBucket(this.bucketName);
+    }
+
+    createBucket(bucket: string): Promise<any> {
```
AlexGilleran: Looks like it returns a consistent object shape, could we make that the return type of the `Promise`? :)

magda-storage-api/src/createApiRouter.ts
```diff
@@ -29,6 +29,66 @@ export default function createApiRouter(options: ApiRouterOptions) {
         return res.status(200).send("OK");
     });
 
+    /**
+     * @apiGroup Storage
+     *
+     * @api {POST} /v0/create/bucket?bucket={bucket} Request to create a new bucket
```
AlexGilleran: Could we do `PUT /v0/create/bucket/{bucketId}` ? Generally `PUT` is more usual if you're creating something and supplying the ID, and repeats of the same operation won't have an effect. Also while I think query params in `POST` requests is technically OK, it's a bit unusual 😬 .

magda-storage-api/src/createApiRouter.ts
```diff
@@ -29,6 +29,66 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+     * @apiParam (Request query) {string} bucket The name of the bucket to be created
+     *
+     * @apiSuccessExample {json} 200
+     *    {
+     *        "message":"Bucket my-bucket created successfully in unspecified-region 🎉"
+     *    }
+     *
+     * @apiErrorExample {json} 400
+     *    {
+     *        "message": "Bucket my-bucket already exists 👍"
```
AlexGilleran: I'm not sure this should be a `400`, given that it's not really a bad request per se - I think it's pretty reasonable to do this multiple times (e.g. on startup) just to make sure a bucket is there right? If this returns a 400 then when this is called, clients will have to have special error handling logic in order to cope with the bucket already existing, which is usually OK.

magda-storage-api/src/test/storage.spec.ts
```diff
@@ -77,6 +77,49 @@ describe("Storage API tests", () => {
...
+                        bucket: dummyBucket
+                    })
+                    .expect(200, {
+                        message:
+                            "Bucket " +
+                            dummyBucket +
+                            " created successfully in unspecified-region 🎉"
+                    })
+            );
+        });
```
AlexGilleran: Should probably have a test for creating a bucket that already exists

magda-storage-api/src/createApiRouter.ts
AlexGilleran: There should probably be a `400` response if you don't provide an id at all though!

magda-storage-api/src/MagdaMinioClient.ts
sajidanower23: Rejected the Promise :+1: 
The message is not being used at the moment, but it's possible that it might be useful later. At the moment, it helps make the `CreateBucketResponse` object consistent.

magda-storage-api/src/MagdaMinioClient.ts
sajidanower23: Yep, created the type `CreateBucketResponse` to reflect that.

magda-storage-api/src/createApiRouter.ts
sajidanower23: Fair enough. Consistency is key :+1:

magda-storage-api/src/createApiRouter.ts
sajidanower23: No ID at all will now result in hitting an invalid endpoint. Trying to create a bucket that exists will return 201 still but with a different message.

magda-storage-api/src/test/storage.spec.ts
sajidanower23: Added a test for that.

sajidanower23:
@AlexGilleran @t83714 
Deployed to https://2693-create-buckets-endpoint.dev.magda.io. To test, make yourself an admin and try creating a bucket.
```
curl -X PUT --cookie 'connect.sid=<cookie>' https://2693-create-buckets-endpoint.dev.magda.io/api/v0/storage/new-bucket-name
```

magda-storage-api/src/MagdaMinioClient.ts
```diff
@@ -17,26 +26,41 @@ export default class MagdaMinioClient implements ObjectStoreClient {
...
+                        err.message ===
+                        "Your previous request to create the named bucket succeeded and you already own it."
+                    ) {
+                        return resolve({
+                            message: "Bucket " + bucket + " already exists 👍",
+                            success: false
+                        });
+                    } else {
+                        console.log("😢 Error creating bucket: ", err);
+                        return reject({
```
AlexGilleran: I don't think it's technically incorrect, but it's a lot more usual to [reject with an error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) if that's ok :)
```suggestion
                        return reject(err);
```

magda-storage-api/src/MagdaMinioClient.ts
```diff
@@ -17,26 +26,41 @@ export default class MagdaMinioClient implements ObjectStoreClient {
...
+                    if (
+                        err.message ===
+                        "Your previous request to create the named bucket succeeded and you already own it."
+                    ) {
+                        return resolve({
+                            message: "Bucket " + bucket + " already exists 👍",
+                            success: false
+                        });
+                    } else {
+                        console.log("😢 Error creating bucket: ", err);
```
AlexGilleran: This probably deserves to be a `console.error` :)
```suggestion
                        console.error("😢 Error creating bucket: ", err);
```

magda-storage-api/src/createApiRouter.ts
```diff
@@ -29,6 +29,64 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+                    return res.status(500).send({
+                        message:
+                            "Bucket creation failed. This has been logged and we are looking into this."
+                    });
+                } else {
+                    // If bucket already exists
+                    return res.status(201).send({
+                        message: createBucketRes.message
+                    });
+                }
```
AlexGilleran: Because you're rejecting from`createBucket` now, you'll need to put the call in a `try catch` and move the 500 status to be part of the `catch` clause.

Not throwing an error for the bucket already existing is good though, because exceptions should be exceptional :)

sajidanower23: Good `catch`! Fixed now

PR #2728, opened 2020-01-30
Make underlength JWTs more obvious and less likely

AlexGilleran:
### What this PR does

Dev just had some problems because our old dev JWT secret was too short to be an effective HS256 secret, and the new Java JWT library fails if this is the case. It throws an exception informing you of this but this wasn't being logged properly.

This change makes it properly log any error encountered during JWT parsing, and also makes the code around that a bit simpler (I realised I was using a deprecated method before).

### Checklist

-   [x] No CHANGES, this fixes up my last change

### DIFF
--- a/scripts/create-secrets/k8sExecution.js
+++ b/scripts/create-secrets/k8sExecution.js
@@ -170,7 +170,7 @@ function doK8sExecution(config, shouldNotAsk = false) {
         }
 
         createSecret(env, namespace, "auth-secrets", {
-            "jwt-secret": pwgen(512),
+            "jwt-secret": pwgen(64),
             "session-secret": pwgen()
         });
     });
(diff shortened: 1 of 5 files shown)


scripts/create-secrets/k8sExecution.js
```diff
@@ -170,7 +170,7 @@ function doK8sExecution(config, shouldNotAsk = false) {
         }
 
         createSecret(env, namespace, "auth-secrets", {
-            "jwt-secret": pwgen(),
+            "jwt-secret": pwgen(512),
```
t83714: This seems produces a 512 chars key (4096 bits) which is not recommended as it's greater than block size.
if we want 512 bits, it should be `64` chars.
https://security.stackexchange.com/questions/95972/what-are-requirements-for-hmac-secret-key

AlexGilleran: Oh yeah :grimacing:

PR #2735, opened 2020-02-14
Add auth for GET endpoint in storage

### LINKED ISSUE #2609: Add GET Authorisation to File Storage
## Description
Integrating the policy-based authz used elsewhere in Magda with the new file storage system for GET access.

## Acceptance Criteria
- A file in storage api can have a record id / aspect id combination stored against it
- When a user makes a GET for a file, the storage API looks up whether the user is allowed to access that aspect of that record in the registry. If so it lets it through, if not it doesn't.

## Technical Notes
- @kring already kinda did this over at https://github.com/terriajs/magda
- This might be possible using minio metadata

### TODO
- [ ] When uploading, insert recordId to the magda-registry
- [ ] When GET-ing, make a request to registry with the recordId and userId
- [ ] If recordId is undefined, send the file through
- [ ] If recordId is defined, check if user can access the registry, and send the file through accordingly

## Out of Scope
- Authentication without the registry... we'll have to think about how that would work / what situation that would work in.
- Write access - this is only for reads

sajidanower23:
### What this PR does

Fixes #2609 

- Inserts `recordId` when uploading a file
- Checks for `recordId` when downloading the file

### Checklist

-   [x] I've updated the README to reflect the API breaking changes


magda-storage-api/src/test/storage.spec.ts
```diff
@@ -236,47 +261,147 @@ describe("Storage API tests", () => {
...
+            registryScope
+                .get(
+                    "/records/storage-test-dataset-id-failure?optionalAspect=publishing&optionalAspect=dataset-access-control"
+                )
+                .reply(404);
+        });
+
+        describe("Uploading and then downloading a simple file", () => {
+            it("With registry auth", () => {
+                console.log("running test 1");
```
AlexGilleran: do we still need these console.log s?

magda-storage-api/src/test/storage.spec.ts
```diff
@@ -236,47 +261,147 @@ describe("Storage API tests", () => {
...
+                            .set("Accept", "text/plain")
+                            .expect(200)
+                            .expect("Testing download")
+                            .expect("Content-Type", "text/plain")
+                    );
+                });
+            });
+
+            it("Without registry auth", () => {
+                console.log("running test 2");
```
AlexGilleran: same here re `console.log`

magda-storage-api/src/test/storage.spec.ts
```diff
@@ -236,47 +261,147 @@ describe("Storage API tests", () => {
...
+                });
+            registryScope
+                .get(
+                    "/records/storage-test-dataset-id-failure?optionalAspect=publishing&optionalAspect=dataset-access-control"
+                )
+                .reply(404);
+        });
+
+        describe("Uploading and then downloading a simple file", () => {
+            it("With registry auth", () => {
```
AlexGilleran: I think a better name for this would be `with successful registry auth`

magda-storage-api/src/test/storage.spec.ts
```diff
@@ -236,47 +261,147 @@ describe("Storage API tests", () => {
...
+                            .set("Accept", "application/json")
+                            .set("Accept", "text/plain")
+                            .expect(200)
+                            .expect("Testing download")
+                            .expect("Content-Type", "text/plain")
+                    );
+                });
+            });
+
+            it("Without registry auth", () => {
```
AlexGilleran: and this should be something like `with unsuccessful registry auth`

magda-storage-api/src/test/storage.spec.ts
```diff
@@ -236,47 +261,147 @@ describe("Storage API tests", () => {
...
+                                    bucketName +
+                                    "/download-unauth-test-file-1"
+                            )
+                            .set("Accept", "application/json")
+                            .set("Accept", "text/plain")
+                            .expect(404)
+                    );
+                });
+            });
+        });
```
AlexGilleran: It'd be good to have a test for what happens when the file doesn't have a record id attached (upload and download) as well.

AlexGilleran:
Looks like some property tests are generating invalid uuids 😬

magda-storage-api/src/test/storage.spec.ts
sajidanower23: Added

sajidanower23:
@AlexGilleran  @t83714  I have made a deployment to https://2609-get-auth-storage.dev.magda.io.

To test:
```
# Uploading the file
curl -X PUT -H "Content-Type:text/plain" --cookie '<cookie>' https://2609-get-auth-storage.dev.magda.io/api/v0/storage/magda-bucket/myFile?recordId=ds-dga-ab7eddce-84df-4098-bc8f-500d0d9776d1 --data '@/some/file'

# Downloading the file
curl --cookie 'connect.sid=<cookie>' https://2609-get-auth-storage.dev.magda.io/api/v0/storage/magda-bucket/myFile
```

magda-storage-api/src/createApiRouter.ts
```diff
@@ -127,23 +132,58 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
 
         const stream = await object.createStream();
         if (stream) {
             stream.on("error", _e => {
-                res.status(500).send("Unknown error");
+                return res.status(500).send("Unknown error");
             });
             stream.pipe(res);
         }
+        return undefined;
```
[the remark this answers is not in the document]
sajidanower23: Yeah we need this. Since we are explicitly `return`ing from other code paths, every possible code path needs to return something. I have made it return a `500` now.

magda-storage-api/src/createApiRouter.ts
```diff
@@ -127,23 +132,58 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
+                        "No such object with fileId " +
+                            fileId +
+                            " in bucket " +
+                            bucket
+                    );
+            } else {
+                return res.status(500).send("Unknown error");
+            }
+        }
+        const recordIdNum = res.getHeader("Record-ID"); //.toString();
```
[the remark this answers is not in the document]
sajidanower23: Removed

AlexGilleran:
@sajidanower23 had a noodle around with it, seems to work OK now.

What was stopping it working in the cluster was that the registry url wasn't getting passed through helm, so it worked fine _locally_, but as soon as you deployed it via K8S it'd be trying to find the registry at localhost:6101 :(. So fixed that.

Before I realised that I was trying to run down the bug via testing, so I've written a bunch more tests - I've separated the registry stuff into its own `describe` that runs the same tests against the upload and PUT methods of adding files. I also changed some of the test descriptions to be a bit more BDD - i.e. rather that "upload file", "it should return 200 when a new file is uploaded". Makes it a bit easier to figure out what's wrong when you get a failure :).

Let me know what you think!

Issue #2627, opened 2019-11-26
Break out non-core components into separate repositories


AlexGilleran:
## Description
If we expect there to be a Magda ecosystem that involves third-party minions and connectors existing in separate repositories, we should ourselves be developing our own minions and connectors in this way, so that we can eat our own dogfood.

Currently all our minions and connectors exist in the monorepo, so they'll need to be moved out into separate repositories, and we'll need to make changes to the core repositories.

I think this should happen roughly like this (but it doesn't have to):

1. Modify our `lerna` setup so that when we publish new versions of Magda, we also push utility libraries like `magda-typescript-common` and `magda-minion-framework` to NPM for use by other projects.
2. Create new repos for all minions and connectors. If there's some clever way of using git to do this to preserve commit history that'd be awesome, but if we've just got to copy-paste then that's life. DON'T remove them from the core repo yet.
3. Hook up devops for each of these new repos - they each need
  - CI
  - A process to build a docker image
  - A helm chart and a process to publish the helm chart
4. Once these work well, remove them from the core repository and make the magda chart dependent on their charts.

## Acceptance Criteria

-   Code for minions and connectors is no longer in the core magda repository
-   A deployment of Magda still deploys connectors / minions just as it did before
-   Minions / Connectors still have CI, scripts to build docker images etc... this might mean publishing magda-scripts and using it outside the core magda repository.
-   Shared libraries (e.g. typescript common) are now available on NPM as javascript w/ separate typescript types - no one using our libraries should have to worry about doing any special build in order to use it.

## Out of Scope

- Figuring out how we manage issues across multiple projects... I'm guessing we'll have to lean harder into Zenhub but we'll tackle that afterwards.

PR #2742, opened 2020-02-18
Add simple document of building connectors & minions

t83714:
### What this PR does

@AlexGilleran 

Think better add some simple document to explain how to build connectors & minions. Hence, this PR.
Please help to update if any of those sentences do not look right 😄

Didn't create a sperate ticket as it's probably related to the previous ticket #2627 

Also:
- Commit the requirement.lock for helm chart Magda
- Replace `helm dep up` with `helm dep build` --- `helm dep build` will not give you an error if lock file is there and charts files are not. When the lock file is not there, it will act exactly like `helm dep up`
- set buildtest:helm-charts dependencies to [] to avoid downloading artifacts (so that it runs faster)
- also add `buildtest:helm-charts` as a dependency to publishing helm chart step so that we don't need to re-download the charts

### Checklist


t83714:
Thanks for the updates 👍

PR #2743, opened 2020-02-18
Issue/2701 Fixed: minio chart will not be deployed if storage-api is not turned on

### LINKED ISSUE #2701: Minio deploys regardless of tags
**Describe the bug**
Right now `minio` is included with all deployments regardless of what tags are passed - in particular this means for ui-only previews, we're allocating a full stateful set for minio every time we deploy, which is more expensive than we'd like.

**To Reproduce**
Deploy via helm and set all `tags` to `false`.

**Expected behavior**
Nothing gets deployed

**Screenshots**
```
kubectl get pods
NAME                           READY   STATUS    RESTARTS   AGE
magda-minio-7b6499644f-rrtd2   1/1     Running   0          19m
```

**Additional context**
This is probably because something is missing in `requirements.yaml`

t83714:
### What this PR does

Fixes #2701 

Fixed: minio chart will not be deployed if storage-api is not turned on

If you test it locally, please run:
```
helm dep build deploy/helm/magda
```
after you switch branch.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@ Storage:
 -   Add endpoint to create a bucket
 -   Restrict file upload to admins only
 -   Support multipart upload
+-   Fixed: minio chart will not be deployed if storage-api is not turned on
 
 Gateway:
 
--- a/deploy/helm/magda-core/requirements.yaml
+++ b/deploy/helm/magda-core/requirements.yaml
@@ -71,6 +71,10 @@ dependencies:
     tags:
       - all
       - opa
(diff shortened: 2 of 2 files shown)


t83714:
Close it first before try if a better solution can be adopted

t83714:
Was trying to add official minio chart as dependencies of storage API chart (so that wether it will be created or not will be fully depend on storage api)--- it turns out we need to move charts out from magda-core/charts.

Will create a seperate ticket for it

Issue #2714, opened 2020-01-28
Add "Link to an API or web service" to the Add Dataset flow


AlexGilleran:
### Description
We need to add the "Link to an API or web service" section to the UI.

![image](https://user-images.githubusercontent.com/900555/73239807-4fcd4d80-41f1-11ea-90dd-83056fbba211.png)

https://projects.invisionapp.com/share/KMVQ7DH652J#/screens/402485642

This issue is simply to add UI components that can be built on later.

This is the equivalent of #2713 but for the "Link to an API or web service" part.

### Flow
1. Starts with an empty text box. The user puts in a URL (or any string) and hits "Fetch".
![image](https://user-images.githubusercontent.com/900555/73224164-b7b57100-41bc-11ea-8541-fd81ee26cb40.png)
2. The loading state shows for 1 second (fake this with `setTimeout`). The label of the loading state should be the URL.
![image](https://user-images.githubusercontent.com/900555/73239643-d6cdf600-41f0-11ea-935b-7e1164a73a26.png)
3. The URL is pre-populated in a "review" box. The text box / fetch button are still present below, so that more URLs can be added. The review box has a URL but nothing else (yet) - those can be added by the user.
![image](https://user-images.githubusercontent.com/900555/73224214-e895a600-41bc-11ea-9afc-a7c44a84044b.png)

### Acceptance Criteria
- The UI elements in the design are added to the page
- The user can put in a URL and hit "Fetch", and this simply adds the URL as a new distribution with no metadata other than the URL itself, which can be edited.

### Out of Scope
- Extracting any metadata from the link
- Validation

PR #2745, opened 2020-02-20
Issue/2713 Added UI for "Link to dataset already hosted online" box

### LINKED ISSUE #2713: Add UI for "Link to dataset already hosted online" box in Add Dataset Flow
### Description
We need to add the "Link to dataset already hosted online" section to the UI.

image
image
https://projects.invisionapp.com/share/KMVQ7DH652J#/screens/402485642

This issue is simply to add UI components that can be built on later.

### Flow
1. Starts with an empty text box. The user puts in a URL (or any string) and hits "Fetch".
image
2. The loading state shows for 1 second (fake this with `setTimeout`). The label of the loading state should be the URL.
image
3. The URL is pre-populated in a "review" box. The text box / fetch button are still present below, so that more URLs can be added. The review box has a URL but nothing else (yet) - those can be added by the user. All the fields should be able to be manually added though!
image

### Acceptance Criteria
- The UI elements in the design are added to the page
- The user can put in a URL and hit "Fetch", and this simply adds the URL as a new distribution with no metadata other than the URL itself, which can be edited.

### Out of Scope
- Extracting any metadata from the link
- Validation

t83714:
### What this PR does

Fixes #2713 

Added UI for "Link to dataset already hosted online" box

image

image

image

PR https://github.com/magda-io/magda/pull/2747 is created for #2714 based on this PR.

### Checklist


magda-web-client/src/Components/Dataset/Add/AddDatasetLinkSection.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useState } from "react";
+import {
+    File,
```
AlexGilleran: Possibly if these are no longer "Files" we should name them to something that takes into account that it could be either a file or a service, like "Distribution"

magda-web-client/src/Components/Dataset/Add/AddDatasetLinkSection.tsx
```diff
@@ -0,0 +1,99 @@
...
+                downloadURL: url,
+                creationSource: FileSource.DatasetUrl,
+                title: url,
+                modified: new Date(),
+                format: "",
+                _state: FileState.Processing,
+                _progress: 50
+            });
+
+            setUrl("");
```
AlexGilleran: This isn't working for some reason - the url box still has the URL in it after you hit fetch:
image

magda-web-client/src/Components/Dataset/Add/AddDatasetLinkSection.tsx
t83714: 👍 The name looks confusing for me as well 😄 
Will change it to `Distribution`~

magda-web-client/src/Components/Dataset/Add/AddDatasetLinkSection.tsx
t83714: It's strange... Checked the preview site in the cluster and it seems the line is not there:
image
Probably gitlab didn't checkout correct commit?
Will look into it~

t83714: Yeah~ it seems a racing condition of the CI pipeline~
I re-run the UI docker build step, re-deploy the UI preview and it seems works ok now~

t83714:
@AlexGilleran 
Go through code and update file / files reference where more appropriate to distribution / distributions.
Also, update the default validation config. I made the validation manager now understand both `files` and `distributions` terms. Therefore, no need to update the existing site config.

I think it's ready to have another look

PR #2748, opened 2020-02-21
Upgraded OPA and added policy-per-record auth to getting a record

### LINKED ISSUE #2362: Add authorization to getting a record's top-level metadata (name/id)

AlexGilleran:
### What this PR does

Fixes #2362 

- Makes auth work in the way described in #2362 
    - Every record now has an `authnReadPolicyId` column, that determines what policy to look up in OPA to authorise access to that record
    - The existing policy id set in config now determines a default fallback policy if this column is null (so that existing implementations, particularly the digital twin, still work).
    - The fallback policy, by default, is a new one - "public" - that just allows everything, a la data.gov.au etc
- Upgraded OPA to the latest version. On the upside, this new version falls back to "support" way less and uses queries way more, on the downside this broke the existing Content API authn, so I've rewritten the parser for that to rely on `queries` and not `support` like the registry api one does. If we had complex content policies this could be a problem but we don't so it's OK for now (I think, please try to prove me wrong!)
- The old views that were used for determining whether something is public are now gone because `authnReadPolicyId = public` now does that.

### Warts
- Because the content api tests mock out the `Database` class, and the OPA logic is all in there, essentially none of the OPA changes to the content api are tested. We should resolve this... but later, given that progressing registry auth is way more important than progressing content auth.

### Checklist


magda-registry-api/README.md
```diff
@@ -51,7 +51,7 @@ users in different organization units. See the comment in the sql file for detai
 ### Start opa server
 
 ```
-cd deploy/helm/magda-core/charts/opa
+cd deploy/helm/magda/charts/magda-core/charts/opa
```
t83714: `magda-core` chart files are not in `magda` 's chart directory so original path should be correct?

AlexGilleran: Oops yeah... I moved them and moved them back, looks like I missed a spot. Good catch :)

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RegistryApp.scala
```diff
@@ -60,9 +60,9 @@ object RegistryApp extends App {
   logger.info("Starting MAGDA Registry")
 
   GlobalSettings.loggingSQLAndTime = LoggingSQLAndTimeSettings(
```
t83714: Would it be easier if we set debug logging in [application.conf](https://github.com/magda-io/magda/blob/45d5f800449aee5cc676f0c921906cd2e2f43615/magda-registry-api/src/main/resources/application.conf) ?
e.g.:
```
scalikejdbc{
    global {
        loggingSQLAndTime.enabled = true
        loggingSQLAndTime.singleLineMode = true
        loggingSQLAndTime.logLevel = DEBUG
        loggingSQLAndTime.warningEnabled = true
        loggingSQLAndTime.warningThresholdMillis = 5
        loggingSQLAndTime.warningLogLevel = warn
    }
}
```
In this way, we probably can override the log setting through runtime parameter without re-compiling the registry-api / docker image.

magda-registry-api/README.md
```diff
@@ -51,7 +51,7 @@ users in different organization units. See the comment in the sql file for detai
 ### Start opa server
 
 ```
-cd deploy/helm/magda-core/charts/opa
+cd deploy/helm/magda/magda-core/charts/opa
```
t83714: probably should be `deploy/helm/magda-core/charts/opa` (without /magda/).

magda-typescript-common/src/OpaCompileResponseParser.ts
```diff
@@ -968,33 +970,68 @@ export default class OpaCompileResponseParser {
...
+            (!_.isArray(this.data.support) || !this.data.support.length)
+        ) {
             // --- mean no rule matched
             return [];
         }
+
+        const queries = this.data.queries;
+
+        if (queries) {
+            queries.forEach((q: any, i: number) => {
```
t83714: Probably easier (and more generic as we can handle any response no matter whether `support` / `queries` are there or not) to convert `this.data.queries` (answer) to a package.
Compared with packages in `support`, the answer/queries section contains rules body without package namespace.
We probably can put all answer queries into a fake namespace `magda.query.answer` (package path `magda.query` & fake rule name `answer`).
And add it to the support packages (if support is empty, we just create an array with package name `magda.query.answer`) so the same rule drying logic can be applied.
Then, we can call `evaluateRule` the fake rule `magda.query.answer`

Ignore me if the current code works well already 😄

deploy/helm/magda-core/charts/opa/templates/deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
 {{- end }}
       containers:
       - name: opa
-        image: openpolicyagent/opa:0.10.7
+        image: openpolicyagent/opa:0.17.1
```
t83714: We probably need to update deploy/helm/magda-core/charts/opa/docker-compose.yml as well

AlexGilleran: 👍

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RegistryApp.scala
AlexGilleran: 😬 didn't mean to commit this but that sounds good. Although I think we can already override it, but having it in application.conf makes sense, I'll make it so 👍

magda-registry-api/README.md
AlexGilleran: double oops 😬

magda-typescript-common/src/OpaCompileResponseParser.ts
AlexGilleran: I wouldn't say well, but it does work afaik.

Maybe write an issue laying out a design for what you mean? I don't think it makes sense to invest any more time into auth for the content api right now, It'd be cool to have for later :).

t83714: 👍

AlexGilleran:
@t83714 I think it's ready for another look now. I ended up deleting the dataset auth stuff out of the authorization API seeing as (I think?) this registry auth stuff should supercede it, if that's OK.

t83714:
> @t83714 I think it's ready for another look now. I ended up deleting the dataset auth stuff out of the authorization API seeing as (I think?) this registry auth stuff should supercede it, if that's OK.

I think it's for generating a list of approvers (on review screen) for sending notifcations~
Yeah~ Let's remove it~ I think we are a new plan for approval process now

PR #2749, opened 2020-02-24
Upgrade kubernetes

### LINKED ISSUE #2727: Upgrade to latest Kubernetes and Helm
## Description
Recently there have been a few developments in Kubernetes land:

- Helm 3 has been released (https://helm.sh/docs/faq)
- Kubernetes 1.16 has been released, which is incompatible with some of our charts.

We should upgrade - and I think it makes sense to group these together to make the test / release a bit easier.

### Acceptance Criteria
- Magda now uses Helm 3 instead of Helm 2
    - Documentation has been updated to reflect this - in particular there should be no instructions about creating a user for tiller anymore seeing how there is no tiller
    - Our helm setup has been modified to use helm 3 - in particular, we'll need to move the contents of `requirements.yaml`
    - Deployment via CI still works
    - Deployment via command line still works
    - Deployment via `terraform` still works
- Magda is compatible with the latest version of Kubernetes
    - Magda can be deployed onto a kubernetes environment, with all services, without any warnings about API deprecation
- Write a migration guide
- Make sure all the non-core repos are also compatible

sajidanower23:
### What this PR does

Fixes #2727 

Upgrades the charts so that they can be used with the latest versions of Helm, Kubernetes and Minikube.

Tested on:
```
$ minikube version 
minikube version: v1.7.2
commit: 50d543b5fcb0e1c0d7c27b1398a9a9790df09dfb

$ kubectl version 
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}

$ helm version 
version.BuildInfo{Version:"v3.1.0", GitCommit:"b29d20baf09943e134c2fa5e1e1cab3bf93315fa", GitTreeState:"clean", GoVersion:"go1.13.7"}
```

### Checklist

-   ~[ ] I've added instructions for making the mentioned upgrades~
-   [x] I've tested that the changes work in Linux (Ubuntu 18.04)
-   [x] I've tested that the changes work in Mac


docs/docs/building-and-running.md
```diff
@@ -110,6 +110,20 @@ kubectl apply -f deploy/kubernetes/local-storage-volume.yaml
 
 Note: If using docker desktop for Windows older than version 19, change the value from "docker-desktop" to "docker-for-desktop" in nodeAffinity in file deploy/kubernetes/local-storage-volume.yaml
 
+### Install the CKAN connector
+
+The Magda CKAN connector (as of v0.0.57) lives outside of the core repository, at https://github.com/magda-io/magda-ckan-connector.
```
AlexGilleran: Should probably say why this is important :)

sajidanower23: Well, why is it important? :see_no_evil:

AlexGilleran:
Can't deploy it via Gitlab: https://gitlab.com/magda-data/magda/-/jobs/447275123

```
 Error: validation failed: unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1beta1"
```

docs/docs/building-and-running.md
AlexGilleran: 😂 because whoever runs this is going to need it when they do their first deploy.

I'm thinking something like "You'll need this in order to populate your magda instance with an initial set of datasets from data.gov.au. If you don't want to get datasets from data.gov.au, you can skip this and set `connectors.initialCrawl` (or whatever it is? I'm not sure) to `false` in `config.yaml`."

t83714:
It seems helm3 will give you errors if:
- `requirements.lock` is there 
- no `charts` directory or empty (i.e. it needs to install the dependencies)
- and you try to run:
```
helm dep build deploy/helm/magda
``` 
(This doesn't happen in CI because we still use helm 2 in CI. @AlexGilleran Should we upgrade to helm3 in CI as well?)

The reason why that could happen is helm3 allows (and recommend) you to put `dependencies` into `chart.yaml` rather than a separate `requirements.yaml`.
Because of that, helm3 will generate a `chart.lock` file (similar to `requirements.lock`) and will report an error if requirements.lock is there.

Possible solution (works for both helm2 & helm3):
- Delete the `requirement.lock` from the repo.
- Still leave `dependencies` info to `requirements.yaml` so it works with helm2 
- Still, run `helm dep build` to install the dependencies

.gitlab-ci.yml
```diff
@@ -603,23 +603,24 @@ inttest:registryAuth:
     - echo "$DB_SERVICE_ACCOUNT_PRIVATE_KEY" > db-service-account-private-key.json
     - kubectl create secret generic storage-account-credentials --from-file db-service-account-private-key.json --namespace $CI_COMMIT_REF_SLUG --dry-run -o json | kubectl apply --namespace $CI_COMMIT_REF_SLUG -f -
     - kubectl create secret generic storage-secrets --from-literal=accesskey=$MINIO_ACCESS_KEY --from-literal=secretkey=$MINIO_SECRET_KEY --namespace $CI_COMMIT_REF_SLUG --dry-run -o json | kubectl apply --namespace $CI_COMMIT_REF_SLUG -f -
-
-    # Helm upgrade
-    - helm init --upgrade
+    - helm version
+    - helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator
```
t83714: You don't need to add repo here (or update the repo).
All dependencies are resolved (and downloaded) at `buildtest:helm-charts` step.
And helm charts files are passed to all deploy steps via artifacts.
You should make sure `buildtest:helm-charts` step run OK as that where the dependencies are resolved & downlaoded

sajidanower23: Yep, you are right. I was using the old image for that part. Removed the update repo and add repo

docs/docs/building-and-running.md
sajidanower23: Added a note on this

PR #2764, opened 2020-03-06
Added authorisation to links through the single-record (records/id) API

### LINKED ISSUE #2368: Apply authorisation to dereferenced records when requested via /record/id API
### Problem description
Building on #2364, this applies authorisation to aspects retrieved through links when `dereference=true` is specified through the [/record/{id}](https://dev.magda.io/api/v0/apidocs/#api-Registry_Record_Service-GetV0RegistryRecordsId) endpoint in the registry.

### Acceptance Criteria
(assuming `dereference=true` is specified)
- If a user is allowed to see a requested record aspect, and that aspect links to another record  that the user is allowed to see, the inner aspect should be returned as part of the outer one, just as it was before
- If the user is allowed to see an outer aspect but not the inner aspect it links to, there should be no inner aspect present - as if the inner record aspect row didn't exist. **(e.g. if the dataset-distributions schema links to a single distribution record but the user isn't allowed to see that distribution, the `distributions` property should be `[]` but you should still be able to see the dataset-distributions aspect.**
- If the user is not allowed to see the outer aspect but is allowed to see the inner one, they still shouldn't be able to see either when dereferencing.

**(with `deference=false`)**
  - **The id of dereferenced records that the user doesn't have access to should be suppressed... e.g. in the example above where a user has access to a dataset record but not the distribution, the `distributions` property should also be `[]`. **

### Technical Notes
- This should be achievable using similar logic to getting records without dereferencing. Dereferencing works by examining the aspect definition, pulling out linked aspects and then querying for them. When the linked aspects are queries, it should be possible to look up the possible policies from OPA, turn them into SQL and include them in the query for the inner aspect.

### Out of Scope
- Aspect-level authorisation

AlexGilleran:
### What this PR does

Merge #2762 first! This includes it so that int tests would work.

Fixes #2368 

This adds authorization to records that link to other records. So if a record has an aspect with a link defined in the schema, this restricts the records that come through that link - so if you don't have access to record A, you can't see it or become aware of its existence, even if you do have record B that links to it. This carries whether `dereference` is true or false - if dereference is false, you shouldn't even be able to see the ID of linked record, if dereference is true you shouldn't be able to see the inner record (unless you have access).

### To test
There's a test env up at https://issues-2368-single-record-dereference-auth.dev.magda.io, which I've pre-populated with some data.

There are four inner records - `inner-dataset-access-control-1` and `inner-dataset-access-control-2`, which have `object.registry.record.owner_orgunit` as their policy, and `inner-dataset-access-public-1` and `inner-dataset-access-public-2`, which have `object.registry.record.public` as their policy.

Then there's records that link to them. These either have `aspect-with-link-single` and `aspect-with-link-multiple` aspects, which have a single link and an array link respectively.

The records I've made are:
- `outer-public-inner-single-public` (`aspect=aspect-with-link-single`): you should be able to see both inner and outer datasets regardless of being logged in or out
- `outer-public-inner-single-control` (`aspect=aspect-with-link-single`): logged out, you should only be able to see the outer dataset, logged in as admin you should be able to see both inner and outer
- `outer-control-inner-single-public` (`aspect=aspect-with-link-single`): logged out, you shouldn't be able to see anything, logged in as admin you should be able to see inner and outer
- `outer-public-multiple` (`aspect=aspect-with-link-multiple`): logged out, you should be able to see two inner records, logged in as admin you should see all 4
- `outer-control-multiple` (`aspect=aspect-with-link-multiple`): logged out, you should see nothing, logged in as admin you should see four inner datasets

### Checklist


AlexGilleran:
Don't delete this deployment until after demos if you merge pls!!! 😬

t83714:
Just logged in with a normal standard account and see this:
image
For this one:
> outer-public-inner-single-public (aspect=aspect-with-link-single): you should be able to see both inner and outer datasets regardless of being logged in or out

It seems I can't see any inner record for that `outer-public-inner-single-public` one?

Check the dataset `inner-dataset-access-public-1` it does have `object.registry.record.public` assigned as `authnreadpolicyid`.

Here is the aspect content for `outer-dataset-access-public-inner-access-public`:
```json
{"innerId": "inner-dataset-access-public-1"}
```
Probably because `innerId` must be an array (only guessing)?

AlexGilleran:
@t83714 this is what I get when logged out for that one:

GET https://issues-2368-single-record-dereference-auth.dev.magda.io/api/v0/registry/records/outer-public-inner-single-public?aspect=aspect-with-link-single&dereference=true
```json
{
  "aspects": {
    "aspect-with-link-single": {
      "innerId": {
        "aspects": {
          "dataset-access-control": {
            "ownerId": "3848a5b8-1ec7-4d94-9e65-990bd01cb9da"
          }
        },
        "id": "inner-dataset-access-public-1",
        "name": ""
      }
    }
  },
  "authnReadPolicyId": "object.registry.record.public",
  "id": "outer-public-inner-single-public",
  "name": "",
  "tenantId": 0
}
```

and

GET https://issues-2368-single-record-dereference-auth.dev.magda.io/api/v0/registry/records/outer-public-inner-single-public?aspect=aspect-with-link-single&dereference=false
```json
{
  "aspects": {
    "aspect-with-link-single": {
      "innerId": "inner-dataset-access-public-1"
    }
  },
  "authnReadPolicyId": "object.registry.record.public",
  "id": "outer-public-inner-single-public",
  "name": "",
  "tenantId": 0
}
```

Are you sure you requested the right aspect?

PR #2775, opened 2020-03-18
Added auth to links from the Registry /records (plural) endpoint

### LINKED ISSUE #2371: Add authorisation for dereferencing on records (plural) API
### Problem description
Building on #2370 and #2368, this covers applying authorisation to the dereferenced aspects returned when using the [/records](https://dev.magda.io/api/v0/apidocs/#api-Registry_Record_Service-GetV0RegistryRecords) endpoint (i.e. when `dereference=true` is set).

### Acceptance Criteria
- If a user is allowed to see both the requested aspect and the inner aspect that it links to, both should be returned (one inside the other, as happens now)
- If a user is allowed to see the outer aspect but not the inner one, the outer aspect should be returned but not the inner one (as if the inner one didn't exist)
- If a user is allowed to see the inner aspect but not the outer one, they should be able to see neither as if the outer one didn't exist.

### Technical Notes
This should be very similar to the technical solution for #2368 but applied to multiple records.

AlexGilleran:
### What this PR does

Fixes #2371 

This adds auth for links to the `/records` endpoint (previously we implemented only `records/<recordid>`. So essentially, when you call `/records` on the registry, inner records should appear or not appear based on the policy set on those records.

To test:
Go to https://issues-2371-links-in-records-plural.dev.magda.io/api/v0/registry/records?aspect=aspect-with-link-array while logged out, you should see that `record-outer-multiple-public` has two inner records, `record-outer-multiple-mixed` has one inner record, and `record-outer-multiple-none-allowed` has no inner records. If you log in and call it again, all should have two inner records.

Similarly, if you go to https://issues-2371-links-in-records-plural.dev.magda.io/api/v0/registry/records?aspect=aspect-with-link-single while logged out you should see that `record-outer-single-public` has an inner record and `record-outer-single-private` has no inner record. Sign in and they should both have an inner record.

### Checklist


t83714:
Tested and worked perfectly on test site 🎉 
probably merge it after the demo

PR #2787, opened 2020-03-30
Tests for auth on /records/summaries API

### LINKED ISSUE #2374: Apply authorisation to the records returned from record summaries (plural) API
### Problem description
Building on #2372 and #2369, apply authorisation to the [record summaries API](https://dev.magda.io/api/v0/apidocs/#api-Registry_Record_Service-GetV0RegistryRecordsSummary) so that only records that the user has access to are returned. 

### Acceptance Criteria
- Only records with a authreadpolicy that allows the user access should show up
- If the user requests 10 records and there's >= 10 records that they have access to, they should get those 10 records back (i.e. getting 10 records from the database then filtering them down isn't an option)
- The number of requests to the DB should be minimised - i.e. this almost certainly needs to use OPA's partial compilation to create an SQL query that includes access policies - continually making requests to the database and filtering them by policy in the application is also not an option.

### Technical Notes
- This should be able to piggy back off #2369's implementation

### Out of Scope
- Filtering the aspect ids in the summary (done in another issue)

AlexGilleran:
WARNING: Follows on from #2782, merge that first!

### What this PR does

Fixes #2374 

The issue is to add auth for the summaries API, but that already works so this is just tests.

#### To test:
- Go to https://issues-2374-authorisation-on-summaries-plural.dev.magda.io/api/v0/registry/records/summary
- There's `record-private`, which has an `owner_orgunit` policy, and `record-public`, which is public
- If you're logged in and have access to `record-private`, you should be able to see both, if you're logged out you should only be able to see one.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,9 +59,10 @@ Authorization:
 -   Added ability to set per-record authorization policies in the registry (for getting a single record)
 -   Added ability to set per-record authorization policies in the registry (for getting multiple records)
 -   Added ability to use OPA policies that use data types other than strings in the registry
--   Added authorization inside links with dereferencing on or off, for the `records/<id>` endpoint
--   Added authorization inside links with dereferencing on or off, for the `records` endpoint
--   Added per-record authorization around the `records/summary/<recordid>` endpoint matching the `records/<id>` one.
+-   Added authorization inside links with dereferencing on or off, for the `/records/<id>` endpoint
+-   Added authorization inside links with dereferencing on or off, for the `/records` endpoint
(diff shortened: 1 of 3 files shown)


t83714:
@AlexGilleran  Are we supposed to allow admins to see all records?
I created an admin (login: `testmagda` pass: `123456`) and without changing the access control aspect, I am able to see both datasets. Once logged out, I only see one dataset.:
image

AlexGilleran:
I think so :grimacing: . I mean, whether it's business-logic correct or not is an open question, but it's there in the owner_orgunit policy so I think it's ok?

```rego
read {
    admin_role
}

read {
    has_permission.read
    owner
}

read {
    has_permission.read
    orgunit
}
```

PR #2788, opened 2020-03-30
Minio as dep

### LINKED ISSUE #2744: Move minio as a dependency of storage-api
## Description
We currently copy Offical `minio` helm chart files and use it as part of `magda-core`.
We probably want to move it as a dependency of storage-api.
Benefit:
- Don't need to maintain our `minio` chart copy & have access to latest official version
- minio will be installed only if storage api is installed

Note: all `mage-core` sub-charts are located in `deploy/helm/charts` at this moment. 
This will create troubles for installing dependencies (as helm will empty `charts` directory first)
We need to move them out all sub-charts and add version & repository fielda to all sub-charts on `magda-core`'s `requirements.yaml` as the followings (or similar depends on the actual file path):
```yaml
version: 0.0.57-0
repository: file://../[xxxx]
```

## Acceptance Criteria
- the `minio` chart files should be deleted from repo
- the offical minio chart should be added to `requirements.yaml` of storage-api as a dependency:
```yaml
dependencies:
  - name: minio
    version: 5.0.18
    repository: "@stable"
```
- Update CI script accordingly
- Make sure all CI deploy buttons work OK
- Make sure helm charts publishing work ok
- Update the document --- we now require user to add `stable` repo in order to install all dependencies

sajidanower23:
### What this PR does

Fixes #2744 

Specifies MinIO as a dependency of storage's helm chart

### Checklist

-   [x] Update CI script


deploy/helm/internal-charts/storage-api/values.yaml
```diff
@@ -9,3 +9,8 @@ resources:
 minio:
   host: "minio"
   port: 9000
+  nameOverride: "magda-minio"
+  fullnameOverride: "magda-minio"
+  certsPath: "/etc/minio/certs/"
```
t83714: We probably don't need set `certsPath ` & `configPathmc ` as the value we set here is same as the default value?

PR #2791, opened 2020-04-01
Added auth to the records/<recordId>/history endpoint

### LINKED ISSUE #2376: Add authorisation for records on record history API
### Problem description
Apply record-level authorisation to the https://dev.magda.io/api/v0/apidocs/#api-Registry_Record_History-GetV0RegistryRecordsRecordidHistory endpoint. This involves only allowing access to the history of a record if the user has access to the record in general.

### Acceptance Criteria
- If the user does not have access to a record via the `records/{recordId}` endpoint, they should get a `404` from this endpoint also, as if the record didn't exist.
- If the user does have access to the record, they should be able to view all the events as normal - don't worry about securing specific events yet.
- If a record no longer exists (because it has been deleted), then respond with 404 unless the user is an admin (for now, we can figure out a more comprehensive solution later - it's fractally complicated otherwise). If the user is an admin then they can see the deleted record's history regardless of permissions.

### Technical Notes
- It should be possible to use a similar strategy as for #2369 for this

### Out of Scope
- Finer-grained authorisation of the events themselves - for now as long as you have access to a record you can see the events

AlexGilleran:
(branches off https://github.com/magda-io/magda/pull/2787, merge that first!)

### What this PR does

Fixes #2376 

This adds authorization to the records/<recordId>/history endpoint. This mostly follows the same logic as the record endpoint and the record summary endpoint, with a few exceptions:
- Current behaviour is that if a record can't be found, rather than return 404 the endpoint returns 200 with `events: []`, so if a user doesn't have access to a record, this is what's returned.
- If a record has been deleted, users will no longer be able to see any events from the record unless they're an admin - this is so that:
    - If sensitive data gets added accidentally, there's a quick and simple way to hide it
    - It avoids having to implement really complex logic around how else we could solve this problem... e.g. possibly we could figure out the record's last state and use that to determine whether the user should be allowed to see it, but then what if you want to change the policy to hide the history after that? You'd have to undelete it somehow, change the policy and then redelete it?
        - We might have to revisit this later but I don't anticipate that will be soon

#### To Test
There are 3 records in https://issues-2376-record-history-auth.dev.magda.io/:
- `record-public`should have its history viewable by anyone
- `record-private` should only be available to those who match the policy
- `record-deleted` was originally public, but has been deleted, so should only be viewable by admins.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,7 @@ Authorization:
 -   Added authorization inside links with dereferencing on or off, for the `/records` endpoint
 -   Added per-record authorization around the `/records/summary/<recordid>` endpoint matching the `/records/<id>` one.
 -   Added per-record authorization around the `/records/summary` endpoint matching the `/records` one.
+-   Added per-record authorization around the `/records/<recordid>/history` endpoint
 
 Others:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,9 +59,10 @@ Authorization:
 -   Added ability to set per-record authorization policies in the registry (for getting a single record)
 -   Added ability to set per-record authorization policies in the registry (for getting multiple records)
 -   Added ability to use OPA policies that use data types other than strings in the registry
--   Added authorization inside links with dereferencing on or off, for the `records/<id>` endpoint
--   Added authorization inside links with dereferencing on or off, for the `records` endpoint
--   Added per-record authorization around the `records/summary/<recordid>` endpoint matching the `records/<id>` one.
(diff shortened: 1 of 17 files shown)


t83714:
It seems just assign an admin role (without setting isAdmin=true) will not give you access to `record-private`. Probably `admin_role.rego` should be:
```
admin_role {
    input.user.roles[_].id = "00000000-0000-0003-0000-000000000000"
}
```
rather than:
```
admin_role {
    input.user.roles[_] = "00000000-0000-0003-0000-000000000000"
}
```
?
But it can't explain why isAdmin field play its role here (once I set isAdmin=true I got access) 🤔

PR #2793, opened 2020-04-02
Added auth to <recordid>/history/<eventId> (time-travel) API

### LINKED ISSUE #2378: Add authorisation for registry time-travel API
### Problem description
The [history API](https://dev.magda.io/api/v0/apidocs/#api-Registry_Record_History-GetV0RegistryRecordsRecordidHistoryEventid) allows a user to get the way that a record looked at some point in time (after a certain event).

This needs to have a similar protection as the `/record` API - on both the outer record and the inner aspects (fortunately there's no dereferencing!).

### Acceptance Criteria
- If a user has access to a record, they should have access to that record's history
- If a user doesn't have access to a record, they should get 404.
- If a user requests a version of a record at an event id for an event they don't have access to (as per #2377), they should get 404 as if that event didn't exist.
- The user should only be able to see aspects that they have access to in the resulting record - equivalent to the aspect filtering in #2373
- If a user requests a deleted record, they shouldn't be able to see it (currently this causes a 500 because it's broken, but it'd be better to make it 404 if possible)

### Out of scope
- Getting a deleted record after a certain event id - as far as I can see this is broken and never worked, so just make sure it doesn't work for this either!

AlexGilleran:
(merge after #2791!)

### What this PR does

Fixes #2378 

This adds authorisation to the version/time-travel API. Pretty much the same logic as for `/record/<recordid>/history` - if the user can see the record, then they can see the time-travel version of it, if they can't see the record in its current state, they can't see the old state either. Only exception is that if a record is deleted, only admins are allowed to see the record.

#### Testing
The time-travel API looks like /record/<recordid>/history/<eventid>, but it's fine to put in a really high event id if you want it to time time travel to the last possible event. I've set up 3 records similar to the last PR.
So
- https://issues-2378-time-travel-auth.dev.magda.io/api/v0/registry/records/record-private/history/10000 should only show you the history if you have access to the record (https://issues-2378-time-travel-auth.dev.magda.io/api/v0/registry/records/record-private returns 200), or if you're an admin.
- https://issues-2378-time-travel-auth.dev.magda.io/api/v0/registry/records/record-deleted/history/6 should only show you a record if you're an admin because the record has been deleted. A later event id than 6 will also show you nothing, because the record was deleted on eventId 7.
 - https://issues-2378-time-travel-auth.dev.magda.io/api/v0/registry/records/record-public/history/10000 should be the same for everyond

### Checklist


t83714:
Will leave the demo site open for now~

Issue #2778, opened 2020-03-23
Make our helm charts helm 3 chart


t83714:
Our helm charts can be used by both helm 2 and helm 3.
But not making it 100% helm 3 and still support helm 2 make it hard to access all helm 3 features.
And it would require more effort to keep it compatible with helm 2 and 3.
Moreover, helm 2 support will end on Nov 13, 2020 (no security fixes after that).

## Acceptance Criteria
- A migration guide for existing helm2-based installations has been written and tested
    - This should allow a user to migrate with no data loss and minimal downtime (preferably < 5 minutes?)
    - This should take into account both pure-helm and terraform-based installations
    - This should take into account the process of getting the new helm command line tool, running helm 2to3 etc - we don't need to be copy-pasting existing documentation about migrating helm 2 installations but we do need to make sure we have a step-by-step document that links out to those guides where necessary and has been tested as working.
- The need for Helm 3 is now present in all the docs that tell you how to install helm (magda-config and building-and-running.md)
- Chart publishing has been moved over to Helm 3 (this will require some investigation because the way that helm3 does this isn't clear to us at the moment)

## Technical Notes
We can do this by:
- Change `apiVersion` in `Chart.yaml` to `v2`
  - May also specify `kubeVersion` so helm won't let users try Magda on unsupported k8s
- move `requirements.yaml`'s content to `Chart.yaml`
- Remove `requirements.lock` and commit `Chart.lock` instead
- Update document and specify the helm3 requirement for installing Magda
- We probably need a new charts repo (e.g. helm3.charts.magda.io or something like that)

t83714:
Regarding the helm repo index command. Based on my test:
And helm v3 just will not generate index file include more versions.
It only generate one version.
And helm v2 works fine as usual (i.e. creating all versions) but only drawback is it will ignore apiVersion v2 charts.
Eventually, I found a way to create all versions index file properly (by looking at helm official repo scripts):
https://github.com/helm/charts/blob/c6e503ccfd9911dd06ea9ac389a40a3490637516/test/repo-sync.sh
We actually need to use --merge switch like:
helm repo index --merge "./index-test.yaml" chart-repo
So you will have index with multiple versions (and different apiVersion as well):
![image](https://user-images.githubusercontent.com/674387/77977168-adddf480-734a-11ea-95e5-ed8c76247b08.png)

PR #2765, opened 2020-03-06
Issue/2715 (Demo) Process remote data URL Solution based on openfaas

t83714:
### What this PR does

Fixes #2715 

### Test Site:

https://issue-2715.dev.magda.io/dataset/add

(Please don't delete the test site as I still need it for Tuesday demo. 😄 )

image

It supports 4 types of ckans urls (we can add support of more systems by updating more connectors):
- a CKAN dataset web url. e.g.  (https://demo.ckan.org/dataset/pruebas3)
  - Will import all distributions (resources belongs to that dataset / packages) to the UI
- a CKAN distribution web url. e.g.  (https://demo.ckan.org/dataset/pruebas3/resource/7c1a5e64-fefd-44db-9c98-d82ee79c4d32)
- a Ckan dataset API URL: https://demo.ckan.org/api/3/action/package_show?id=nombres-masculinos-mas-frecuentes-en-la-provincia-de-barcelona
  - Will add all distributions (resources belongs to that dataset / packages) to the UI
- a Ckan distribution API url: https://demo.ckan.org/api/3/action/resource_show?id=a6a23cd9-f2b7-4ed2-a4b8-a7ab593b01d7

How it works:
- The frontend will 
  - retrieve all functions with label `data-url-processor` (that's how a custom connector can add support of different systems. )
  - call all those functions
  - the first successful result will be used to populate distributions

You can find more test urls from:   https://demo.ckan.org/dataset

#### About Openfaas:

- Openfass is deployed into two namespaces
  - openfaas-fn: namespace for all functions i.e. the actual workload
  - openfaas: namespace for the gateway / controller part 
    - Reason for not deploying `openfaa`s core modules (gateway etc.) into `magda` namespace: 
`openfaas` helm chart comes with a `gateway` deployment as well 😄 
- Currently, `openfaas` is deployed separately, although I added it as the sub-chart of Magda (details see installation instruction below)
  - We can improve the deployment process around this --- Just need to customise the official helm chart a bit (official helm chart only allow you to specify the namespace of function and set core namespace to same as helm's `--namespace` parameter)
- You can 
  - deploy a function (to namespace `openfaas`) like ordinary k8s objects through k8s api. 
  - And Query them like ordinary k8s objects. e.g. `kubectl -n open-faas get functions`
[...]

PR #2795, opened 2020-04-02
Issue/2715 Extract Metadata from Data portal URL Solution using OpenFaas (Ready for Review)

### LINKED ISSUE #2715: Spike: How can we extract details from data portals through the UI?
## Description
For the new Add Dataset, we need to be able to support a flow where:
- The user puts in the URL of a dataset in a data portal and presses "Fetch"
- We retrieve that URL and somehow pull the metadata of the dataset and all the distributions off it and prepopulate that.

### To be determined:
- What's the best way to go from a URL of a landing page to the metadata for:
    - CKAN
    - Socrata
    - ESRI
    - GeoNetwork
    - Magda!
- What's the best way to retrieve metadata for:
    - WMS
    - WFS
    - ESRI REST
    - ??? what other APIs are popular
- Can we re-use the logic from the connectors?
- Should/can this be purely client side or do we need a server-side component?
   - If it should be server-side, maybe this might be a good opportunity to look at OpenFAAS or some other serverless thing?

### Acceptance Criteria
- A write-up has been added to this issue answering all the above
- Some kind of PoC branch exists that demonstrates:
   - Fetching metadata for ONE portal URL (CKAN, Socrata etc)
   - Fetching metadata for at least ONE kind of API (WMS, WFS etc)

### Keep In Mind
- How might we make this extensible for third-party connectors

t83714:
### What this PR does

Fixes #2715 

This PR is based off PR: https://github.com/magda-io/magda/pull/2765 and make the functionality production-ready.

Change Summary & Deployment Info:
- More streamline & integration between Magda / openfaas deployment.
   - No extra step. No need to change current deployment command.
   - Openfaas core/gateway will be deployed to `[main Magda namespace]-openfaas` namespace (configurable)
     - e.g. if Magda is deployed to `default`, openfaas will be deployed to `default-openfaas` & `default-openfaas-fn`
   - Openfaas functions will be deployed to `[main Magda namespace]-openfaas-fn` namespace (configurable)
    - you can turn off openfaas namespaces auto-creation by set `magda-core.openfaas.createFunctionNamespace` or `magda-core.openfaas.createMainNamespace`. Please note, openfaas namespace requires proper labels set. Simply `kubectl create namespace` won't work
- **Use Helm 3 to deploy it**: The helm chart probably is still helm 2 deployable until PR for [this ticket](https://github.com/magda-io/magda/issues/2778) is landed, but you would have to take multiple steps (as it uses [helm 3's approach](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) to deploy it )
- Unlike helm 2, Helm 3 `--namespace` won't create the namespace if it does not exist. e.g. `helm upgrade --namespace magda` --- you need to create namespace `magda` before run it with helm 3. (openfaas namespaces will be auto created though)
- CRDs (CustomResourceDefinition): Helm3 will install it before all charts are installed. If it exists already, nothing will happen. 
- Openfaas gateway used to be exposed through gateway without auth. Now, by default, only admin users are allowed to access it (e.g. [this API](https://issue-2715-fin.dev.magda.io/api/v0/openfaas/system/functions)).  you can turn off auth by set `global.openfaas.allowAdminOnly`= false
- Scale to zero is turned on by default. the function pod will be scale to zero after 30 mins no usage. You can change this parameter via `magda-core.openfaas.faasIdler.inactivityDuration`="30m" .  **More available openfaas configuration option can be found [here](https://github.com/openfaas/faas-netes/tree/master/chart/openfaas#configuration)**
- helm 3 assume your command is namespaced. i.e. if you run `helm list` it will only list all releases in `default` namespace. if you want to delete the release, you need to specify the namespace (magda deployment namespace). e.g. `helm delete --namespace magda magda` openfaas namespaces will be auto-removed as well.
- ckan-connector: all JSON schema & js connector templates are asynchronously loaded at the same time (see [here](https://github.com/magda-io/magda-ckan-connector/blob/master/src/loadAspectBuilderFiles.ts)). This saves a few seconds of start-up time.  We still have some codebase (e.g. JsonConnector related code) are not `pure` and just import those libs (without running anything) might take up to 2 seconds --- I guess we can optimise it later (or migrate to node12 --- [seems some work has been done in this area](https://medium.com/@nodejs/introducing-node-js-12-76c41a1b3f3f) ). 

#### Test Site:

https://issue-2715-fin.dev.magda.io/

You need to login as admin in order to access the function.
Openfaas API is also protected by the auth.

### Checklist

-   [x] unit tests are included


deploy/helm/openfaas/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+description: Enable Kubernetes as a backend for OpenFaaS (Functions as a Service)
```
AlexGilleran, line 2: Why did we need to copy-paste this in? Probably would be good to put the reason in here, or maybe in a readme.md file in this directory?

magda-gateway/src/createOpenfaasGatewayProxy.ts
```diff
@@ -0,0 +1,32 @@
...
+    jwtSecret: string;
+    authenticator: Authenticator;
+    tenantMode: TenantMode;
+    allowAdminOnly?: boolean;
+}
+
+export default function createOpenfaasGatewayProxy(
+    options: OptionsType
+): express.Router {
+    const router = express.Router();
```
AlexGilleran, line 19: Could use a test for this - in particular to make sure that if anyone ever breaks this so that the `mustBeAdmin` thing stops working, it'll cause a build failure!

magda-web-client/src/Components/Dataset/Add/DatasetLinkItem.tsx
```diff
@@ -255,16 +260,117 @@ const DatasetLinkItemEditing = (props: EditViewProps) => {
     );
 };
 
+async function getAllDataUrlProcessorsFromOpenfaasGateway() {
```
AlexGilleran, line 263: Could use a JSDoc here - what's a data processor? How does one get it out of an open faas gateway?

magda-web-client/src/helpers/promiseAny.ts
```diff
@@ -0,0 +1,41 @@
...
+                }
+                isResultReturned = true;
+                resolve(r);
+            }).catch(e => {
+                if (isResultReturned) {
+                    return;
+                }
+                result[idx].isCompleted = true;
+                result[idx].error = e;
+                if (result.findIndex(item => !item.isCompleted) === -1) {
```
AlexGilleran, line 33: Could this be `!result.some(item => item.isCompleted)` ?

magda-web-client/src/helpers/promiseAny.ts
```diff
@@ -0,0 +1,41 @@
...
+                    isResultReturned = true;
+                    reject(result.map(item => item.error).filter(item => item));
+                    return;
```
AlexGilleran, line 36: Not sure you need this return, it's the last line of the function?

magda-web-client/src/helpers/promiseAny.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Take an array of promises and return a new promise that will be resolved
+ * when any of promises is resolved and will be rejected when all promised are rejected.
+ *
+ * @export
+ * @param {Promise<any>[]} items
+ * @returns {Promise<any>}
+ */
+export default function promiseAny(items: Promise<any>[]): Promise<any> {
```
AlexGilleran, line 9: Not crazy about this name... how do you feel about something like `raceSuccessOnly`, seeing how it shares some of its semantics with `promise.race` ?

deploy/helm/openfaas/Chart.yaml
t83714: Mainly for:
- embedding it as part of Magda. e.g. control the namespace it's deployed into
- make it helm 3 compatible.
Will update the readme.md

magda-gateway/src/createOpenfaasGatewayProxy.ts
t83714: will add a test case.

magda-web-client/src/Components/Dataset/Add/DatasetLinkItem.tsx
t83714: Will add comments --- probably includes a url to openfaas gateway API swagger file

magda-web-client/src/helpers/promiseAny.ts
t83714: removed

magda-web-client/src/helpers/promiseAny.ts
t83714: replace with `!result.some(item => !item.isCompleted)`

magda-web-client/src/helpers/promiseAny.ts
t83714: I actually try to replicate the functionality of this coming promise API (Promise.any):
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any
but no preference on any particular names --- let me know if you prefer a `raceSuccessOnly` or we just leave as it is

deploy/helm/openfaas/Chart.yaml
t83714: Updated readme.md

magda-gateway/src/createOpenfaasGatewayProxy.ts
t83714: Added test cases

magda-web-client/src/Components/Dataset/Add/DatasetLinkItem.tsx
t83714: Added comments

t83714:
@AlexGilleran I think it's ready to have another look

magda-web-client/src/helpers/promiseAny.ts
AlexGilleran: Ooh ok fair enough. Maybe mention that in the jsdoc then?

t83714: 👍 That's a good point~ Have added more comments in the jsdoc

t83714:
@AlexGilleran 
Have added more comment to promiseAll.
Please have another look and let me know if anything else requires changes

magda-web-client/src/helpers/promiseAny.ts
AlexGilleran: :+1: I'll put that link the javadoc in my spelling changes commit :)

AlexGilleran:
@t83714 OK so had a test - architecturally it's great! Testing through the API showed some problems though:

- https://data.gov.au/data/dataset/indigenous-land-use-agreements-registered-or-in-notification returned an error - "Not valid CKAN system URL" a number of times
- https://data.gov.au/data/dataset/0b1243c5-2cab-4675-b3f5-6b920daf377b doesn't work because it doesn't have any distributions (which makes sense), but it needs to tell the user this rather than just saying that it's invalid.

t83714:
> * https://data.gov.au/data/dataset/indigenous-land-use-agreements-registered-or-in-notification

@AlexGilleran 
Fixed the bug and adjust the error message handling.
The demo site is also updated.
Please have a play when you have time

PR #2804, opened 2020-04-15
Added support for microk8s

AlexGilleran:
### What this PR does

This updates the docs and create secrets script to make running in Canonical's MicroK8S a possibility - this is the best way I've found to run K8S on Linux so far.

### Checklist


AlexGilleran:
After a day of playing with it I'm finding microk8s has some problems, I think I'll leave this out of master :(

PR #2807, opened 2020-04-20
Implemented Text Querying in NodeJS

### LINKED ISSUE #2355: Rewrite Search API in Node JS - Querying
Rewrite the existing Search API as a Node JS service, with 1-1 API parity such that we can run a copy of the Scala service and the Node service together with the same Kubernetes service.

This issue is part one, which is making a nodejs search api that can do text-based querying but doesn't yet do anything with facets or regions.

AlexGilleran:
### What this PR does
Fixes #2355 

This implements part of the search API in Node JS. What should work currently is:
- Anything related to keyword search
- Getting the correct date range back

What isn't done yet is:
- Anything related to facets/filters
- Regions
- Anything related to Auth

The new API sits alongside the scala one for now - in the test deployment at https://nodejs-search-api.dev.magda.io, I've changed the gateway config so that it points `api/v0/search` to the new API.

Next I'm going to make the nodejs search api have feature-parity with the old one, then I'm going to implement auth on top of that.

### Checklist


deploy/helm/internal-charts/search-api-node/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
t83714, line 1: Probably increase to v2 so we can safely to use helm3 feature / ignore helm2 issues.
Also add `kubeVersion: ">= 1.14.0-0"` will make sure users get proper error message when they deploy Magda to a unsuppported Kube version

deploy/helm/minikube-dev.yml
```diff
@@ -13,9 +13,6 @@ global:
   enablePriorityClass: false
   defaultContactEmail: "magda@mailinator.com"
   enableMultiTenants: false
-  openfaas:
-    # turn off auth over openfass gateway for ease of debugging
-    allowAdminOnly: false
```
t83714, line 18: This values is actually required by gateway (when it forward request to openfaas). 
We also have some other optional setting under `global.openfaas`:
- namespacePrefix: "" # auto determined from the actual Magda-core deployment namespace
- functionNamespace: openfaas-fn  # Default namespace for functions
- mainNamespace: openfaas # Default namespace for gateway and other core modules
Thought this one is more relevant to openfaas. Thus, put here.

deploy/helm/minikube-dev.yml
```diff
@@ -87,9 +84,21 @@ magda-core:
       pullPolicy: IfNotPresent
       imagePullSecret: false
 
+  openfaas:
```
t83714, line 87: Here we probably need to config openfaas different for local testing (previously forgot about it):
```yaml
openfaas:
  faasnetes:
    imagePullPolicy: Always
```
This will make openfaas always re-pull the docker image by default

deploy/helm/minikube-dev.yml
```diff
@@ -87,9 +84,21 @@ magda-core:
...
+    data:
+      resources:
+        limits:
+          cpu: 1000m
+
 tags:
   all: true
+  search-api-node: true
   connectors: false
+  ckan-connector-functions: false
```
t83714, line 101: Probably no need to set it off as it won't actually create a pod?

magda-search-api-node/src/createApiRouter.ts
```diff
@@ -0,0 +1,216 @@
+import * as express from "express";
+import * as _ from "lodash";
+import buildJwt from "@magda/typescript-common/dist/session/buildJwt";
```
t83714, line 3: Probably `magda-typescript-common/src/dist/session/buildJwt` just to consistent with other project plus you don't have to build `typescript-common` first before run this

magda-search-api-node/src/createApiRouter.ts
```diff
@@ -0,0 +1,216 @@
...
+import * as joi from "joi";
```
t83714, line 4: It seems `joi` has been moved to `@hapi/joi`. If it's easier, probably just update the dependency name.
> This module has moved and is now available at @hapi/joi. Please update your dependencies as this version is no longer maintained an may contain bugs and security issues.

magda-search-api-node/src/createApiRouter.ts
```diff
@@ -0,0 +1,216 @@
...
+        options.elasticSearchUrl,
+        options.datasetsIndexId,
+        options.regionsIndexId,
+        options.publishersIndexId
+    );
+
+    installStatusRouter(router);
+
+    const baseValidators = {
+        start: joi.number().default(0),
```
t83714, line 34: Looks like nice & easier than json schema to validate data 😄

project/DockerSetup.scala
```diff
@@ -10,12 +10,12 @@ import sbt.TaskKey
 
 object DockerSetup {
   def setupDocker(stage: TaskKey[File], dockerFileMod: Dockerfile => Dockerfile = identity) = {
-    val nameSpacePrefix = Option(System.getProperty("repository")) match {
+    val nameSpacePrefix = Option(System.getProperty("repository", System.getenv("MAGDA_DOCKER_REPOSITORY"))) match {
```
t83714, line 13: Do we need to support `System.getProperty("repository")` as well (support passing value both ways)? 
We still have `docker-build-local` script needs it:
```
"docker-build-local": "cd .. && sbt \"-Drepository=localhost:5000\" searchApi/dockerBuildAndPush",
```

scripts/create-docker-context-for-node-component.js
```diff
@@ -32,7 +32,8 @@ const argv = yargs
         repository: {
             description:
                 "The repository to use in auto tag generation. Will default to '', i.e. dockerhub unless --local is set. Requires --tag=auto",
-            type: "string"
+            type: "string",
+            default: process.env.MAGDA_DOCKER_REPOSITORY
         },
         name: {
```
t83714, line 38: It'd be nice if we could add env var support to `name` parameter as well.
Currently, most modules have default name `data61/xxxxx`. 
It works for most docker repo.
But for repo like github registry, it cares about pushing repository url pattern and the  `data61/` part make it complaint.

project/DockerSetup.scala
AlexGilleran: Yep - that's what this should do! System.getProperty gets the value passed in via -D, but if it's not present it falls back to the MAGDA_DOCKER_REPOSITORY env var. Is it not working like that for you?

t83714: Ah~ Sorry I missed that line. All good now 👍

AlexGilleran:
Should be ok for another look.

PR #2819, opened 2020-04-30
Issue/2784  Disable edit button on harvested datasets

### LINKED ISSUE #2784: Disable edit button on harvested datasets
## Description
At this point there's no reason to edit a harvested dataset, because it'll get clobbered on the next re-harvest. So we should disable the edit button on those and explain why.

## Acceptance Criteria
- If a dataset has been harvested from an external source, the "Edit" button is still visible for admins, but is disabled. This includes CSV for now
- A tooltip is present explaining why the dataset can't be edited
    - Explain that because the source is outside magda it will create a conflict
    - @alyce to review the actual text once we PR it
    - Use the standard question-mark-in-a-circle tooltip thingy
- Non-harvested datasets continue to display the edit button as usual

t83714:
### What this PR does

Fixes #2784

Disable edit button on harvested datasets

### Checklist


magda-web-client/src/Components/Dataset/DatasetPage.tsx
```diff
@@ -31,6 +33,11 @@ const DatasetPage: FunctionComponent<PropsType> = props => {
 
     const publisherId = dataset?.publisher?.id ? dataset.publisher.id : null;
 
+    const editButtonTooltipText =
+        "The `Edit` function is only available to datasets created by Magdad.";
```
AlexGilleran, line 37: I think it'd make more sense to flip this around to something like "Datasets harvested from outside Magda can't be edited".

magda-web-client/src/Components/Dataset/DatasetPage.tsx
```diff
@@ -31,6 +33,11 @@ const DatasetPage: FunctionComponent<PropsType> = props => {
...
+
+    const isDatasetEditable = dataset.identifier?.indexOf("magda-") === 0;
```
AlexGilleran, line 39: I'm not comfortable with this as the detection mechanism. If we ever change the way that we generate ids for internal datasets (e.g. what if we renamed the product?) it'll break this :(. 

Aren't we storing the system that created the metadata in an aspect somewhere? If so we should start for Magda.

t83714: 😄 was also wondering whether we should do it in a more formal way.
How about we create the `source` aspect for all datasets that are created by Magda.
And check whether a dataset is an internal dataset by looking at this aspect?
In this way, we can safely assume that if a dataset has no `source` aspect, it must NOT be an internal one

magda-web-client/src/Components/Dataset/DatasetPage.tsx
t83714: 👍 will update

t83714:
@AlexGilleran 
I think it's ready to have another look~

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
```diff
@@ -227,6 +229,15 @@ type Access = {
     notes?: string;
 };
 
+function getInternalDatasetSourceAspectData() {
+    return {
+        id: "magda",
```
AlexGilleran, line 234: We probably need to allow slightly better for when a instance of Magda is crawling another instance of Magda - like maybe this should be "magda-internal" or something? And the name "_This_ Magda metadata creation tool"? 🤔

t83714: @AlexGilleran 
Yeah~ I think it's good idea to provide more details on this to make it more future proof.
Do we want to use the `type` field to provide more details on the source?
Otherwise, it looks like:
id: `magda-internal`, type: `internal`
Probably, id: `magda`, type: `authoring tool`?

AlexGilleran: Maybe we pass in an id for the instance like `magda-data-gov-au` or `magda-dawe`, then have `type: internal` still? 🤔 .

I think ideally when another magda ingests this, it should know to copy the `id` but change the `type` to `external` or something.

t83714: If we go for the `magda-data-gov-au` or `magda-dawe` for the `id`, we probably need an extra config option?
For a similar purpose (i.e. able to tell what's the original Magda system create it), will the current `url` field (of source aspect) be a good alternative without adding an extra config option? I mean we probably can set it to `global.externalUrl`?

AlexGilleran: Yeah using `global.externalUrl` sounds like a good idea!

t83714:
@AlexGilleran 
I updated the code to make sure `global.externalUrl` is used in `source` aspect `url` field.
Seems we actually didn't make `global.externalUrl` pass to web-client probably. Thus, updated the `web-server` for this.
Please have another look when have time and let me know if any other changes are required 👍

AlexGilleran:
Looks good @t83714 ! The only remaining thing is what we talked about on slack the other day re the (?) next to the disabled edit button. Can we make it align with what @alyce said?

> What about a plain old text link below it - like 'why can't I edit?'

So I guess rather than the (?) to launch the tooltip, we have a bit of text below it that launches the tooltip.

t83714:
Ah~ Yes forgot about it. Will do 👍

t83714:
@AlexGilleran the tooltip style has been updated (text underneath now).
Please have a look and let me know anything needs to adjust~

AlexGilleran:
@t83714 I fiddled with the text / spacing / font size a bit and changed the tooltip so it's able to open onclick on mobile, let me know if it's ok :).

magda-web-client/src/Components/Common/TooltipWrapper.tsx
```diff
@@ -98,10 +90,16 @@ class TooltipWrapper extends React.Component<Props, State> {
...
+        this.setState({
+            open: true
+        });
+    };
+
     render() {
         const className = this.props.className ? this.props.className : "";
-        const openClass =
-            this.props.startOpen && !this.state.dismissed ? "tooltip-open" : "";
+        console.log(this.state.open);
```
t83714, line 101: Good job to make it work on mobile 👍 
Tested on my iPhone and works well 💯 
only one tiny thing is this `console.log` line, if we don't need it, we probably want it removed?

AlexGilleran: oops :|

PR #2820, opened 2020-05-04
Issue/2777 Allow config gateway cookie options via helm charts

### LINKED ISSUE #2777: Local dev can't authenticate against dev.magda.io in Chrome
Trying to develop locally (`yarn run dev`) in `magda-web-client` against the dev.magda.io API, using very recent versions of Chrome will lead to this error:

```
A cookie associated with a cross-site resource at http://dev.magda.io/ was set without the `SameSite` attribute. It has been blocked, as Chrome now only delivers cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
```

Those links are https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.

Which is fair enough, they're trying to secure the web. We need to _un_secure our dev instance, which means allowing for those attributes to be set.

### Acceptance Criteria
- It's possible to configure Magda in such a way that you can authenticate against it cross-domain, even via the latest version of Chrome.
- This is NOT on by default
- The setting needs to have a comment above it explaining that it should not be turned on except in a very select set of circumstances!!

t83714:
### What this PR does

Fixes #2777 

> my chrome login (local dev) always works even without `sameSite=none`. So this PR is more for future proof and also allow us to adjust all cookie settings

- Make cookie option configurable on the gateway
  - this's done via passing config options to [express-session](https://github.com/expressjs/session#cookie)
- Set [cookie secure](https://github.com/expressjs/session#cookiesecure) option **default** to `auto`. (i.e. auto-set `secure` cookie attribute when HTTPS). Previously, was always `not set`.
  - Chrome actually require BOTH `secure` & `sameSite=none` to enable crosssite request cookie
  - It's more secure and recommended to turn it on if you have HTTPS available. `auto` make it works for both https and http
- Set `sameSite=none` in `magda-dev.yaml`
- Upgrade `express`, `express-session` & `cookie-parser` to make sure [cookie@0.4.0](https://github.com/jshttp/cookie/releases/tag/v0.4.0) is used (which is the first version supports `sameSite` attributes)
- Upgrade `@type/express` & `@types/express-serve-static-core` to get typescipe compiled ok 

### Test Site:

https://issue-2777.dev.magda.io/

Deployed with the following added to preview.yaml gateway config:
```yaml
cors:
  credentials: true
  origin: true
```

image

### Checklist


t83714:
I think it's ready for review now~

magda-search-api-node/src/createApiRouter.ts
```diff
@@ -162,15 +162,15 @@ export default function createApiRouter(options: ApiRouterOptions) {
...
 
             try {
                 const results = await searchQueryer.search(
                     processedQuery,
-                    queryString.start,
-                    queryString.limit,
-                    queryString.facetQuery
+                    queryString.start as any,
+                    queryString.limit as any,
+                    queryString.facetSize as any
```
t83714, line 173: Probably worth to point out this change --- from the argument name it looks like a `facetSize` parameter. Thus, changed it from `facetQuery` to `facetSize`. Correct me if I were wrong

AlexGilleran: 👍

AlexGilleran:
I can't get this to work in the preview @t83714 , not sure what I'm doing wrong.

In `preview.yml` I've tried completely removing the `cookie` part from the config, which results in the cookie that gets set still being set to SameSite: None

image

The `Set-Cookie` header that comes back is:
`connect.sid=s%3Aw13RyhexFC__7K2A2jMDv2Gll2CQr7ra.sMxzsYvC6GBWZHSIeqLdQlkIjlspVf8f1Hh5bLNpOIQ; Path=/; Expires=Thu, 07 May 2020 15:18:23 GMT; HttpOnly; Secure` - so no SameSite. I think we should probably be setting it to `auto` by default?

Trying to force it to use `auto` like so:

```
# (in magda-core/gateway)
    cookie:
      sameSite: "auto"
```

Leads to an error message for "invalid value sameSite" :(.

This is on Firefox on MacOS.

t83714:
> I can't get this to work in the preview @t83714 , not sure what I'm doing wrong.
> 
> In `preview.yml` I've tried completely removing the `cookie` part from the config, which results in the cookie that gets set still being set to SameSite: None
> 
> image
> 
> The `Set-Cookie` header that comes back is:
> `connect.sid=s%3Aw13RyhexFC__7K2A2jMDv2Gll2CQr7ra.sMxzsYvC6GBWZHSIeqLdQlkIjlspVf8f1Hh5bLNpOIQ; Path=/; Expires=Thu, 07 May 2020 15:18:23 GMT; HttpOnly; Secure` - so no SameSite. I think we should probably be setting it to `auto` by default?
> 
> Trying to force it to use `auto` like so:
> 
> ```
> # (in magda-core/gateway)
>     cookie:
>       sameSite: "auto"
> ```
> 
> Leads to an error message for "invalid value sameSite" :(.
> 
> This is on Firefox on MacOS.

I just had a play on the testing site (without changing the deployment) and it seems it works?

I mean the set-cookie header shows no `sameSite` is set:
image

And the cookie view shows blank under the SameSite column:

image

And it's consistent with configmap setting (`cookie: {}`):

image

Probably you need to logout & login (thus, cookie will be deleted and reset) & close the chrome dev tool to show the latest info?

By the way, the cookie object in the configMap will be directly sent to express-session module without any changes. 

So accept value type is consistent with this doc:
https://github.com/expressjs/session#cookiesamesite

AlexGilleran:
Oh right! Sorry, the issue is a bit unclear here - what I was hoping for is that the SameSite property would be set to disable sending the cookie x-domain by default - i.e. not blank by default, but set to "Strict", unless that's going to break something (shouldn't do though, right?).

My bad with setting it to `auto`, I got it confused with the secure cookie value, I'll try a valid one.

AlexGilleran:
Yep specifically setting samesite works good :)

t83714:
When the `sameSite` is not set, browser (e.g. chrome) will treat it as if you set it to `Lax` value. 
It's already quite strict (but still allow top-level navigation see here: https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-8.8.2).
Our session management system might not be ready for the `strict` setting.
Not sure our auth plugins will break or not. Plus, it might not be desired for users that you will be appearing as logout if you visit Magda through a link click  

By the way, just realised google already drop-off this feature due to covid-19:
https://blog.chromium.org/2020/04/temporarily-rolling-back-samesite.html
That's probably why our local dev login started to work again.
Stay on the default value `Lax` (or not set) will leave us on chrome's default behaviour (which Google is working on to make most existing session management system still work well)

AlexGilleran:
I think we should probably be setting it to _something_ while we're here - if we can't do `Strict` then let's do `Lax`?

That way we should get consistent behaviour across browsers, and we cover ourself in case some random browser decides to default it to some random value in the future?

Also because Mozilla say we should :):

> While you could rely on modern browsers to apply SameSite=Lax automatically, you should rather specify it explicitly to clearly communicate your intent which SameSite policy applies to your cookie. This will also improve the experience across browsers as not all of them default to Lax yet.

t83714:
No problem~ Will set to `Lax` 👍

t83714:
@AlexGilleran Set the default setting for sameStie to `lax` now. tested in minikube and seems works:
image
Also, update the building & running document for a few lines --- seems the helm command for windows section was not right
It's ready for another look~

Issue #2718, opened 2020-01-28
UI for "Deploy to Data.gov.au" option in the Add Dataset flow


AlexGilleran:
## Description
We need to add the ability to publish to data.gov.au as an option in the Add Dataset flow.

<img width="654" alt="image" src="https://user-images.githubusercontent.com/900555/73412955-76f55d80-435e-11ea-9937-199ee5e27d0f.png">
https://projects.invisionapp.com/share/KMVQ7DH652J#/screens/402485646

### Acceptance Criteria
- This should record itself a new key in the `publishing` aspect for the dataset
    - In order to keep this generic we should probably call this `publishAsOpenData` or something like that, rather than having something specific to data.gov.au
    - The value should be a map of portals to flags, e.g.
```javascript
{
    "data.gov.au": true
}
```
- This section can be removed based on a feature flag, which is off by default
- If this is set to "Yes" and classification / sensitivity is set to anything other than "Unclassified", this should cause a validation error when the user tries to go to the next page.

### Out of Scope
- Actually exporting anything to anywhere - the scope of this issue only affects recording the preference to publish to data.gov.au
- The process of actually pushing datasets to ckan / data.gov.au will be handled by a minion built as part of a different issue.

t83714:
closed via PR: https://github.com/magda-io/magda/pull/2758

PR #2829, opened 2020-05-12
Minion for publishing to ckan

### LINKED ISSUE #2723: Minion for Publishing to CKAN
## Description
We want to be able to have data maintained on Magda, but actually published to other portals - in particular within the Australian government, we have the use case that data should be managed in the organisational data catalog (Magda), but still be pushed to CKAN on data.gov.au so that file storage, APIification etc can happen. It's possible that we might want to extend this pattern further too (e.g. pushing out to other portals, other magdas, etc).

I think the best way to fit this into the Magda architecture is to build a minion that listens for a flag stating whether a dataset should be released openly or not, and for datasets that are both published and open, it should push them out at that point. It also needs to listen for changes after that so it can keep the data up to date, and delete it if it's deleted or un-opened.

Assuming that we can nominate a stable id for the record inside CKAN, we should be able to do this without storing any other information, although naturally we should be logging what gets pushed out.

## Acceptance Criteria

- This should line up with how  #2718 sets preferences around publishing.
- A new minion is created which listens for new/changed `publishing` aspects, and also for changed to all aspects that are being used to create the CKAN record
- If a new publishing aspect is created stating that a dataset should be open data where previously this value was `false` or non existent, it should be pushed out to the configured `CKAN` portal
- If there are any changes to the dataset, these should be pushed out to CKAN as an update.
- If a publishing aspect is changed such that the value has changed from `true` to `false`, or the record is being deleted, or just the publishing aspect is being deleted, the minion should delete the record in CKAN.
- When an update is pushed out, this is logged inside the minion pod with information about what record is being pushed out.
- This should be a generic solution that works with all CKAN portals if possible (i.e. try not to do anything data.gov.au specific unless absolutely necessary).
- A new aspect should hold the technical details about the status around pushing it to CKAN - where it is, when it was pushed, any error encountered etc. It could be called `ckan-status` or something like that. It should allow for multiple ckan publishing minions to exist at the same time (pushing to different portals)... i.e. have a map with a key for each ckan portal that the minion `PATCH`es
- If there's an error that's not related to the availability of the CKAN portal as a whole (e.g. the format of the dataset is wrong, no permission for that particular dataset), this should be recorded and the minion should move to the next event
- If there's an error that's related to the CKAN portal as a whole (e.g. the portal is down), the minion should keep retrying with a back-off

## Technical Notes
- You can find the way that we map from CKAN to Magda in the CKAN connector, this should be able to be done pretty much as the inverse of that.

## Out of Scope
- UI to make this all happen (see #2718)

## Related
#2718

sajidanower23:
### What this PR does

Fixes #2723 

This is the sister PR of https://github.com/magda-io/magda-minion-ckan-publisher/pull/1

Adds a minion for publishing to CKAN.

### Checklist

-   [x] Remove the "Publish To CKAN" button
-   [x] Rename `ckan-publish` to `ckan-export`
-   [x] Structure aspect to work with multiple CKAN instances


deploy/helm/internal-charts/search-api-node/templates/service.yaml
```diff
@@ -8,7 +8,7 @@ spec:
     port: 80
     targetPort: 80
 {{- if .Values.global.exposeNodePorts }}
-    nodePort: 30122
```
t83714, line 11: why do we want to change this?

deploy/helm/internal-charts/storage-api/Chart.lock
```diff
@@ -1,6 +0,0 @@
-dependencies:
```
t83714, line 1: We shouldn't delete the lock file. And I think your PR doesn't need to touch `internal-charts`

magda-typescript-common/src/CkanClient.ts
```diff
@@ -0,0 +1,364 @@
+import fetch from "isomorphic-fetch";
```
t83714, line 1: Do we need this for main repo?
It's only used by the minion

magda-web-client/src/Components/Dataset/DatasetPage.tsx
```diff
@@ -142,6 +143,20 @@ const DatasetPage: FunctionComponent<PropsType> = props => {
                             </button>
                         </div>
                     ) : null}
```
t83714, line 145: We probably don't need this button as we have a new UI for it see https://github.com/magda-io/magda/issues/2718

sajidanower23: Yep this is for ease of testing. I will remove this before marking the PR as ready :smile:

t83714: 👍

deploy/helm/internal-charts/search-api-node/templates/service.yaml
sajidanower23: It was conflicting with the minion. Changed it back and changed the minion's nodePort instead

magda-typescript-common/src/CkanClient.ts
sajidanower23: Oops. Yeah we don't need this file in the main repo. I thought I would delete it after the minion works, and then promptly forgot about it :smile:

magda-web-client/src/Components/Dataset/DatasetPage.tsx
sajidanower23: Removed

yarn.lock
```diff
@@ -3417,6 +3417,11 @@
     "@types/express-serve-static-core" "*"
     "@types/mime" "*"
 
+"@types/shortid@^0.0.29":
```
t83714, line 3420: Is this `shortid` still used?

magda-typescript-common/package.json
```diff
@@ -58,6 +59,7 @@
     "moment": "^2.17.1",
     "read-pkg-up": "^3.0.0",
     "request": "^2.88.0",
+    "shortid": "^2.2.15",
```
t83714, line 62: probably not used anymore

sajidanower23:
@t83714 @AlexGilleran I made a deployment to https://2723-publish-to-ckan.dev.magda.io
To test, add a dataset, and while doing so, select "Yes" to publishing as open data to DGA.
This will create a dataset in https://demo.ckan.org/dataset.
Edit the dataset and that should also reflect on CKAN.
Edit the dataset and select "No" to publishing to DGA, and that will delete the dataset.

magda-registry-aspects/ckan-export.schema.json
```diff
@@ -0,0 +1,53 @@
+{
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "title": "Ckan Export Information Aspect",
+    "description": "An aspect that stores information regarding exporting the associated dataset to Ckan. Any changes to this aspect will trigger the export action performed by the Ckan export minion.",
+    "type": "object",
+    "properties": {
+        "status": {
```
AlexGilleran, line 7: So how does this work with multiple export CKANs?

t83714:
Since the data exported to DGA ckan will be crawled by DGA magda connector, we should make sure the data fields that are exported should match the current DGA ckan system.

Currently, fields like `temporal_coverage`, `license` etc are not transfered.

We probably should at least to make sure at least currently ckan connector supported fields are covered. i.e. all fields that are processed in https://github.com/magda-io/magda-ckan-connector/blob/master/aspect-templates/dcat-dataset-strings.js and https://github.com/magda-io/magda-ckan-connector/blob/master/aspect-templates/dcat-distribution-strings.js

Please note, expect a few handful main fields (e.g. title description etc), most other fields are handled as `additional fields` (user-defined field) in ckan.

Here is the test example:
Dataset created in Magda:
https://2723-publish-to-ckan.dev.magda.io/dataset/magda-ds-efadd014-a3df-4cfe-a7b5-cbb7062b79c0/details

Data in ckan (via API):
https://demo.ckan.org/api/3/action/package_show?id=250-logo-ckan

Compared with one dataset in the current DGA CKAN system, you can see more fields are returned:
https://data.gov.au/data/api/3/action/package_show?id=asic-credit-licensee

magda-typescript-common/package.json
sajidanower23: Removed

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
```diff
@@ -933,6 +946,7 @@ async function convertStateToDatasetRecord(
             "dataset-distributions": {
                 distributions: distributionRecords.map((d) => d.id)
             },
+            "ckan-export": ckanExport,
```
t83714, line 949: Here we save the whole `ckan-export` aspect using record API may erase the `ckanId` (which is used by minion to identified ckan package created by minion) that is generated from backend (as our record API will not merge the object and only replace the whole ckan-export object). 
As mentioned in slack, we probably want to use patch APIs (https://dev.magda.io/api/v0/apidocs/index.html#api-Registry_Record_Aspects-PatchV0RegistryRecordsRecordidAspectsAspectid or http://192.168.64.3:30100/api/v0/apidocs/index.html#api-Registry_Record_Service-PatchV0RegistryRecordsId) to handle export aspect saving separately.
i.e. Using a separate patch API call to specifically ask registry API to update status fields (and `exportRequired` field) of export aspect without touching other data fields (e.g. `ckanId`) might be generated from the backend.
We probably want to implement the logic in the function  `submitDatasetFromState` (still in DatasetAddCommon.ts) as it will only be called when the dataset is published.

magda-web-client/src/actions/recordActions.ts
```diff
@@ -152,6 +158,41 @@ export function fetchDistributionFromRegistry(id: string): any {
...
 }
 
+const DefaultCkanExportProperties: CkanExportAspectProperties = {
+    status: "withdraw",
+    hasCreated: false,
+    exportRequired: false,
+    exportAttempted: false
+};
+
+async function notifyCkanExportMinion(datasetId: string) {
```
t83714, line 168: The changes in this file were used by the turn on / off button on dataset page.
As we now control export-minion on/off from add dataset workflow, all changes in this file should not be required anymore.

magda-web-client/src/config.ts
```diff
@@ -181,6 +185,18 @@ const vocabularyApiEndpoints =
               "https://vocabs.ands.org.au/repository/api/lda/ands-nc/controlled-vocabulary-for-resource-type-genres/version-1-1/concept.json"
           ];
 
+let defaultCkanServer: string;
+if (process.env.NODE_ENV === "development") {
```
t83714, line 189: Do we want to make it configurable (through helm and web-server/server.config)?
@AlexGilleran

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
sajidanower23: Created #2887 to come back to this. At the moment it fetches the `ckanExport` aspect again to stay current

PR #2847, opened 2020-05-25
Issue/2837 Add magda internal auth provider to allow email/password based login

### LINKED ISSUE #2837: Internal Email/password based authentication
## Description
Currently you have to plug Magda into some other auth provider in order to have user accounts / authentication.

This:
- Makes it hard to setup for new users
    - In particular, it prevents us from having a working admin account from day 1
- Forces a dependency on another service
- Isn't really appropriate for users who can't integrate with an existing trusted SSO service, but also don't want to use something like Google or Facebook.

So it's time to provide our own username (or email) and password authentication.

## Acceptance Criteria
- An administrator with technical skills and access to the DB (i.e. one of us, or someone else who's set magda up) should be able to create a new user and set a password for them. This doesn't need to be via the UI (it can be a command-line tool that just connects direct to the DB for the first pass)
- An administrator (as above) should be able to reset the password (once again, this could be via a CLI tool)
- The password should be properly hashed #️⃣ and salted 🧂 - see OWASP, Troy Hunt guides on doing this properly. 
- This new login method should be available through the UI to the user
- The login in the UI should have a "Forgot your password? Email <admin email address> `mailto:` link

## Out of Scope
- Changing passwords through the UI
- Resetting passwords through the UI
- "I forgot my password" button
- Strong password requirements (initially it'll only be us using it, I trust you not to set the password to "password" unless it doesn't matter)

## Technical Notes
- Ideally this could just be another plugin for passport, along with the data.gov.au one, vanguard one etc etc etc
- Hopefully someone's done something like this before in OSS land and we can just ~steal~ reuse it.
- Keep everything in `magda-core` for now. Building it straight into the gateway is also fine if that's easier, we can pull it out later.

t83714:
### What this PR does

Fixes #2837

Changes Summary:
- Add Magda internal auth provider
  - can be turned off via `gateway.auth.enableInternalAuthProvider`. Default to ON
  - use user email & password to login
- Change default auth strategy name for some auth provider to avoid conflict (e.g. ckan also uses `local-strategy`) 
- Add `set-user-password` script
  - run `yarn set-user-password` (after yarn clean & yarn install) to see the usage
- Frontend adjustment around login page
  - Reuse `global.defaultContactEmail` for password contact email

image

### Test Site:

https://issue-2837.dev.magda.io

To test:
- port-forward the db and create a user record using db admin tool
  - the user record source field must be `internal`
- run `yarn set-user-password` and following the usage info to setup the password

### Checklist


scripts/set-user-password.js
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env node
+const pkg = require("./package.json");
```
AlexGilleran, line 2: Would it be hard to add the ability to create a new user at the same time as setting the password?

AlexGilleran: And maybe make it possible to select user by email address as well? :grimacing:

t83714: Added more parameters to allow adding user:
```
Usage: set-user-password [options]

A tool for setting magda users' password. Version: 0.0.57-0
By Default, a random password will be auto generate if -p or --password option does not present.
The database connection to auth DB is required, the following environment variables will be used to create a connection:
  POSTGRES_HOST: database host; If not available in env var, 'localhost' will be used.
  POSTGRES_DB: database name; If not available in env var, 'auth' will be used.
  POSTGRES_USER: database username; If not available in env var, 'postgres' will be used.
  POSTGRES_PASSWORD: database password; If not available in env var, '' will be used.

Options:
  -V, --version                          output the version number
  -u, --user [User ID or email]          Specify the user id or email of the user whose password will be reset. If -c switch not present, this switch must be used.
  -c, --create [user email]              Create the user record before set the password rather than set password for an existing user. If -u switch not present, this switch must be used.
  -p, --password [password string]       Optional. Specify the password that reset the user account to.
  -n, --displayName [user display name]  Optional, valid when -c is specified. If not present, default display will be same as the email address. Use double quote if the name contains space.
  -a, --isAdmin                          Optional, valid when -c is specified. If present, the user will be created as admin user.
  -h, --help                             output usage information
```

t83714:
@AlexGilleran I think it's ready to have another look when you have time~

magda-gateway/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM node:10
-
 RUN mkdir -p /usr/src/app
 COPY . /usr/src/app
+RUN rm -Rf /usr/src/app/node_modules/bcrypt && cd /usr/src/app/component && npm install bcrypt
```
AlexGilleran, line 4: This could use a comment explanation. I assume it needs linux binaries or something like that?

t83714: Added comments above

PR #2858, opened 2020-06-09
Hide UI placeholders

### LINKED ISSUE #2857: Hide placeholder UI in Add Dataset flow
## Description

There are a number of bits of UI in Add Dataset that might save something to the registry but don't actually have the effect that they imply. We should hide these behind a feature flag.

## Acceptance Criteria

Unless this feature flag is on:
- The "Sharing" section is hidden
  - And authorisation works such that the entire organisation can see a dataset once it's out of draft
- Under "People", anything other than "Which organisation is responsible for publishing this dataset? (*)" is hidden

sajidanower23:
### What this PR does

Fixes #2857 

Adds a feature flag `placeholderWorkflowsOn` for components that are not 100% complete yet.

Note that I have applied the flag at a more granular level than was necessary so that we can easily turn on smaller components if we complete them.

### Checklist

-   [x] ~There are unit tests to verify my changes are correct~ unit tests aren't applicable

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -54,6 +54,7 @@ UI:
 -   Make `add dataset page` save state data to registry (unless in preview mode) rather than localStorage
 -   Change search text to say "Search for data" instead of "Search for open data"
 -   Set flag `previewAddDataset` to `false` by default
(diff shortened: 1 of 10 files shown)


AlexGilleran:
Part of #2863 now

PR #2859, opened 2020-06-09
Added upload to Add Dataset

### LINKED ISSUE #2721: Add upload functionality to add dataset
## Description

This issue connects the UI from #2716 to the backend of #2691, and hence requires both!

Now that we've got a UI to opt-in to uploading files, and an API to allow uploading files, we can put them together. When the user has selected "Yes" for storing files on Magda, they should be uploaded as part of the metadata extraction process.

## Acceptance Criteria

If the radio button for storing files is set to "Yes":
- Dragging and dropping a file causes it to be uploaded to the storage API, concurrently with metadata extraction
- The progress bar for processing the file does not complete until the upload is complete (it should stop at 80% until the upload is complete).
- The `GET` url of the new file should be saved in `dcat-distribution-strings` as a download url
- The bucket and file name should also be saved in a separate aspect
- If the file is subsequently deleted by the user, this should also cause the file to be deleted via the storage API. If this operation fails, an extremely urgent error should be shown to the user instructing them to contact the Magda admin email address in order to resolve it. The user should be restricted from leaving the page until the delete operation is complete.
- The new file should be associated with its distribution as per #2609

If the radio button for storing files is set to "No":
- Nothing should be uploaded when a file is dropped - metadata extraction should proceed as it did before

## Technical Notes
- This will require a multipart file upload.

AlexGilleran:
### What this PR does

Fixes #2721 

This makes it so if desired, files are uploaded to the Magda storage API as well as having their metadata harvested when they're dropped onto the Add Dataset flow.

In order to make this work I've done a few extra things (and one big yak shave):
- Made a default magda-datasets bucket that's created on startup in storage-api
- Made it so that you can upload to an arbitrary path in storage api, within the bucket selected
- Refactored a lot of the metadata extraction to make it faster and lighter. Now everything executes in parallel, and in a webworker, linked by `comlink` back to the main thread
- Upgraded CRA, typescript and prettier in order to do `import type` inside the webworker.

How upload works is that when drag-drop happens, an upload promise is executed in parallel with the metadata extraction webworker. Because we can't monitor upload progress, there's a fake progress thing that goes fast at first and slows down exponentially over time.

### Checklist


AlexGilleran:
Part of #2863 now

PR #2860, opened 2020-06-09
Fix missing cloud SQL proxy

t83714:
### What this PR does

It turns out we never add cloud SQL proxy to `requirements.yaml`.
It worked before because we used to put all charts in `charts` folder and access it locally (rather than build it from `dependencies` ).
After we re-structured our charts, we never deploy an instance using cloud SQL. Thus, only encounter it today 😭 

### Checklist


t83714:
superseded by this PR: https://github.com/magda-io/magda/pull/2861

PR #2861, opened 2020-06-09
Couple fixes found when deploying PMC instance

t83714:
### What this PR does

Fixes:
- Missing cloud SQL proxy in the dependency list ([Previous PR](https://github.com/magda-io/magda/pull/2860) has been closed as superseded by this PR)
- Remove debug config in gateway module
- Make sure only show enabled login option on login screen
- Fixed the rare situation that `express-session` could not initialise session properly

### Test Site:

https://pmc-demo.magda.io/

### Checklist

-   [x] unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.

PR #2855, opened 2020-06-05
Incorrect dates from csv

sajidanower23:
### What this PR does

Fixes #2831 

Reads raw data from spreadsheets so that dates are not interpreted incorrectly.

## Optimization

In current master (ad93d91), to ingest the dataset at https://data.gov.au/dataset/ds-nsw-d3d51a4f-0efb-434d-b42c-90a35280ecb7/details?q=nsw%20covid%2019, it takes about 1 m 37s (on 5 June 2020, it was around 486,000 lines of CSV).

In this branch, it takes about 1 m 30 s.

I changed the order of some loops, made some small tweaks here and there.

### Checklist

-   [x] ~There are unit tests to verify my changes are correct~ Unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.
-   [x] I've linked this PR to an issue in ZenHub (core dev team only)
-   [x] Test that the optimization actually works

PR #2848, opened 2020-05-25
Approval workflow feature flag

sajidanower23:
### What this PR does

Fixes #2835 

Adds a feature flag `datasetApprovalWorkflowOn` to control whether or not approvals are present in the workflow

### Checklist

-   [x] ~There are unit tests to verify my changes are correct~ Unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.
-   [x] I've linked this PR to an issue in ZenHub (core dev team only)
-   [x] Add the flag to deployment charts
-   [x] Remove the "Draft"/"Pending Approval" flag if the approval flow is turned off

PR #2863, opened 2020-06-12
Pmc superbranch

### LINKED ISSUE #2838: Display "My Datasets" on the home screen
## Description
Datasets can be found via search, but right now there's not a very easy way to simply go back to dataset that you've already made - especially given that there's likely to be a delay between adding a dataset to the registry and it being ingested in the search index.

In particular, this makes a workflow where datasets are regularly updated quite difficult, because in order to find the dataset and edit it the user needs to search for it.

We already have designs for the "logged in homepage" with a bunch of functionality, so I think we should do the most stripped down version of that we possibly can while still supporting the regular update workflow - i.e. a list of that user's datasets, ordered by most recent update, descending.

## Screenshots
image
https://projects.invisionapp.com/share/GMXFFVZ5Y8X#/screens

## Acceptance Criteria
- If logged in, the bottom of https://projects.invisionapp.com/share/GMXFFVZ5Y8X#/screens  (see screenshot) is added at the bottom of the home page (i.e. no news stories)
- If logged out, it still looks the same
- Only the "My Data" and "Sign Out" buttons on the sidebar are visible
- My data shows both "Drafts" and "Published" tab - datasets that the user hasn't published (i.e. they're only visible to the user) live on the Drafts tab, everything else lives in "Published". All these datasets should have the user as an owner
- Show the date instead of the status
- Datasets should be sorted by most recently updated to least recently updated (unless this is really hard - if it's super impossible we can do it on the front end for now).
- Don't include the search bar (yet, we'll do it later)
- The "Edit" icon next to each dataset should open it for editing
- Put a "Add Dataset" link on the sidebar under "My data"

## Technical Notes
- We will need to make it possible through the registry to:
    - Query by both user and publishing status at the same time
    - Order by last date modified
    - If at all possible, we should try to make this a generic registry API feature (i.e. it's not just that you can query by user/publishing status but by other things too, similar to how the current aspectQuery allows you to query any path)
- Ordering by last date modified might be very very difficult because I think we only record that as the date of the last event.

### LINKED ISSUE #2843: COVID-19
We need to put together a proof of concept for an organisation that's collecting data from many sources, currently via email.

What we know about their workflow:
- They get new data files every day - mainly in a single email inbox, sometimes in private inboxes.
- These are usually updates to existing datasets - so they'll either get a new version of an existing file, or an entirely new file (probably more often the latter)
- These files are a range of formats, but are rarely nice clean data - they're more likely to be multi-tab excel, PDFs with tables in them etc.

Our mission is to make Magda work as well as we can for this workflow, and put it all up at a test site for them.

sajidanower23:
### What this PR does

Fixes #2843 

This branch fixes #2848 , fixes #2855 , fixes #2858 , fixes #2859, fixes #2838 and fixes #2861 together with ad-hoc fixes.

### Checklist


magda-gateway/src/createBaseProxy.ts
```diff
@@ -69,17 +69,20 @@ export default function createBaseProxy(tenantMode: TenantMode): httpProxy {
...
-            req.method === "GET" &&
-            !proxyRes.headers["Cache-Control"] &&
-            !proxyRes.headers["cache-control"] &&
-            !req.headers["Cache-Control"] &&
-            !req.headers["cache-control"]
-        ) {
-            proxyRes.headers["Cache-Control"] = "public, max-age=60";
-        }
+
+        // FIXME FIXME FIXME TODO: Add this back and make it configurable.
```
AlexGilleran, line 75: Better fix this first!

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsServiceRO.scala
```diff
@@ -126,6 +177,32 @@ class RecordsServiceRO(
         value =
           "Filter the records returned by a value within the aspect JSON. Expressed as 'aspectId.path.to.field:value', url encoded. NOTE: This is an early stage API and may change greatly in the future"
       ),
+      new ApiImplicitParam(
+        name = "aspectOrQuery",
```
AlexGilleran, line 181: @t83714 should this be aspectQuery?

AlexGilleran: Oh wait I get it nvm!

.gitlab-ci.yml
```diff
@@ -27,58 +27,52 @@ builders-and-yarn:
   stage: builders
   image: registry.gitlab.com/magda-data/magda/data61/magda-builder-docker:master
   retry: 1
+  before_script:
+    - yarn install --no-progress --cache-folder .yarn
   cache:
-    key: $CI_JOB_NAME-$CACHE_VERSION
+    key: JS-$JS_CACHE_VERSION
```
t83714, line 33: cache is shared between pipelines and jobs at the project level by default. So `JS-$JS_CACHE_VERSION` key make it always shares cache across stages and between pipeline (until we change the cache version).
This could be an early stage job cache (e.g. build:yarn ) replace a later stage job (e.g. : dockerize:typescript-apis) and may cause issue when we have more than one branch pipeline running?
how about we still use cache to `speed up runs of a given job in subsequent pipelines` (as it's suggested [here](https://docs.gitlab.com/ee/ci/caching/)), i.e. cache vertically --- across pipeline but do not share cache between jobs.
So here, probably a key `$CI_JOB_NAME-$JS_CACHE_VERSION` is what we want?

And `artifcates` use to pass intermediate build results between stages.

.gitlab-ci.yml
```diff
@@ -27,58 +27,52 @@ builders-and-yarn:
...
-  stage: sbt-prebuild
+check-scala-formatting:
+  stage: prebuild
   image: registry.gitlab.com/magda-data/magda/data61/magda-builder-scala:$CI_COMMIT_REF_SLUG
   cache:
-    key: $CI_JOB_NAME-$CACHE_VERSION
+    key: SCALA-$SCALA_CACHE_VERSION
     paths:
       - "$CI_PROJECT_DIR/sbt-cache"
+      - "$CI_PROJECT_DIR/target"
```
t83714, line 53: Probably we still want to use `artifacts`  to pass intermediate build results between stages / jobs rather than cache (as suggested [here](https://docs.gitlab.com/ee/ci/caching/))?
If we passing scala `target` folder which contains build result, we might having issue that different commit share different build result between jobs in the same pipeline.
e.g. job step 2 may build on build result from (older / other commit) upstream job step1

.gitlab-ci.yml
```diff
@@ -88,41 +82,30 @@ buildtest:search-with-index-cache:
...
     - apk update
     - apk upgrade
     - apk add python3 python3-dev py3-pip build-base openssl-dev libffi-dev
     - pip3 install --upgrade pip
     - pip3 install docker-compose
     - docker-compose --version
-  cache:
-    key: $CI_JOB_NAME-$CACHE_VERSION
-    paths:
-      - "$CI_PROJECT_DIR/pip-cache"
```
t83714, line 109: Any reason we want to remove the pip-cache?
We still can share cache across pipeline only for this step/job (and this cache hardly changes)

.gitlab-ci.yml
```diff
@@ -149,14 +132,7 @@ buildtest:search-no-index-cache:
   stage: buildtest
   image: registry.gitlab.com/magda-data/magda/data61/magda-builder-scala:$CI_COMMIT_REF_SLUG
   retry: 2
-  dependencies:
-    - builders-and-yarn
-    - sbt-prebuild
```
t83714, line 154: We probably still want to pass build result via `artifacts`? Or I missed anything?

magda-gateway/src/createBaseProxy.ts
t83714: I thought the disabling cache from client side was working.
Do we have a reproducible case?

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsServiceRO.scala
```diff
@@ -317,7 +420,72 @@ class RecordsServiceRO(
...
+    *   - `:>=` greater than or equal to
+    *   - `:<`  less than
+    *   - `:<=` less than or equal to
+    *
+    *   Example URL with aspectQuery `dcat-dataset-strings.title:?%rating%` (Search keyword `rating` in `dcat-dataset-strings` aspect `title` field)
+    *
+    *   `/v0/records?limit=100&optionalAspect=source&aspect=dcat-dataset-strings&aspectQuery=dcat-dataset-strings.title:?%2525rating%2525`
+    *
+    *   Please note: when both `aspectQuery` and `aspectOrQuery` present:
+    *   - if there is only ONE `aspectOrQuery` specified, query conditions generated from `aspectQuery` and `aspectOrQuery` will be joined with `OR`
```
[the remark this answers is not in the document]
t83714: I thought people would expect `aspectOrQuery` always do some `OR` thing.
But I have to admit even myself felt is a weird behaviour recently.
I have updated already on my branch [pmc-superbranch-fixes](https://github.com/magda-io/magda/blob/b6c1dd99c4db649f4af279b3df78772f031d87b4/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsServiceRO.scala#L469)

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/SqlHelper.scala
```diff
@@ -252,4 +260,13 @@ object SqlHelper {
...
   }
+
+  /**
+    * safely escape an identifier string (e.g. column, table names)
+    * sqls"" (SQLSyntax object) will always treats external input values as binding parameters. Thus, not suitable for identifier
+    * @param name identifier string
+    * @return escaped identifier as SQLSyntax object
+    */
+  def escapeIdentifier(name: String): SQLSyntax =
+    SQLSyntax.createUnsafely('"' + name.replaceAll("\"", "\"\"") + '"')
```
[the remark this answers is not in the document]
t83714: I think there are two risks we might want to look at:
- 1: SQL injection risk (this function is for prevent this)
- 2: we shouldn't allow user to select `any` column they want
The function is used for processing query reference like aspect0, aspect1 etc. 
without this function we have to put like:
SQLSyntax.createUnsafely(ref) like [here](https://github.com/magda-io/magda/blob/ad93d911def22470f9bf1552089877793170b741/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala#L1441)

So `escapeIdentifier` here  just add some extra protection (for protecting SQL injection risk) otherwise we have to call `SQLSyntax.createUnsafely` which is not sql injection safe.

We only have two kind of references  is passed to `escapeIdentifier `:
- ColumnNamePrefixType.PREFIX_TEMP --- before is passed to `SQLSyntax.createUnsafely` directly
- aspectSelectors --- generated by [aspectIdsToSelectClauses](https://github.com/magda-io/magda/blob/ad93d911def22470f9bf1552089877793170b741/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordPersistence.scala#L1711). Its logic seems guarantee only token like `aspect0`, `aspect1` will be produced.

What's your thoughts on this?

magda-web-client/src/AppContainer.js
```diff
@@ -48,7 +50,15 @@ class AppContainer extends React.Component {
                         <Routes />
                     </div>
 
-                    <Footer />
+                    <Switch>
+                        {/** turn off top margin for home page only */}
+                        <Route
+                            exact
+                            path="/"
+                            render={() => <Footer noTopMergin={true} />}
```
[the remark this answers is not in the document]
t83714: Good catch 👍 updated~

magda-web-client/src/Components/Account/AccountLoginPage.tsx
```diff
@@ -146,31 +146,35 @@ export default function Login(props) {
...
-                    <a onClick={() => setLoginFormProvider("ckan")}>
-                        <img
-                            src={ckanLogo}
-                            className="login__logo"
-                            alt="logo"
-                        />
-                        Data.gov.au / Ckan
-                    </a>
-                </li>
+                {providers.indexOf("internal") === -1 ? null : (
```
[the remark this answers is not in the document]
t83714: Updated~
I think just get used to explicitly return `null` to indicate rendering nothing rather `false` 😄 
Back to old days I think we had an old bug that react not smartly recognise all those `empty` value as rendering nothing

magda-web-client/src/config.ts
```diff
@@ -156,7 +159,7 @@ const baseExternalUrl = serverConfig.baseExternalUrl
     ? window.location.protocol + "//" + window.location.host + "/"
     : baseUrl;
 
-const fetchOptions: RequestInit =
+const credentialsFetchOptions: RequestInit =
```
t83714, line 162: Fetch option also allows us to modify other things e.g. `headers`:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch
Just in case we want to globally add custom header or something. Probably should call this a more generic name?

magda-web-client/src/api-clients/RegistryApis.ts
```diff
@@ -181,6 +237,176 @@ export async function fetchRecord(
...
+    if (start) {
+        parameters.push(`start=${encodeURIComponent(start)}`);
+    }
+
+    if (limit) {
+        parameters.push(`limit=${encodeURIComponent(limit)}`);
+    }
+
+    if (orderBy) {
+        if (typeof orderBy === "string") {
```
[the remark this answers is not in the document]
t83714: Updated~
It was passing parameters via parameter list rather option objects --- and I was trying to reduce the number of parameters that you might have to list when call it --- I forgot to refactor it after change the parameter passing method.
Yeah, after change to option object, it doesn't make sense anymore 👍

magda-gateway/src/createBaseProxy.ts
AlexGilleran: It does for the bits we put it in, but I didn't want to discover a place where we'd missed it once we'd already got the demo out. The only time I noticed this being a problem was when navigating back to the home page from Add Dataset, I don't think it knows to not cache then.

Either way I think this should be configurable though.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/SqlHelper.scala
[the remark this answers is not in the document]
AlexGilleran: > SQL injection risk (this function is for prevent this)

This is what I'm most worried about - because although we're replacing characters here and that _should_ prevent SQL Injection, we're replacing a very very hard to bypass method of mitigating it (parameters) with something (perhaps imho) much easier to bypass... e.g. what if postgres adds another escape character, or we missed something, or the implementation of `replaceAll` changes.

If we're only passing in PREFIX_TEMP and aspectSelectors then that's good, but it's still really easy for someone to come along later and pass user input to it. Can we create a special type for those and make it so that this function only accepts that type? And maybe make it so that rather than making identifiers with escape characters safe, it just throws an exception (because I don't think we really need to support escape characters do we?).

.gitlab-ci.yml
AlexGilleran: @t83714 yeah I think I'll revert this whole thing. It was initially to try to make the kubernetes build work better but it actually slows down things on the build box.

magda-web-client/src/config.ts
AlexGilleran: @t83714 yeah but I want to make it really obvious that by using these options you're making your `fetch` pass credentials! Otherwise it's too easy to think "oh a bunch of other `fetch`es use this, so my one that doesn't need credentials should too".

If we have other fetch options that do other things we should probably make them available here under other names and merge them together if a `fetch` needs more than one.

magda-gateway/src/createBaseProxy.ts
AlexGilleran: (which is something I'll do :+1: )

t83714: Fetch API seems support a `cache` option (not the cache header):
https://developer.mozilla.org/en-US/docs/Web/API/Request/cache
the `no-cache` mode seems could 100% prevent client read data from the cache.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/SqlHelper.scala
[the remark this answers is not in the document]
t83714: Yeah~ I don't think we have used any special chars in identifiers.
How about we remove this `escapeIdentifier ` function and revert it back to the `SQLSyntax.createUnsafely`?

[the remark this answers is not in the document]
AlexGilleran: Yeah if you like :)

magda-web-client/src/Components/Dataset/Add/DatasetAddMetadataPage.tsx
```diff
@@ -333,11 +356,22 @@ class NewDataset extends React.Component<Props, State> {
...
 
-            this.setState(state => ({
+            this.setState((state) => ({
                 ...state,
                 isPublishing: true
             }));
 
+            // Setting datasets as approved if
+            // approval flow is turned off
+            if (!config.featureFlags.datasetApprovalWorkflowOn) {
```
t83714, line 366: @AlexGilleran Currently, when `datasetApprovalWorkflowOn` is off, the submitted dataset is still saved with `draft` status. Because of, this dataset will not be shown under either `draft` or `published` dataset tab (as it has no `dataset-draft` aspect but status != published). People would think the dataset is disappeared. Do you mind if I change this to always set the dataset to `published`?
Alternatively, we can change the `draft` dataset tab query and make `dataset-draft` aspect and optionalAspect?
The way of updating publishing status may also have a timing issue:
The `this.setState` may not be completed immediately before  `await submitDatasetFromState`. I probably rewrite this part as well.

AlexGilleran: Sure!

PR #2864, opened 2020-06-15
MyDataset section: Page Count & Fixes

t83714:
### What this PR does

**Targeting `pmc-superbranch`**

Fixes:
- Add record count info to UI (tab & pagination area)
- Adjust `aspectOrQuery` behaviours: will always join with `aspectQuery` with `AND`
  - e.g. For the following `aspectQuery` & `aspectOrQuery` queries are specified:
    - `aspectQuery`:
     - q1
     - q2
    - `aspectOrQuery`:
     - q3
  - The generated query conditions will be `((q1 AND q2) AND q3 )` 
- Use `start` & `limit` for pagination (instead of `pageToken` as we orderBy last update time)
  - Before this fix, my dataset tab might not show all records start from page 2.  
- Add a new `orderNullFirst` parameter to control whether nulls appear before (`true`) or after (`false`) non-null values in the sort ordering.  
- Update API docs

@AlexGilleran 

~All API changes should be backwards compatible (with `master`) and pass all existing test cases.~
~If you think the fixes are important (e.g. for PMC), you can merge it now.~
I will keep working on test cases for new features --- it might take some times~

Test cases added

### Checklist

-   [x] Test cases include


t83714:
@AlexGilleran Test cases added --- I think it's ready for another look

AlexGilleran:
@t83714 one niggle, fix it if it's easy otherwise make another issue - the keyword search is case sensitive (e.g. if I create a dataset called "Dataset", searching for "dataset" gets no results). Can we make it case insensitive?

t83714:
the `:?` was using `ILIKE` which is case sensitive but I changed to `SIMILAR TO` later in order to support regular expression with one operator. 
Have to say postgresql is a very powerful DB but its default config / setting is not useful --- it turns out `SIMILAR TO` is case sensitive and no way to change it.
Thus, I changed `:?` back to support pattern only (case insensitive using `ILIKE`) and added `:~` to support POSIX regular expression (which could be more useful).
Test case added for the new operators.

PR #2890, opened 2020-07-02
Protected or above security classification is an error

### LINKED ISSUE #2888: Add/Edit Dataset - A Dataset with PROTECTED or above clearance should not be allowed to be saved
## Description

Currently if a user puts in a dataset with a classification of protected or above, we warn them that they shouldn't add it to Magda, but we don't actually _enforce_ this warning by preventing them from going any further.

We should make it so that this works like a validation failure (e.g. the same as them not setting a description) - when the user tries to go to the next page it should scroll them back up and explain that they need to change the classification.

## Acceptance Criteria

-   If the user puts in a classification level of PROTECTED or higher:
    - A warning shows, as it does now
    - When trying to go to the next page:
        - An additional error message appears, explaining that they cannot proceed with this classification level as Magda can't be used to store that level of data
        - The user's viewport is scrolled up to see this message, as with other validation failures
        - The user cannot proceed to the next page until they've changed to an acceptable classification
- If the user puts in a classification level below PROTECTED, everything proceeds as usual

sajidanower23:
### What this PR does

Fixes #2888 

If dataset security clearance is PROTECTED or above, it doesn't let the user proceed and sets an error.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -66,6 +66,7 @@ UI:
 -   Fix magda & ckan login options are missing
 -   Always set dataset publishing state to "published" no matter the approval flow is on / off
 -   Fix a sheet file contains date would cause blank screen during processing
+-   Set an error if security classification is at PROTECTED or above
 
 Storage:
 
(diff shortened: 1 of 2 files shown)


sajidanower23:
@AlexGilleran @t83714 I have made a deployment to https://2888-protected-classification.dev.magda.io.
To test:
1. Add/Edit a dataset
1. Click through to the `Access and User` page
1. Try setting the security clearance to `PROTECTED` or above
1. Notice the error
1. Try going to the next page anyway
1. Notice that the error will persist

PR #2903, opened 2020-07-17
Issue/2887 Fixed ckan-export aspect's fields should be only changed via JSON patch from the frontend

### LINKED ISSUE #2887: ckan-export aspect's fields should be only changed via JSON patch from the frontend
**Describe the bug**

`ckan-export` aspect's fields should be only changed via JSON patch from the frontend.

i.e. using the patch API here: https://dev.magda.io/api/v0/apidocs/index.html#api-Registry_Record_Service-PatchV0RegistryRecordsId

**Reason:** 

`ckan-export` aspect includes some critical state fields (e.g. `ckanId`) that should ONLY be updated by the backend (minion).

However, the [record modify API](https://dev.magda.io/api/v0/apidocs/index.html#api-Registry_Record_Service-PutV0RegistryRecordsId) we currently used at the frontend has no capability of updating partial fields without touching the rest of the fields for an aspect. i.e. it only works as data object replace rather than merge.

This will lead to a racing condition that a critical backend updated field (e.g. `ckanId`) could be deleted while the frontend tries to update the `ckan-export` aspect with an object data without `ckanId`. 

### Acceptance Criteria
- The add/edit dataset workflow should only handle updating `ckan-export` aspect separately in function `submitDatasetFromState` (in `magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts`)
  - i.e. we should move any `ckan-export` aspect related logic out of function `createDatasetFromState` and `createDatasetFromState`
- `ckan-export` aspect should be updated using [Record Patch API](https://dev.magda.io/api/v0/apidocs/index.html#api-Registry_Record_Service-PatchV0RegistryRecordsId)
- the action of updating `ckan-export` aspect should only be taken place AFTER any other dataset changes have been done.

t83714:
### What this PR does

Based on this PR: https://github.com/magda-io/magda/pull/2898

Should merge that one first. After merge, refresh this PR will see less file changes (should see 5 file changes only).

Fixes #2887
- ckan-export aspect's fields should be only changed via JSON patch from the frontend
- escaped the special chars in json path according to RFC
- Change schema of `ckan-export`. Set only "status" & "exportRequired" as compulsory fields.
  - avoid complicate logic or json schema error
  - "hasCreated" & "exportAttempted" are only be set by minion
  - minion code logic is safe with ckan-export without `hasCreated` or `exportAttempted ` 
  - Due to the schema chanegs, when deploy to the existing site, you should update the schema accordingly

### Test

you can use this API to verfiy aspect data easily:

/api/v0/registry/records/[datasetId]/aspects/ckan-export

### Checklist


magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
```diff
@@ -1234,20 +1230,59 @@ async function convertStateToDistributionRecords(
     return distributionRecords;
 }
 
+/**
+ * Update ckan export aspect status acccording to UI status.
+ * We use JSON Patch request to avoid edging cases
```
AlexGilleran, line 1235: Might need to mention what the edge case is

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
```diff
@@ -1234,20 +1230,59 @@ async function convertStateToDistributionRecords(
...
+ *
+ * @param {string} datasetId
+ * @param {State} state
+ */
+async function updateCkanExportStatus(datasetId: string, state: State) {
+    /**
```
AlexGilleran, line 1241: This isn't really a jsdoc comment, it should be `/* */`.

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
```diff
@@ -1234,20 +1230,59 @@ async function convertStateToDistributionRecords(
...
+     * Please note: any frontend code should ONLY (and should ONLY NEED to) change those three fields:
```
AlexGilleran, line 1242: Two fields or three fields?

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
```diff
@@ -1234,20 +1230,59 @@ async function convertStateToDistributionRecords(
...
+     * - status
+     * - exportRequired
+     * Those two fields should ONLY been updated via patch request. Otherwise, other runtime fields may lose value.
+     *
+     * Other fields are all runtime status fields that are only allowed to be altered by minions.
+     * Any attempts to update those from frontend will more or less create edging cases.
```
AlexGilleran, line 1248: Once again, it'd be good to say what the edge case is.

magda-web-client/src/Components/Common/ErrorMessageBox.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * optional; auto move viewport to this error message box; default to false
+     *
+     * @type {boolean}
+     */
+    scrollIntoView?: boolean;
+};
+
+const ErrorMessageBox: FunctionComponent<PropsType> = (props) => {
```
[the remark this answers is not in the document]
t83714: It's now being used by a lot of other places (e.g. modals etc) and feels it's strange to pull it from `/add` folder.
Its functionality is quite generic e.g. you can set error either by props or redux.
You also can overwrite the error message using a customise Error (and getErrorContent)
I think we probably can gradually move all code to use this ErrorMessageBox (probably not now) to avoid some repetition job.

magda-web-client/src/Components/Common/ErrorMessageBox.tsx
```diff
@@ -0,0 +1,183 @@
+import React, { FunctionComponent, useRef, useEffect } from "react";
+import { connect } from "react-redux";
+import { config } from "config";
+//import ServerError from "";
```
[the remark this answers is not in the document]
t83714: Removed~

magda-web-client/src/Components/Dataset/Add/DatasetLinkItem.scss
```diff
@@ -2,241 +2,368 @@
...
+                line-height: 1.54;
+                letter-spacing: normal;
+                color: #484848;
+
+                text-overflow: ellipsis;
+                overflow-wrap: break-word;
+                white-space: nowrap;
+                overflow: hidden;
+            }
+            /*div:last-child {
```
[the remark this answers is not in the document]
t83714: removed

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
t83714: Add more comments re: edge case

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
t83714: good catch --- should be two fields

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
t83714: Added comment above

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
t83714: You mean you want the `*` to be removed from 
```
```

AlexGilleran: So a tsdoc or jsdoc is supposed to be documentation for the variable or function directly under it, like in that example. This looks like it's supposed to just be a general comment for someone reading the code (like what you'd put `//` in front of if it was on one line), rather than variable documentation for `exportDataPointer`? Or am I missing something?

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
```diff
@@ -1234,20 +1230,62 @@ async function convertStateToDistributionRecords(
...
+    )}`;
+
+    const uiStatus = state?.datasetPublishing?.publishAsOpenData?.dga
+        ? true
+        : false;
+
+    await ensureAspectExists("ckan-export");
+
+    /**
+     * Here is a trick based my tests on the [JsonPatch implementation](https://github.com/gnieh/diffson) used in our scala code base:
```
AlexGilleran, line 1265: This looks like it should be a normal comment rather than a tsdoc too?

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
t83714: The docs software (tsdoc) won't pick up those comments as they are in the function body.
But I think those comment might make people think the comment is for `exportDataPointer` due to the comment style, as you said --- which will cause confusion.
It's now updated to use `//` now

magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts
t83714: Updated to `//` now

PR #2924, opened 2020-08-10
Issue/2766 Allow supplying custom OPA policy

### LINKED ISSUE #2766: Spike: Discover a simple way of adding / replacing OPA policy when deploy Magda
We probably need a way of changing OPA policy when deploying a Magda site.

Currently we manage OPA policy as part of the OPA helm chart. It's quite simple for us to change the OPA policy, but not easier for other projects based on Magda to change OPA policy.

There are options to implement more complicated policy management options  (e.g. policy bundle or REST API).

But simpler than that would be to just make it so that policies can be passed in through Helm.

### What we want to try to achieve
- A new installation of Magda still has a handful of default policies (e.g. public, based on org unit)
- These default ones can be overridden and modified
- New ones can be added
- This whole process can be done without forking the magda-core repo, rebuilding etc. Ideal would be just changing helm config. But is this possible.

### To mark as done
- A working demo of a new way of modifying the policies, and a justification of why this is the best method

## Past discussions
How should we do that?

One possibility:
- Added a switch `opa.useDefaultPolicies` so that `opa-policies` configMap is only created when `opa.useDefaultPolicies` is `true`

If a user wants to deploy with his own policy set, they can:
- create a new deploy chart e.g. named `new-magda`
- Add `magda-core` (or `magda`) official helm chart as dependencies
- Set `magda.opa.useDefaultPolicies` = `false`
- Add a configMap named `opa-policies` to `new-magda`'s templates folders

Potentially we could also provide the default set of policies as a YAML map and enable them to be overridden in helm config?

t83714:
### What this PR does

This PR reply on this init container docker image to decode dir structure from configMap: [magda-configmap-dir-loader](https://github.com/magda-io/magda-configmap-dir-loader) 

Fixes #2766 
- Allow supplying one (or more) configMaps that contains users' own policy files
  - This also fixes the issue of `1mb configMap limit` issue

### How it works
- Due to the limit of allowed chars for configMap key, we need to encode a file directory structure into JSON before encode into configMap (Re: the encoding structure, please see [test cases in magda-configmap-dir-loader repo here](https://github.com/magda-io/magda-configmap-dir-loader/blob/master/src/test/main.spec.ts)). 
  - For this purpose, `magda-core` provides [magda.filesToJson](https://github.com/magda-io/magda/blob/21499b75c7a7ee00d68886338713217d83ccb91f/deploy/helm/magda-core/templates/_helpers.tpl#L244) helm chart template to load files with directory structure into a k8s configMap.
- Opa pod use `initContainer` [magda-configmap-dir-loader](https://github.com/magda-io/magda-configmap-dir-loader) to decode and put policy files in place

This template support 2 parameters:
- `filePattern`: Glob file search pattern string. All files (and their dir path) match the `Glob` pattern will be encoded and included in the configMap.
- `pathPrefix` : Optional. Add `pathPrefix` to all file path generated in configMap JSON.

Example Usage:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: "my-extra-policy-files-1"
data:
  my_extra_policy_files_1.json: {{ include "magda.filesToJson" (dict "root" . "filePattern" "policy_dir_1/**/*" ) }}
```

Or with `pathPrefix`:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: "my-extra-policy-files-2"
data:
  my_default_files.json: {{ include "magda.filesToJson" (dict "root" . "filePattern" "policy_dir_2/**/*" "pathPrefix" "test/" ) }}
```

> Note: `pathPrefix` is for splitting a giant policy directory (e.g. if larger than k8s 1mb limit) into more than one configMap

- After the configMap is in place, you can supply your custom policy file configMap by set the `customPolicyConfigMaps` value of `opa` chart. Here is an example:

```yaml
magda:
  magda-core:
    opa:
      customPolicyConfigMaps:
      - my-extra-policy-files-1
      - my-extra-policy-files-2
```

#### Test Site:

https://issue-2766-demo.dev.magda.io/

The test site is created via the changes (of a local magda deployment chart) on a demo branch. See this commit:
https://github.com/magda-io/magda/commit/9339dde1c30a43cec44646aea5aeda1391275931

Two custom policy files are provided:
- `system/echo.rego`
  - This will overwrite the built-in policy 
- `system/timestamp.rego`
  - This is a new policy file

Both policy files should be accessible via URL (**you need to port-forward opa pod first**):
http://localhost:8181/v1/data/system/echo
or 
http://localhost:8181/v1/data/system/timestamp

To list all policy files:
http://localhost:8181/v1/policies

### Checklist


t83714:
I think we will not need to update `magda-configmap-dir-loader` docker image probably for a while as its functionality is quite stable.
Thus, I think I'd better push a new commit so that OPA will not attempt to use `global.image`'s value to determine `magda-configmap-dir-loader` docker image's version or repo.
This could avoid the trouble of always having to override the image setting for `magda-configmap-dir-loader` docker image when `global.image` is set.

t83714:
I think it's ready for review now~

PR #2943, opened 2020-08-26
Issue/2929 Make magda-ui serves from a different URL path & other fixes

### LINKED ISSUE #2929: Allow Magda serves multiple Web UI server
### Allow Magda serves multiple Web UI server 

When customising Magda as their own solution, people may want to keep existing Magda UI running while implementing their own UI (to leverage magda's existing `login` or `Admin panel`) functionality.

We currently allow people to replace the UI but Magda web UI can't be running at the same time from a different path.

We probably can:
- No changes to `Gateway` (rather than #2928)
- Update `web-server` to:
  - Add helm chart parameter `uiBaseUrl`. Default Value "/". 
    - e.g. `/magda-web-ui` means serving magda UI from path `/magda-web-ui/`
  - Add proper `<base />` tag to 
- Update `web-client` 
  - Add `basename` to `BrowserRouter` component in `index.js`. Its value should be same as `baseUrlPath` that receives as part of [web-server-config](https://github.com/magda-io/magda/blob/master/magda-web-client/src/config.ts)
  - Replace all `<a>` tags with `<Link>` where applicable

@mwu2018  already implement most it [at here](https://github.com/TerriaJS/magda/blob/16c2b67b7e78a45f59f3e6bf68f920a37e02543e/magda-web-server/src/getIndexFileContent.ts#L12)

Although, we probably can simplify a bit (e.g. we don't need to replace `getIncludeHtml`)

### LINKED ISSUE #2941: Incorrect papaparse version bump in yarn.lock cause CSVFileLoader fail
**Describe the bug**
papaparse has been incorrectly bumped to 5.2.0 in yarn.lock
https://github.com/magda-io/magda/blob/5c61a9da723ebd681d6ef03d3342d8050a413de5/yarn.lock#L14822
However, version 5.1.0 is the last version include the DGA server range header fix:
https://github.com/magda-io/PapaParse/commit/1abb0ded3c5bb31cfc5f6426e895ff589cfa4a24

### LINKED ISSUE #2942: Reduce Size of Webserver docker image
**Describe the issue**
Currently, the web-server docker image includes some quite large web-client dependencies due to here:
https://github.com/magda-io/magda/blob/5c61a9da723ebd681d6ef03d3342d8050a413de5/magda-web-client/package.json#L17
We still want to keep the `dependencies` as it is because it's useful for `scss-compiler` to replicate an identical scss compiling environment.
However, we do want to exclude those dependencies in web-server docker image (i.e. only include `web-client/build` dir) to save around 500Mb.

t83714:
### What this PR does

This PR includes PR https://github.com/magda-io/magda/pull/2938. 

Fixes #2929 #2941 #2942 
- Added a new Helm parameter `uiBaseUrl` to web-server
  - This value will be supplied to frontend UI via `config.uiBaseUrl` as well
  - Its value should have a leading slash, but no trailing slash
- Make create-react-app to build resource with relative paths e.g. "./xxx"
- Added <base> tags to `create react app` index.html template
- Update web-server to auto-replace <base> tags with `uiBaseUrl`
- Set React Router `basename` with the value of  `uiBaseUrl` to so that we don't have to update all internal URLs in frontend
- Replace <a> with `<CommonLink>` that:
  - render a <Link> for internal url (except url for API e.g. /auth/xxx)
  - render a <a> for internal API url as it should starts from server root
  - render an external url as <a>
- Fixed some internal image urls
- Fixed #2941 CSV file loading issue
- Fixed #2942 reduce the docker image size
- Removed sameSite: "none" in `minikube-dev.yaml` --- chrome 80 is out. sameSite=none is now requires `secure` (otherwise, the cookie will be rejected by default)

### Tests

Tested:
- Local Cluster (minikube)
- CI deployed NO data deployment
- CI deployed UI Only Preview deployment
- CI deployed NO data deployment (UI serves at `magda-ui` with default route replace with `preview-map` service)

### Test Site:

https://issue-2929-demo.dev.magda.io/

- CI deployed NO data deployment (UI serves at `magda-ui` with default route replace with `preview-map` service)
- Home page is now replaced by magda preview map (forked version terria-map)
  - it's normal you see a blue screen --- Magda actually send config via iframe window.postMessage programming.
  - Just click zoom out button to see the global
- Got to https://issue-2929-demo.dev.magda.io/magda-ui for existing magda ui

See here for config changes required for setting up the demo site:

https://github.com/magda-io/magda/compare/60402b...4f44c7

### Checklist


t83714:
@mwu2018 @soyarsauce 
Just in case you're interested.
I will leave this PR open for a few days before merge it.
It comes with a demo site so you can have a play if you want to.
For how to set up a site with magda-ui serves at a different URL, see my comment in description.

magda-web-server/src/standardiseUiBaseUrl.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    if (urlStr.indexOf("/") !== 0) {
+        urlStr = "/" + urlStr;
+    }
+
+    return urlStr.lastIndexOf("/") === urlStr.length - 1
+        ? urlStr.substr(0, urlStr.length - 1)
+        : urlStr;
+}
```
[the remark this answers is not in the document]
soyarsauce: e.g. if base href is set to `/magda/`,

assuming dev.magda.io host,
a link pointing to `/this-route/` will resolve to `https://dev.magda.io/magda/this-route/`

whereas the same link, with href set to `/magda`, will resolve to `https://dev.magda.io/this-route/` because it will strip out the `magda` portion in `/magda`

soyarsauce (changes requested):

magda-web-server/src/getIndexFileContent.ts
```diff
@@ -92,6 +94,16 @@ async function getIndexFileContent(
...
         );
+
+        if (uiBaseUrl === "/") {
+            // --- if `uiBaseUrl` is '/' do nothing
+            return indexFileContent;
+        }
+
+        indexFileContent = indexFileContent.replace(
+            '<base href="/">',
+            `<base href="${uiBaseUrl}/">`
```
soyarsauce, line 105: Sorry I missed that the trailing slash gets appended here!

soyarsauce: I still think it's confusing if we append it here though - when you think about the default value of `/`, that itself can work in all cases (e.g. concatenate it with other paths) - whereas if you give it `/magda`, suddenly you have to treat it differently without doing extra stuff to standardise it anywhere you end up using it

t83714:
@soyarsauce Thanks a lot for looking at it  😄 
1> the `should have a leading slash, but no trailing slash` format is actually trying to match `React Router` 's document https://reactrouter.com/web/api/BrowserRouter/basename-string
2> Understand that different people might prefer a different URL / path format. Therefore, I try to make `uiBaseUrl` more transparent:
- I cleaned the input using `standardiseUiBaseUrl.ts` function --- actually, no matter you supply `magda-ui`, `/magda-ui` or `/magda-ui/`, they all work. I didn't mention this in docs though, as I think we should only give people who need instruction only one option to do things.
- In Magda frontend, people should avoid using `uiBaseUrl`. Instead, use react router's `<Link>` or my `<CommonLink>` (which renders <Link> depends on situations) --- we want to care less about the situation when Magda can serve at different path when work on new features. Plus, I also want to do a `clean-up` --- we had some place incorrectly used <a> which make js/css bundle re-load on every link click.
By the way, what do you think how it is configured:
image
it's done in PR 2938, mainly try to make sure you can have more control over web routes config as well.
e.g. if you need, you can turn on the `auth` support so the magda session header will be included when gateway forward your requests.

Issue #2978, opened 2020-09-24
Pull in updated minio version 5.0.33


soyarsauce:
## Description

Running helm upgrade will restart minio even with no change due to https://github.com/helm/charts/issues/21590 but is fixed in 5.0.27+

(On top of that, helm `charts` repo is deprecated)

Aside while on updating charts & storage api, another thing that would be more useful for the storage-api is to remove the hardcoded `defaultBuckets` & pull in `defaultBuckets` via chart values

PR #2979, opened 2020-09-25
Issue/2978

soyarsauce:
### What this PR does

Fixes https://github.com/magda-io/magda/issues/2978 except the `defaultBuckets` described in the issue

Currently you can't actually run the storage api to run with GCS backing instead of PV in cluster - this will address the issue of trying to use storage api with GCS as you can actually specify the region. Also brings in 5.0.33 so that the deployment isn't restarted on every single `helm upgrade`

(Probably shouldn't have bumped the version too - i can rebase it out, just did it so I could use the image immediately)

### Checklist


t83714:
🎉 great job~ I always wanted to fix the auto pod rolling bug of minio but never got a chance to do it~
it's also nice to make it support GCS 💯 
- regarding the version --- yeah it will be good be revert it back to `0.0.58-alpha.0`.  At this moment, trigger the release is too easy so, to avoid mistake, `-alpha.0` will always be the version placeholder on the master branch until we formally release version 58. 
  - If too hard to rebase it, you can just run `yarn set-version` to update version on all modules (including helm charts). 
    - That script will create a version tag as well --- just don't push that tag as it will release a version
  - We still can release a different pre-release version even when the version on Master stays on `0.0.58-alpha.0`
    - We keep just do it in a separate release branch, use `yarn set-version` to update the version and then push the tag to trigger the release. So release can have any version without impact the version on master branch. 
  - The master by default use helm chart app version as docker image version. For helm chart testing, you can set `global.image.tag` in your deploy values file to overwrite it [like this](https://github.com/magda-io/magda/blob/7a55fe884a77842210c96e848f97bc9001a19975/deploy/helm/minikube-dev.yml#L8). Therefore, you can test your local chart without releasing a different tag docker image.
- The error on the CI pipeline is from the [helm docs checking job](https://gitlab.com/magda-data/magda/-/jobs/756667758).
  - the documents are here: https://magda.io/docs/helm-charts-docs-index
  -  I put the job in recently so if you touched any values.files which lead to the helm charts document changes that are not committed, It will report an error. You just need to run [helm-docs](https://github.com/norwoodj/helm-docs) and review the document changes and commit them.

soyarsauce:
Superseded by https://github.com/magda-io/magda/pull/2990

PR #2990, opened 2020-10-01
Issue/2978 rebased

soyarsauce:
### What this PR does
Rebased https://github.com/magda-io/magda/pull/2979 with the version tweak dropped

Fixes https://github.com/magda-io/magda/issues/2978 except the `defaultBuckets` described in the issue

Currently you can't actually run the storage api to run with GCS backing instead of PV in cluster - this will address the issue of trying to use storage api with GCS as you can actually specify the region. Also brings in 5.0.33 so that the deployment isn't restarted on every single `helm upgrade`

(Probably shouldn't have bumped the version too - i can rebase it out, just did it so I could use the image immediately)

### Checklist


t83714:
worked well in dev cluster (a new deployment).
However, when upgrading my local cluster deployment, I keep getting the following error:
```
ERROR Unable to initialize backend: Insufficient permissions to access path
      > Please ensure the specified path can be accessed
```
It might be just my local issue. 
Will play more and report back.

soyarsauce:
@t83714 I remember seeing that briefly, but seems like it should have been fixed
https://github.com/minio/minio/issues/9746
https://github.com/minio/minio/issues/9454

t83714:
@soyarsauce I think the issue probably still there.
How to re-produce
- delete existing minio deployment & pvc
- deploy magda with minio 5.0.18
- Using add dataset function to create a dataset with uploaded files (or use storage API upload a few files)
- deploy magda with minio 5.0.33
- the minio pod will stuck with the following error:

image

Notice there is `close to merging` PR here:
https://github.com/minio/minio/pull/9844
not sure why people didn't work on it eventually.

I will have more play too see if there is anyway for migration exiting versions.

t83714:
@soyarsauce 
Alright~ I found a solution.
- delete minio deployment object
- deploy a [shell pod](https://gist.github.com/t83714/b52f699c394f4a2150212f24fd7aa4fa) to access the PV: 
```bash
kubectl apply -f https://gist.githubusercontent.com/t83714/b52f699c394f4a2150212f24fd7aa4fa/raw/6bd4904fa056f4ae4e44841b62325d6c9a6e86f0/shell-pod.yaml
```
- Shell into the shell pod (if you use k9s can simply press `s` key)
- Go to `/export` and run the followings to all directories (especially the hidden dir `.minio.sys`)
```
chown -R 1000 <data-dir> && sudo chmod u+rxw <data-dir>
```
- redeploy Magda and bring back minio 5.0.33. it will now be running ok

So it seems to migrate to v33 change the owner to `root` while the deployment security context try to access it.

Considering the hassle of migrating it, do you think we probably just migrate to 7.1.2 using minio's official helm chart repo here: https://github.com/minio/charts ?

soyarsauce:
Yeah using the new repo may be the best option if that avoids the migrating-existing-versions problem (and if `7.1.2` has no breaking changes that can't readily be adjusted for)

t83714:
I will probably have a play later to see how 7.1.2 goes.
If it's too hard, will just leave it to 5.0.33 and write a short migration doc for v0.0.57 to v0.0.58.
Otherwise, will set it to 7.1.2 so we don't need to worry about upgrading for sometime

t83714 (approved):
👍

t83714:
@soyarsauce Tested `minio` 7.1.2 in the cluster for Magda usage and seems works well.
By comparing v5 vs v7 chart code, there are no major changes actually --- the issue we had is one of the incompatible things.
I have created a migration document (for existing users) for this purpose.
Also, added in some default values:
```
persistence:
    size: 10Gi
```
The default value used to be part of `minio` but seems has been moved in the newer version.
Add here so existing users won't get stuck when upgrading their deployment (if they rely on the default values)

I believe the `7.1.2` should also work for saas project GCS usage as well.
Could you please have a test for that use case when you have time?
I've added the PR approval so if it works, you can merge the PR anytime.

soyarsauce:
As discussed, you won't get anything useful by testing this on the digital twins as they don't use this specific implementation of the storage API. Try it on development saas, with the basic goal of:
- simply that it will work / migrate across smoothly if you have major version 5 in-place
- and that any chart changes between major versions 5 & 7 (the one proposed here) will not be adversely affected by any default values, etc

Bonus,
- 0 downtime, but I think this is unfortunately impossible in the current state of storage API `createApiRouter` anyway

t83714:
have to merge the master otherwise, it won't build due to the dependencies requirements on builder docker images

t83714:
tested over draft PR: https://github.com/TerriaJS/saas/pull/64
Will leave this PR open till Friday and merge it (unless any issue discovered)

soyarsauce:
My goodness thanks Jacky!! 🙏🙏  @t83714

Issue #2992, opened 2020-10-02
We need to re-look at liveness / readiness Probs logic of our pods


t83714:
### We need to re-look at liveness/readiness Probs logic of our pods

We have done some work around liveness/readiness Probs for zero-downtime deployment here: #1471

We need to re-look at it to make sure that, during k8s rolling update (particularly for registry API), if DB is not accessible, will liveness/readiness Probs reports this correctly.

t83714:
Registry seems not checking anything with database at all and just simply reply OK (for both liveness & readiness):
https://github.com/magda-io/magda/blob/f52fcc43380070fb71a06ce5e11427d1aa1412b2/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/Api.scala#L139

soyarsauce:
:feature:

soyarsauce:
At a minimum we should resolve this for these ones that utilise DB, namely authorization-api, content-api, registry-api

soyarsauce:
correspondence utilises this well for a smtp dep (instead of db dep for the above)

https://github.com/magda-io/magda/blob/master/magda-correspondence-api/src/createApiRouter.ts#L56-L67
https://github.com/magda-io/magda/blob/master/magda-correspondence-api/src/test/createApiRouter.spec.ts#L113-L141

soyarsauce:
turns out authorization & content are OK
registry PR at https://github.com/magda-io/magda/pull/2997

soyarsauce:
Same problem in storage-api

https://github.com/magda-io/magda/blob/v0.0.58-rc.3/magda-storage-api/src/createApiRouter.ts#L31-L39

t83714:
Add #3024 as the blocker as, currently, there is a performance bottleneck that only `registry-full` pod serve the `/api/v0/registry` endpoint and we can't scale it up. 

Our UI always sends read requests to `/api/v0/registry-read-only` but we can't guarantee that third-party software will do the same --- especially, for metadata crawlers.  

Querying DB in readiness probe may add extra burden to it when `registry-full` is already on full speed.

PR #2997, opened 2020-10-06
Added basic readiness check for registry API

soyarsauce:
### What this PR does

Fixes https://github.com/magda-io/magda/issues/2992

### Checklist

### DIFF
--- a/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/ProbesService.scala
+++ b/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/ProbesService.scala
@@ -1,0 +1,61 @@
+package au.csiro.data61.magda.registry
+
+import java.util.concurrent.TimeoutException
+
+import akka.actor.{ActorRef, ActorSystem}
+import akka.event.Logging
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Route
+import akka.stream.Materializer
+import au.csiro.data61.magda.client.AuthApiClient
+import au.csiro.data61.magda.directives.AuthDirectives.requireIsAdmin
+import au.csiro.data61.magda.directives.TenantDirectives.{
+  requiresTenantId,
+  requiresSpecifiedTenantId
+}
+import au.csiro.data61.magda.model.Registry._
+import com.typesafe.config.Config
+import gnieh.diffson.sprayJson._
+import io.swagger.annotations._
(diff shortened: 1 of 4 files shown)


magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/ProbesService.scala
```diff
@@ -0,0 +1,62 @@
...
+    path("ready") {
+      DB readOnly { session =>
+        recordPersistence.getReadiness(Some(logger))(session)
+      } match {
+        case Some(true) => complete("OK")
+        case Some(false) | None =>
+          complete(StatusCodes.InternalServerError, "Database not ready")
+      }
+    }
+  }
```
t83714, line 57: Tested in the cluster and it seems most exceptions are not captured.
e.g. 
```
java.util.NoSuchElementException: key not found: to_regclass
OR org.postgresql.util.PSQLException: Connection to registry-db:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
OR org.postgresql.util.PSQLException: FATAL: role "client" does not exist
```
I think you probably need to capture any failure from the DB.readonly method as the execution is happening there.
you might want to try scala's `util.Try` instead of `try/catch`  --- it's more common in scala as it's more flexible / offer more streamlined exception processing 
https://www.scala-lang.org/api/2.9.3/scala/util/Try.html 

You can do something like this:
```scala
import scala.util.{Try, Failure, Success}
Try{
        DB readOnly { session =>
          recordPersistence.getReadiness(Some(logger))(session)
        }
    } match {
        case Success(Some(true)) => complete("OK")
        case Success(Some(false)) | Success(None) =>
          complete(StatusCodes.InternalServerError, "Database not ready")
        case Failure(exception) => complete(StatusCodes.InternalServerError, "Database not ready Other")
    }
```
The `recordPersistence.getReadiness` doesn't seem to need to capture any exceptions (so you don't have to capture at multiple place) nor has a chance to produce a NONE. I guess a Boolean could simpler

t83714 (changes requested):
I think just some minor error handling changes are required.

soyarsauce:
I'm sorry @t83714 I've run out of time for this one, are you able to do the error handling changes you have in mind, then get this in? 🙏 I think it would still be extremely useful for the registry API to have this check!

t83714:
@soyarsauce 
Thanks for your work on this and just leave to me.
I will get this done after this one #3024

PR #3026, opened 2020-11-04
add migration info for authentication plugins

t83714:
### What this PR does

- add migration info for authentication plugins

### Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -97,6 +97,8 @@ Based on [PassportJS](http://www.passportjs.org/), Magda's authentication system
     </video>
 </p>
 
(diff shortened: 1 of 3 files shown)


t83714:
@soyarsauce  I added some migration info for users who migrates from v57 to v58 as you suggested.
If you're interested, you can have a look~

soyarsauce:
Yeah nice!! This is exactly what I was thinking and would find useful! Just a tiny update, and the link to the dev cluster example I can see being useful for people upgrading to 0.0.58

soyarsauce (approved):

t83714:
Thanks, @soyarsauce for your review 👍

PR #3027, opened 2020-11-04
Issue/2985 Allow default web route to be configured like all other route (allow specify method, auth etc.)

### LINKED ISSUE #2985: Allow default web route to be configured like all other route (allow specify method, auth etc.)
### Allow default web route to be configured like all other web routes (allow specify methods, auth etc.)

#2928 already made `webRoutes` be able to configure with more options (e.g. auth, method etc).

However, the default web route still use (and the only one) `createGenericProxy`

https://github.com/magda-io/magda/blob/d3f85b69e69f325f7712b412a63336299286372f/magda-gateway/src/buildApp.ts#L270

Due to security concerns, we probably can't hardcode `createGenericProxy` to proxy all methods and should let the user to config instead.

Also, we will want the helm config to be compatible with the existing config (easier to upgrade existing sites).

Thus, propose the following solution:

### Possible Solution:
- By default, the string `web` field config [here](https://github.com/magda-io/magda/blob/d3f85b69e69f325f7712b412a63336299286372f/deploy/helm/internal-charts/gateway/values.yaml#L88) still works as before. i.e.
  - Assume only proxy GET method & no auth etc.
- accept an optional new `defaultWebRouteConfig ` section that allows to config all `proxy target` options. i.e.:
  - `methods`: array of string. "all" means all methods
  - `auth`: whether this target requires session. Otherwise, session / password related midddleware won't run
  - `redirectTrailingSlash`: make /xxx auto redirect to /xxxx/ 
  - `statusCheck`: check target's live status from the gateway
  - proxy target URL should still be configured via `web` config field above
- Remove `createGenericProxy` from codebase (only default web route uses it now) and use `createGenericProxyRouter` instead for proxying default web route.

t83714:
### What this PR does

Fixes #2985 
- Allow default web route to be configured like all other routes (allow specify method, auth etc.)
- Removed `createGenericProxy` from Gateway codebase
- Existing config still work (i.e. only specify `web` option). No impact with current behaviour: by default still only proxy by GET if `defaultWebRouteConfig` option not specified
- Test cases are included

More info re: config see Gateway Helm Chart Document

### Checklist


t83714:
tested locally & in cluster plus test cases

Issue #3036, opened 2020-11-19
Gateway should only check route readiness when `statusCheck` == true


t83714:
### Gateway should only check route readiness when `statusCheck` == true

1> Most gateway routes are our internal components. 
Unless they are critical to gateway's core functionality (rather than a proxy target only), we shouldn't check those route's availability (as part of gateway readiness check) as they should be checked by their own live / readiness probes already.

2> Moreover, at the moment, we actually didn't use the gateway readiness probe (liveness probe is correctly used). See

https://github.com/magda-io/magda/blob/39a9b34e54870ded4eaddacce63a57b209dfccfd/deploy/helm/internal-charts/gateway/templates/deployment.yaml#L113-L119

We should change it to "/status/ready"

t83714:
closed via pr https://github.com/magda-io/magda/pull/3037

Issue #3035, opened 2020-11-19
Update our docs to replace Helm Stable Repo usage


t83714:
## Update our docs to replace Helm Stable Repo usage

Helm stable repo (https://charts.helm.sh/stable) is deprecated and some of the charts (e.g. docker registry) are no longer available.

We should replace it with charts published on https://artifacthub.io/

[incubator](http://storage.googleapis.com/kubernetes-charts-incubator), however, should be removed from our docs as we used our own version of `magda-io/kube-registry-proxy`

t83714:
closed via https://github.com/magda-io/magda/pull/3037

PR #3037, opened 2020-11-19
Issue/3009 Allow Magda API endpoints to be served at non-root path

### LINKED ISSUE #3009: Allow Magda API endpoints to be served at non-root path
### Allow Magda API endpoints to be served at a non-root path

When integrated with external service, it's possible that the client wants to host Magda at a non-root path.
e.g. Having the home page fully managed by his own CMS.

We have made UI frontend workable from a non-root path in #2929 

This ticket is about to make sure all API endpoints can work at a non-root path

t83714:
### What this PR does

Fixes #3009 Allow Magda API endpoints to be served at non-root path
Also addressed:
- #3035 Update our docs to replace not working deprecated Helm Stable Repo usage
- #3036 Gateway should only check route readiness when `statusCheck` == true
- #3038 Remove ejs views from Gateway
- Fixed: Distribution Page may show a blank page when the distribution has no downloadUrl #3039
- Fixed: Date Extractor for metadata creation tool will fail to process CSV or excel with incomplete rows #3040
- Other Date Extractor extration accuracy improvement

### Checklist


t83714:
tested:
- gateway / backend apis amounted at non-root path
- ui is mounted at non-root path (via web-server.uiBaseUrl)
- UI only local test server

PR #3043, opened 2020-11-30
Admin Panel User Admin feature should set/unset both isAdmin field & admin role for compatibility purpose

### LINKED ISSUE #3033: Admin Panel Make Admin feature should add adminRole to the user as well
### Admin Panel Make Admin feature should add adminRole to the user as well

The user management page currently only set `isAdmin` field:

https://github.com/magda-io/magda/blob/f5686bbd0d643e6b93af0893039b61548d04b12a/magda-web-client/src/Components/Account/AccountsAdminPage.js#L77

We need to make this function to add an admin role to the user as well

### Updates

We probably also need a simply dedicated cmd tool to set/unset a user as admin. This is will be useful for the newly installed instance when you don't have an admin account yet.

t83714:
### What this PR does

Fixes #3033 
- Admin Panel User Admin feature should set/unset both isAdmin field & admin role for compatibility purpose 
-   Make GET /auth/users/all API return user role info as well
-   Make PUT /auth/users/:userId API accept other user field rather then `isAdmin` field only
- Bug fixes & UI adjustment

### Checklist

-   [x]  unit tests aren't applicable


t83714:
tested locally & in the cluster. Worked well.
Merging

PR #3050, opened 2020-12-15
Issue/3024 Make Gateway should proxy all GET requests to registry-read-only pods

### LINKED ISSUE #3024: Gateway should proxy all GET requests to registry-read-only pods
### Gateway should proxy all GET requests to registry-read-only pods
We have registry-read-only pods/services to serve read-only registry requests.
They are designed to reduce the workload from the `registry-full` service that allows writes (we only can have one `registry-full` pod ).
However, `registry-full` service is still the default access endpoint behinds `/api/v0/registry`.
Therefore, unless you send requests to `/api/v0/registry-read-only`, `registry-full` is still serving most of "read" requests (People who crawl registry API is likely to use this default endpoint ).
To solve this, we should make gateway proxy all GET requests that sent to `/api/v0/registry` to registry-read-only pods.

~[http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) probably can help this.~

Update: we made gateway support new (optional) syntax for method (current config is still supported and not affected):
```
registry:
    to: http://registry-api/v0
    methods:
    - method: get
      target: http://registry-api-read-only/v0
    - method: head
      target: http://registry-api-read-only/v0
    - method: options
      target: http://registry-api-read-only/v0
    - method: post
      target: http://registry-api/v0
    - method: put
      target: http://registry-api/v0
    - method: patch
      target: http://registry-api/v0
    - method: delete
      target: http://registry-api/v0
    auth: true
```
- The main idea is to proxy incoming gateway request by HTTP method. 
  - All GET requests will be forwarded to read-only registry nodes
- However, Hooks API GET endpoints are only available from main registry node (Some webhook info is from registry event processing actor. Thus, can only be retrieved from the main node). We will:  
  - Proxy `/registry/hooks` at the gateway to the main registry API hooks endpoint.
    - However, expressjs order router routes by the sequence of registering routes with expressjs, rather than longer path got higher priority.  Therefore, route `/registry/hooks` might still be overwritten by route `/registry`. We will re-order all routes configuration (by length of seqment) before set-up proxy to avoid this issue.

t83714:
### What this PR does

Fixes #3024 

- Gateway should proxy all GET requests (at `/api/v0/registry`) to registry-read-only nodes
  - Registry Hooks APIs Still forward to main registry node as the hook info has to be retrieved from the main registry node
- handle proxy path handling manually so that trailing slash '/' won't be auto-added to target URL. 
  - (e.g. Before `/route` to `/target-route/` Now: `/route` to `/target-route`)
  - This for making proxying `registry/hooks` request possible (and not conflict with `registry/hooks/id` endpoint)
- Indexer will pull datasets from registry read-only nodes

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 -   Make PUT /auth/users/:userId API accept other user field rather then `isAdmin` field only
 -   Release `acs-cmd` & `org-tree` as separate NPM packages
 -   Add set admin function to `acs-cmd` commandline utility package
+-   #3024 Gateway will automatically proxy all GET requests received at `/api/v0/registry` to registry read-only nodes (excepts Hooks APIs)
 
 ## 0.0.58
 
--- a/deploy/helm/internal-charts/gateway/README.md
+++ b/deploy/helm/internal-charts/gateway/README.md
@@ -63,7 +63,11 @@ Kubernetes: `>= 1.14.0-0`
 
 A proxy target definition that defines `defaultRoutes` or `webRoutes` above support the following fields:
 - `to`: target url
-- `methods`: array of string. "all" means all methods
+- `methods`: array of string or objects with fields "method" & "target".
+  - When the array item is a string, the string stands for the HTTP method. "all" means all methods.
+  - When the array item us an object, the object can have the following fields:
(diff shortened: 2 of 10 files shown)


t83714:
Tested in local & test cluster. Mereged~

PR #3058, opened 2021-01-16
#3057 Fixed API `/auth/plugins` will skip the failed loading config item without throwing error

### LINKED ISSUE #3057: Skip failed loading auth config item rather than break whole config loading process
### Skip failed loading auth config item rather than break whole config loading process

Currently, API `/auth/plugins` will break if one of auth config item is failed to load.

The API should, instead, skip the failed loading config item and return other config items without breaking request processing.

t83714:
### What this PR does

Fixes #3057 

Fixed API `/auth/plugins` will skip the failed loading config item without throwing error

### Checklist


t83714:
tested locally in local cluster

Issue #2736, opened 2020-02-17
K8S Namespace is unable to be set to non-default in Admin API


AlexGilleran:
**Describe the bug**
From Dominik at Merck:

> I painfully found out that K8S namespace is hardcoded to “default” here https://github.com/magda-io/magda/blob/release/0.0.56/magda-admin-api/src/k8sApi.ts#L20 and that some additional RBAC settings were required

t83714:
closed via https://github.com/magda-io/magda/pull/3075

Issue #1774, opened 2018-10-12
Admin-api needs to run under a seperate k8s service account


t83714:
### Problem description

Currently, admin-api using service account `system:serviceaccount:default:default` to access k8s API.
The default service account is the account all pods have access to.
In the newer version k8s cluster, the default account won't get permission to access most resources.

We can simply create a roleBinding like below:
```
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: srv-acc-admin
subjects:
  - kind: ServiceAccount
    name: default
    namespace: default
roleRef:
  kind: ClusterRole
  name: cluster-admin
  apiGroup: rbac.authorization.k8s.io
```
To assign `default` account the `cluster admin` role to make it work.

However, it won't be safe as the default service account is the account every pod run on.

We should:
- create a separate service account with cluster admin role
- modify our k8s API code to authenticated as that `service account`

We will need service account CA & token to config our k8s client.
CA & token can be specified via Config.SERVICEACCOUNT_CA_PATH & Config.SERVICEACCOUNT_TOKEN_PATH

PR #3075, opened 2021-02-09
Fixed: A few bugs that make admin-api pod keep crash in non-default namesapce

t83714:
### What this PR does
- #1774 Admin-api should run as a separate k8s service account
- #2736 K8S Namespace is unable to be set to non-default in Admin API

### Checklist


t83714:
We still need to look at #3078 sometimes in future.
After this PR, at lease Admin API will not keep falling which may interrupt the deployment process

PR #3082, opened 2021-02-24
Update Ask A Question options

yayalu:
### What this PR does

Updates dataset pages and Ask A Question options for datasets with no contact points.

### DIFF
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -1,11 +1,11 @@
 import PropTypes from "prop-types";
 
+import React from "react";
 import MarkdownViewer from "Components/Common/MarkdownViewer";
 import MagdaNamespacesConsumer from "Components/i18n/MagdaNamespacesConsumer";
 import ToggleButton from "./ToggleButton";
 import { gapi } from "analytics/ga";
 
-import React from "react";
 import "./ContactPoint.scss";
 
 class ContactPoint extends React.Component {
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -52,7 +52,7 @@ class ContactPoint extends React.Component {
                                 </div>
                             </React.Fragment>
                         ) : (
-                            <div style={{ marginTop: "0.5rem" }}>
+                            <div className="description-text">
                                 This publisher has not provided a contact point.
                                 Try visiting the original resource for more
                                 information:{" "}
@@ -75,11 +75,13 @@ class ContactPoint extends React.Component {
 }
 
 ContactPoint.propTypes = {
-    contactPoint: PropTypes.string
+    contactPoint: PropTypes.string,
+    source: PropTypes.string
 };
 
 ContactPoint.defaultProps = {
-    contactPoint: ""
+    contactPoint: "",
+    source: ""
 };
 
 export default ContactPoint;
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -5,6 +5,7 @@ import MagdaNamespacesConsumer from "Components/i18n/MagdaNamespacesConsumer";
 import ToggleButton from "./ToggleButton";
 import { gapi } from "analytics/ga";
 
+import React from "react";
 import "./ContactPoint.scss";
 
 class ContactPoint extends React.Component {
@@ -58,6 +59,7 @@ class ContactPoint extends React.Component {
                                 <a
                                     href={this.props.landingPage}
                                     target="_blank"
+                                    rel="noreferrer"
                                 >
                                     {this.props.source}
                                 </a>
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -1,4 +1,3 @@
-import React from "react";
 import PropTypes from "prop-types";
 
 import MarkdownViewer from "Components/Common/MarkdownViewer";
@@ -76,12 +75,14 @@ class ContactPoint extends React.Component {
 
 ContactPoint.propTypes = {
     contactPoint: PropTypes.string,
-    source: PropTypes.string
+    source: PropTypes.string,
+    landingPage: PropTypes.string
 };
 
 ContactPoint.defaultProps = {
     contactPoint: "",
-    source: ""
+    source: "",
+    landingPage: ""
 };
 
 export default ContactPoint;
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -1,6 +1,5 @@
-import PropTypes from "prop-types";
-
 import React from "react";
+import PropTypes from "prop-types";
 import MarkdownViewer from "Components/Common/MarkdownViewer";
 import MagdaNamespacesConsumer from "Components/i18n/MagdaNamespacesConsumer";
 import ToggleButton from "./ToggleButton";
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -26,9 +26,9 @@ class ContactPoint extends React.Component {
             <MagdaNamespacesConsumer ns={["datasetPage"]}>
                 {(translate) => (
                     <div className="dataset-contact-point">
-                        <span className="description-heading">
+                        <div className="description-heading">
                             {translate(["contactPointTitle", "Contact Point"])}:
-                        </span>
+                        </div>
                         {this.props.contactPoint ? (
                             <React.Fragment>
                                 <div className="no-print">
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -1,11 +1,10 @@
-import PropTypes from "prop-types";
-
+import { gapi } from "analytics/ga";
 import MarkdownViewer from "Components/Common/MarkdownViewer";
 import MagdaNamespacesConsumer from "Components/i18n/MagdaNamespacesConsumer";
-import ToggleButton from "./ToggleButton";
-import { gapi } from "analytics/ga";
-
+import PropTypes from "prop-types";
+import React from "react";
 import "./ContactPoint.scss";
+import ToggleButton from "./ToggleButton";
 
 class ContactPoint extends React.Component {
     state = { reveal: false };
--- a/magda-web-client/src/Components/Common/ContactPoint.js
+++ b/magda-web-client/src/Components/Common/ContactPoint.js
@@ -27,27 +27,46 @@ class ContactPoint extends React.Component {
             <MagdaNamespacesConsumer ns={["datasetPage"]}>
                 {(translate) => (
                     <div className="dataset-contact-point">
-                        <div className="heading">
+                        <span className="description-heading">
                             {translate(["contactPointTitle", "Contact Point"])}:
-                        </div>
-                        <div className="no-print">
-                            {this.state.reveal ? (
-                                <MarkdownViewer
-                                    markdown={this.props.contactPoint}
-                                />
-                            ) : (
-                                <ToggleButton
-                                    onClick={this.onRevealButtonClick}
+                        </span>
+                        {this.props.contactPoint ? (
+                            <React.Fragment>
+                                <div className="no-print">
+                                    {this.state.reveal ? (
+                                        <MarkdownViewer
+                                            markdown={this.props.contactPoint}
+                                        />
+                                    ) : (
+                                        <ToggleButton
+                                            onClick={this.onRevealButtonClick}
+                                        >
+                                            <span>Click to reveal</span>
+                                        </ToggleButton>
+                                    )}
+                                </div>
+                                <div className="print-only">
+                                    <MarkdownViewer
+                                        markdown={this.props.contactPoint}
+                                    />
+                                </div>
+                            </React.Fragment>
+                        ) : (
+                            <div style={{ marginTop: "0.5rem" }}>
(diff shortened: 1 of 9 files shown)


magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
```diff
@@ -85,27 +85,36 @@ export default class DatasetPageSuggestForm extends React.Component {
         };
         const formProps = {
             title: false,
+            description:
+                "Your question will be sent to the publishers of the data: " +
+                this.props.contactPoint,
```
sukhrajghuman, line 90: could be worth replicating server side code here
i.e.

this.props.contactPoint.match(/([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9._-]+)/g);

or 

```
function getEmailFromText( text ) {
  return text.match(/([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9._-]+)/g);
}
```

sukhrajghuman: actually might be better to do without the /g, if we only want the first match

t83714, line 90: May need to re-phrase the text to consider the situation when `this.props.contactPoint` doesn't exist?
Otherwise, it might show as:
`Your question will be sent to the publishers of the data: undefined`

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
```diff
@@ -85,27 +85,36 @@ export default class DatasetPageSuggestForm extends React.Component {
...
             namePlaceHolder: "Dorothy Hill",
             emailPlaceHolder: "dorothyhill@example.com",
             textAreaPlaceHolder:
                 "Ask a question or report a problem about this dataset.",
             textAreaLabel: "What would you like to ask about this dataset?",
-            successHeader: "Your request has been submitted!"
+            successHeader:
+                "Your request has been submitted to: " +
+                this.props.contactPoint +
```
t83714, line 98: Currently, this request will be emailed to one single recipient (email address), right?
It seems the text here suggests different behaviour?
Or we have plan to adjust the logic of sending emails?

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
```diff
@@ -85,27 +85,36 @@ export default class DatasetPageSuggestForm extends React.Component {
...
+                        <div className="dataset-button-container no-print">
+                            <button
+                                className="au-btn au-btn--secondary ask-question-button"
+                                onClick={this.toggleShowForm}
+                            >
+                                Ask a question about this dataset
+                            </button>
+                        </div>
+                    ) : (
+                        <React.Fragment />
```
t83714, line 116: This can be a `null` instead --- <React.Fragment /> will be an object of type fragment which will have to be created by react

t83714 (changes requested):
@yayalu Thanks & Looks good 👍 
Have left a few comments.
Also created a test site via CI, here is the link (in case you want to have a play):
https://update-ask-question.dev.magda.io

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
yayalu: Thanks! Updated it. included the /g in the end as without it there were multiple matches for the same email address, resulting in repeated outputs.

magda-web-client/src/Components/Common/ContactPoint.js
```diff
@@ -27,27 +27,46 @@ class ContactPoint extends React.Component {
...
+                                    )}
+                                </div>
+                                <div className="print-only">
+                                    <MarkdownViewer
+                                        markdown={this.props.contactPoint}
+                                    />
+                                </div>
+                            </React.Fragment>
+                        ) : (
+                            <div style={{ marginTop: "0.5rem" }}>
```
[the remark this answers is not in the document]
yayalu: Done!

magda-web-client/src/Components/Common/ContactPoint.js
```diff
@@ -27,27 +27,46 @@ class ContactPoint extends React.Component {
...
-                                    <span>Click to reveal</span>
-                                </ToggleButton>
-                            )}
-                        </div>
-                        <div className="print-only">
-                            <MarkdownViewer
-                                markdown={this.props.contactPoint}
-                            />
-                        </div>
+                                    {this.props.source}
```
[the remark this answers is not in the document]
yayalu: Done!

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
yayalu: When the primary contact point (in this case, email) doesn't exist, I've refactored it to output the remaining attributes in the modal instead. You can see this with "/dataset/ds-aurin-aurin%3Adatasource-TUA_PHIDU-UoM_AURIN_DB_sa2_incomesupport" published by GeoServer.

We have also removed the "Ask a question" button for all datasets with no contact points. This should solve for the remaining undefined contact point issues.

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
yayalu: I've rephrased this, hopefully it will be clearer that there is only one publisher.
Requests will only be sent to the publisher of the dataset, so only one email address.

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
yayalu: Done!

magda-web-client/src/Components/Common/ContactPoint.js
```diff
@@ -56,11 +75,13 @@ class ContactPoint extends React.Component {
 }
 
 ContactPoint.propTypes = {
-    contactPoint: PropTypes.string
+    contactPoint: PropTypes.string,
+    source: PropTypes.string
```
t83714, line 79: probably still need landingPage here

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
```diff
@@ -83,20 +83,23 @@ export default class DatasetPageSuggestForm extends React.Component {
                 maxHeight: "95vh"
             }
         };
+        var contactPointMatch = this.props.contactPoint.match(
```
t83714, line 86: 1> better use `let` or `const` rather than var --- I think it (`var`) might not pass our CI code formating check
2> as `this.props.contactPoint` potentially could be a null or undefined value (that has no `.match`), to avoid throw an error, we probably want check its value first.
e.g. let contactPointMatch = ""; if(typeof this.props?.contactPoint === "string") {....}
3> And I don't think `match` will return a string type? it should return an array
4> `"sdds dssd@dsd.com.au sdsd".match(/([a-zA-Z0-9.-]+@[a-zA-Z0-9.-]+.[a-zA-Z0-9._-]+)/g)` will match
`"dssd@dsd.com.au sdsd"`  --- we probably need a better email regex here

t83714 (changes requested):
some minor things to have a look

magda-web-client/src/Components/Common/ContactPoint.js
yayalu: Woops, updated thanks.

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
yayalu: 1) fixed (although var seemed to pass the tests with the previous commit)
2) Did an initial check
3) Also fixed
4) I struggled to find a better email regex, the short-term solution I found was just getting the first email match (should be sufficient as there is only one publisher), and just split by whitespace. If you have other regex recommendations let me know.

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
sukhrajghuman: Hey Jacky,
The desired functionality is to not send to the default recipient (data@digital.gov.au) when there is no contact email provided.

I believe if we [update lines 187-194](https://github.com/magda-io/magda/blob/373c89f074d8aa99a4f6b1dc385722198b3531ec/magda-correspondence-api/src/createApiRouter.ts#L187-L194) in the `/public/send/dataset/:datasetId/question` endpoint, it would achieve this functionality.

We can use this

```
                const emails = [contactPointEmail, datasetPublisherEmail].filter(...)

                if (!validEmail) {
                    return res.status(404).json({
                        status: "Failed to send",
                        error:
                            "Could not find a contact email for the publisher"
                    });
                }

                const recipient = emails[0];
```
I believe the contactPointEmail corresponds to this.props.contactPoint?

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
```diff
@@ -83,29 +83,47 @@ export default class DatasetPageSuggestForm extends React.Component {
                 maxHeight: "95vh"
             }
         };
+        let contactPointMatch;
+        if (
+            this.props?.contactPoint &&
+            typeof this.props?.contactPoint === "string"
+        ) {
+            contactPointMatch = this.props.contactPoint.match(
```
t83714, line 91: How about doing this? Current regex missed `asd_ssd@dmail.com` etc and unnecessarily captured whitespace. 
Plus, we already used regex no need to manually process the string 
```
const emailRegex = /[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}/ig;
const matches = this.props.contactPoint.match(emailRegex);
contactPointMatch = matches?.length ? matches[0] : this.props.contactPoint;
```

t83714 (changes requested):
a few changes & nearly done~ 😄

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
```diff
@@ -83,29 +83,47 @@ export default class DatasetPageSuggestForm extends React.Component {
                 maxHeight: "95vh"
             }
         };
+        let contactPointMatch;
+        if (
+            this.props?.contactPoint &&
```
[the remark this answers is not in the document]
yayalu: Fixed

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
yayalu: Looks good! Thanks, fixed.

magda-web-client/src/Components/Common/ContactPoint.js
```diff
@@ -1,11 +1,11 @@
-import React from "react";
 import PropTypes from "prop-types";
 
 import MarkdownViewer from "Components/Common/MarkdownViewer";
 import MagdaNamespacesConsumer from "Components/i18n/MagdaNamespacesConsumer";
 import ToggleButton from "./ToggleButton";
 import { gapi } from "analytics/ga";
 
+import React from "react";
```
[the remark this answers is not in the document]
yayalu: Seems to be an issue with my code-editor's auto-formatting/auto-arranging, have fixed up the spacing now.

magda-web-client/src/Components/Dataset/DatasetPage.scss
```diff
@@ -187,3 +187,9 @@
 .dataset-heading {
     margin-top: 10px;
 }
+
+.description-heading {
+    font-weight: bold;
+    font-size: 1.1rem;
+    display: block;
```
t83714, line 194: We probably remove the `display: block;` here and use `div` instead

magda-web-client/src/Components/Dataset/DatasetPage.tsx
```diff
@@ -147,16 +147,23 @@ const DatasetPage: FunctionComponent<PropsType> = (props) => {
...
-                        {dataset.contactPoint ? (
-                            <ContactPoint contactPoint={dataset.contactPoint} />
-                        ) : (
-                            <></>
-                        )}
+
                         {dataset.accrualPeriodicity ? (
-                            <div>Updated: {dataset.accrualPeriodicity}</div>
+                            <div>
+                                <span className={"description-heading"}>
```
t83714, line 153: probably should use "div" instead of "span"

magda-web-client/src/Components/Dataset/View/DatasetPageSuggestForm.js
```diff
@@ -83,29 +83,43 @@ export default class DatasetPageSuggestForm extends React.Component {
                 maxHeight: "95vh"
             }
         };
+        let contactPointMatch;
+        if (typeof this.props?.contactPoint === "string") {
+            const emailRegex = /[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}/gi;
```
t83714, line 88: That's my bad here. I think we probably can remove the `g` modifier so it only searches for the first occurrence. This change doesn't impact other code.

t83714 (changes requested):
@yayalu sorry I still missed a few minor things

magda-web-client/src/Components/Common/ContactPoint.js
```diff
@@ -27,27 +26,47 @@ class ContactPoint extends React.Component {
             <MagdaNamespacesConsumer ns={["datasetPage"]}>
```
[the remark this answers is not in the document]
yayalu: Do we still want to add this if we have the "{this.props.contactPoint ? (..." in line 31? We want to provide an alternate contact point text if the contact point doesn't exist, rather than remove text altogether.

t83714 (approved):
Tested in cluster~ Thanks again @yayalu 
merging

PR #3109, opened 2021-04-08
#3094 Make openfaas a dependecy of magda chart instead of magda-core

### LINKED ISSUE #3094: Helm will attempt to install CRD even when `global.openfaas.enabled` is set to false
### Helm will attempt to install CRD even when `global.openfaas.enabled` is set to false

Currently, when `global.openfaas.enabled` is `false`, the `openfaas` dependency is supposed to be turned off.

However, due to [this known bug](https://github.com/helm/helm/issues/8508) of helm, the CRD will still be installed even when dependency is turned off.

### Solution

We should remove `openfaas` as a dependency of `magda-core` chart and add it as the dependency of `magda` chart.
People who 100% don't want anything from `openfaas` (due to permission issue e.g.) can choose to depends on `magda-core` to remove `open-faas`

t83714:
### What this PR does

Fixes #3094

Make openfaas a dependecy of magda chart instead of magda-core

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
 -   Make `scssVars` option for content-api helm chart accept dictionary rather than JSON string
 -   Set default value of `enableMultiTenants` to `false` for `magda-core` chart
 -   Set gateway chart `enableAuthEndpoint` option default value to `true`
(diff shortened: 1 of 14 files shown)


t83714:
tested in dev cluster

PR #3112, opened 2021-04-11
#3111 Make SVG icons load via inline React Components rather than `<img/>`

### LINKED ISSUE #3111: SVG should be loaded via inline React Component In order to control its color via CSS
### SVG should be loaded via inline React Component In order to control its color via CSS

We have some SVG icon loaded via inline React Component recently.

e.g. https://github.com/magda-io/magda/blob/5dbbacc1b6253e77b74f40a61f41792c2d57101a/magda-web-client/src/Components/Dataset/Edit/Pages/EditFiles/SupercedeSelectionBox.tsx#L8

But some older codebase still load SVG via  tag which can't be themed by CSS.

We should at least change those "key" SVG icons to facilite some of current client project requirements.

This change should not bring any visual or functionality differences.

t83714:
### Make SVG icons load via inline React Components rather than ``

Fixes #3111

Make SVG icons load via inline React Components rather than `` so their color is changeable via CSS

### Checklist


t83714:
test locally.

Issue #3140, opened 2021-05-16
Ability to supply external CSS files to magda-web-client via Config


t83714:
### Ability to supply external CSS files to magda-web-client at runtime via Config

#3099 added the ability to supply external react component to replace built-in UI components of magda-web-client. 

Sometimes, it may be also required to supply external CSS files to magda-web-client at runtime via Config.

This ticket is about adding an "externalCssFiles" config option to our `web-server` helm-chart that allows users to supply a list of external css URLs.

t83714:
closed via https://github.com/magda-io/magda/pull/3141

Issue #3138, opened 2021-05-14
Allow Edit Dataset Button to be replaced by External UI Plugins


t83714:
### Allow Edit Dataset Button to be replaced by External UI Plugins

We have an Edit Dataset Button that currently allows Admin users to edit dataset.

We should allow Edit Dataset Button to be replaced by External UI Plugins in order to support the case that:
- we need external system to handle the edit dataset function
  - e.g. when the dataset is crawled from external system

### External UI Plugin Interface

Besides the common props below:
https://github.com/magda-io/magda/blob/4ec41254461f8b8515773dd3bf55b6aac6c2f138/magda-web-client/src/externalPluginComponents.ts#L13

We should expect the following props:
```typescript
interface {
  dataset: // the dataset data of the current dataset;
}
```

### Required Changes
- Load external plugin as "Edit Dataset Button" component when it exists
- Current "Edit Dataset Button" component should decides its visibility base on user's role in order to be compatible with the new interface 

Related ticket: https://github.com/magda-io/nsw-magda-plugin-ui-components/issues/1

t83714:
closed via https://github.com/magda-io/magda/pull/3141

PR #3034, opened 2020-11-17
Post message in terria v8 format.

mwu2018:
### What this PR does

Fixes #3019 

This PR should only be deployed after NationalMap is upgraded to terriajs v8. That is, this PR is blocked by https://github.com/TerriaJS/nationalmap/issues/968.

Before the `NationalMap` is upgraded to v8 `terriajs`, this [site](https://nationalmap.dev.saas.terria.io) can be used for testing.

### Checklist

-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable
-   [x] I've updated CHANGES.md with what I changed.

PR #3141, opened 2021-05-16
#3137 #3140 #3137 #3138

### LINKED ISSUE #3137: Allow Home Page URL to be configured for frontend UI
## Allow Home Page URL to be configured for frontend UI

We have a number of places on UI that will take users to the "Home Page". e.g:
- in Breadcrumbs
- Logo
- any redirection

Those home page urls will take users to Magda home page.

However, there might cases when we want to take user to other system's home page.

## Solutions

- Introduce optional `homeUrl`. 
  - When the value set, all home link / url should take user to that url instead

t83714:
### What this PR does

Supersede PR https://github.com/magda-io/magda/pull/3034

Fixes #3137 #3140 #3137 #3138

### Checklist


magda-web-client/src/Components/Common/DataPreviewMapOpenInNationalMapButton.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    {
+                        catalog: [
+                            {
+                                name: distribution?.title,
+                                type: "magda",
+                                recordId: distribution?.identifier,
+                                url: config.baseExternalUrl,
+                                id:
+                                    dga_id_prefix +
+                                    this.props.distribution.identifier
```
mwu2018, line 76: Minor improvement:
const identifier= dga_id_prefix + distribution?.identifier;
then use it in both catalog and workbench to avoid duplicates.

magda-web-server/src/index.ts
```diff
@@ -245,11 +245,43 @@ const argv = yargs
         type: "string",
         default: "Australia/Sydney"
     })
+    .option("homePageUrl", {
+        describe: "an alternative home page url.",
+        type: "string"
+    })
```
steve9164, line 251: Is it possible to set the default homePageUrl here rather than writing out the default in each place the homePageUrl is used?

magda-web-client/src/Components/Common/DataPreviewMapOpenInNationalMapButton.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    }
+                ]
+            };
+        }
+
+        if (!withoutBaseMap) {
+            //--- will not set baseMap if pass config by URL
+            catConfig.initSources[0].baseMapName = "Positron (Light)";
+        }
+        return catConfig;
```
steve9164, line 91: This looks good. Note that no version 8 map will work with IE11 any more. Maybe the map preview shouldn't be shown if the combination of IE11 and site using v8 preview map is detected.

magda-web-client/src/Components/Common/DataPreviewMapOpenInNationalMapButton.tsx
t83714: Good idea~ Updated

magda-web-server/src/index.ts
t83714: 👍  Good idea~ considering the local dev use case for web client, I made the default value resolved in "config.ts" in web-client module instead.

magda-web-client/src/Components/Common/DataPreviewMapOpenInNationalMapButton.tsx
t83714: Yeah make it only render the button when support v7 turned on or not IE 11

t83714:
tested in test cluster

PR #3143, opened 2021-05-27
Make authentication plugin work for logout process

### LINKED ISSUE #3136: Gateway should manage / coordinate logout process for Auth Plugins
### Gateway should manage / coordinate logout process for Auth Plugins

At this moment, we simply close Magda session when logout.

But sometimes we want to log user / out from identify provider as well when user choose to log out from Magda.

We should allow:
- Auth Plugin to provide a logout endpoint
- Auth Plugin can opt to save a logout redirection URL in session. e.g. `logoutUrl` when log user in
- When log out
  - Gateway should check if  `logoutUrl` exist.
    - If No, just close off Magda session only
    - If Yes:
      - close off Magda session only 
      - redirect user to `logoutUrl`
      - with optional `redirect` parameter set to post logout redirect URL

t83714:
### Make authentication plugin work for logout process

Fixes #3136

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,1238 +2,1239 @@
 
 ## 0.0.60
 
--   Updated dataset pages and Ask A Question options for datasets with no contact points
--   Updated Suggest A Dataset form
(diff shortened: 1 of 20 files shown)


t83714:
tested with test cases only but need a urgent release

Issue #3157, opened 2021-06-15
Dataset result page might freeze on google chrome due to HTML description content


t83714:
## Dataset result page might hang on google chrome due to HTML description content

Currently, dataset html description content will be displayed as it is on dataset search result page.

For complex HTML content, it might freeze on google chrome during the rendering.

We should strip HTML tags before rendering html content on dataset result page.

t83714:
close via https://github.com/magda-io/magda/pull/3158

PR #3158, opened 2021-06-17
Issue/3156 & 3157

### LINKED ISSUE #3156: More accurately report original dataset source on web UI
# More accurately report original dataset source on web UI

As per issue: https://github.com/magda-io/magda-ckan-connector/issues/4, some datasets crawled from external portal might be from other source.

We need to identify the original source when it's available:
- Modify "source" aspect and add optional fields:
  - originalUrl: the url of the dataset source data on the original data portal (normally should be API data source)
  - originalName: the name of the original data portal
- The dcat-dataset-string aspect `landingPage` should also be updated to reflect this when the dataset is harvested from other external sources
- When displaying dataset source on dataset details page, we will use originalUrl field when it's available.
  - otherwise, will use landingPage field

t83714:
### What this PR does

Fixes #3156 #3157 #3159

### Checklist


t83714:
Tested locally

Issue #3160, opened 2021-06-17
Add html_strip character filter to description field Analyser in elasticsearch


t83714:
## Add html_strip character filter to description field Analyser in elasticsearch

Text content of description field might contain with HTML tags.

We should Add html_strip character filter to description field indexing Analyser in elasticsearch so it won't interference the searching.

t83714:
close in favour of https://github.com/magda-io/magda/issues/3155

t83714:
closed via https://github.com/magda-io/magda/pull/3158

PR #3162, opened 2021-06-18
Issue/3155 & 3160

### LINKED ISSUE #3155: Indexing Error: description field oversize
### Indexing Error: description field oversize

The [magdaSynonymTextField](https://github.com/magda-io/magda/blob/b5c4b5d5bfc849421a0aa40be9bd1e9b8f9b4848/magda-scala-common/src/main/scala/au/csiro/data61/magda/search/elasticsearch/IndexDefinition.scala#L54) function that we use for setting indexing fields for "description" will auto add an "keyword" field.

However, the keyword field has `32766` limitation --- see error below:

```
[WARN] [06/10/2021 23:54:31.707] [indexer-akka.actor.default-dispatcher-11056] [akka.actor.ActorSystemImpl(indexer)] Failure when indexing ds-nsw-ckan-334f67b9-2055-4119-9349-33d29c7550ee: Some(BulkError(illegal_argument_exception,Document contains at least one immense term in field="description.keyword" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped.  Please correct the analyzer to not produce such terms.  The prefix of the first immense term is: '[60, 112, 32, 115, 116, 121, 108, 101, 61, 39, 109, 97, 114, 103, 105, 110, 45, 116, 111, 112, 58, 48, 99, 109, 59, 109, 97, 114, 103, 105]...', original message: bytes can be at most 32766 in length; got 32936,null,0,null))
```

We actually didn't use "description.keyword" for searching (except [autoCompleteQuery](https://github.com/magda-io/magda/blob/b5c4b5d5bfc849421a0aa40be9bd1e9b8f9b4848/magda-search-api/src/main/scala/au/csiro/data61/magda/search/elasticsearch/ElasticSearchQueryer.scala#L175) function might allow user query a user specified field).

We should modify `magdaSynonymTextField` to remove the keyword field that created for "description"

~To avoid the trouble of bumping index version, the following solution will be adopted:~
~- Indexer will use [jsoup](https://jsoup.org/) to remove any HTML tags~
~- If still oversize, cut it off at 32766 byte~

Update:
Make indexer to strip html tags requires a lot changes to test cases.

Will add strip html tags function as elasticsearch analysis filter

t83714:
### What this PR does

Fixes #3155 #3160 
- Indexing Error: description field oversize
- Add html_strip character filter to description field Analyser in elasticsearch

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,8 @@
 - #3156 web UI: display original source name when it's available
 - #3157 strip html tags from dataset description on dataset result page
 - #3159 Upper case first char of text display for "Update Frequency", "Tags" & "Themes" on dataset details page
+- #3155 remove "keyword type" defined for "description" fields to avoid "term" oversize
+- #3160 Add strip_html filter to "description" fields for elasticsearch indexing analyzer
 
 ## 0.0.59
 
--- a/magda-indexer/src/main/resources/application.conf
+++ b/magda-indexer/src/main/resources/application.conf
@@ -3,6 +3,10 @@ http {
   port = 6103
 }
 
+time {
+    defaultOffset = "+10:00"
+}
+
 # we need to have this section in every module
 # Plus, logback.xml in resource of every module
 # Put into common module since not working properly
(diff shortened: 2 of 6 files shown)


t83714:
tested in local cluster

PR #3187, opened 2021-07-14
#3170 Remove unused old admin UI endpoint from Gateway

### LINKED ISSUE #3170: Remove unused old admin UI endpoint from Gateway
### Remove unused old admin UI endpoint from Gateway

#3038 has remove all ejb view files & render engine from gateway (as part of effort of reducing gateway's dependencies and functionality-wise "pure")

However, we still haven't remove the HTTP access point for those UI here:

https://github.com/magda-io/magda/blob/cdebec35af7df68b81fee59118a7e9f746947c03/magda-gateway/src/createAuthRouter.ts#L101

Thus, when user access those endpoint, he will get a 500 error rather than 404.

We should remove those endpoint from gateway.

t83714:
### What this PR does

Fixes #3170

Remove unused old admin UI endpoint from Gateway

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,7 @@
 - Upgrade format & rating minion to 1.0.0, Preview Map to 1.0.1
 - #3179 Fixed Invalid HTML in description make Google Chrome Unresponsive
 - Further simplify the error message from map preview module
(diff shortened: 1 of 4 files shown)


t83714:
tested in dev cluster

PR #3189, opened 2021-07-15
#3183 Make search input accept empty string as search query string

### LINKED ISSUE #3183: Making UI accepts the empty string to return all datasets
## Making UI accept empty string and returns all datasets

Currently, if you want to match / returns all datasets, you need to input a single character `*`.

It might be counter-intuitive for some users.

We should allow users listing all datasets by inputting an empty string.

t83714:
### What this PR does

Fixes #3183 Make search input accept empty string as search query string

Also:
- upgrade minion / connector versions in local deployment charts
- consolidate local deployment chart values files

### Checklist

-   [x]  unit tests aren't applicable


t83714:
seems not work on home page by press enter key.

t83714:
tested in test cluster & local

Issue #3193, opened 2021-07-20
Registry PUT or POST request should auto filter \u0000 from JSON data


t83714:
## Registry PUT or POST request should auto filter \u0000 from JSON data

See #3161

This PR https://github.com/magda-io/magda/pull/3194 only addressed the issue at frontend.

This ticket is about a more generic fix that implemented in registry API:
- `\u0000` in all incoming request JSON data should be auto filtered out
- test cases should be in place

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3525

PR #3194, opened 2021-07-20
#3161 UTF-16 encoded CSV causes "unsupported Unicode escape sequence" error when saving a dataset draft

### LINKED ISSUE #3161: \u0000 needs to be removed from JSON all data
## \u0000 needs to be removed from JSON all data

Some users see an error message `unsupported Unicode escape sequence, \u0000 cannot be converted to text` while input data to registry.

This is due to design change in postgresql v9.4 (GPDB v6). There is a postgresql community post which talked about this change, a quote was selected from the community post, pasted below:

Source:
https://www.postgresql.org/message-id/8239.1500489054%40sss.pgh.pa.us

> Matthew Byrne writes: Are there any plans to support \u0000 in JSONB and, relatedly, UTF code > point 0 in TEXT?

> No. It's basically never going to happen because of the widespread use of C strings (nul-terminated strings) inside the backend. Making \0 a legal member of strings would break all those internal APIs, requiring touching far more code than anyone would want to do. It'd likely break a great deal of client-side code as well.
 
> The GPDB R&D team also confirmed, even in postgresql v12, this usage will report the same error. The R&D team has no plans to change this design.

[a possible workaround](https://community.pivotal.io/s/article/loading-data-report-error-unsupported-Unicode-escape-sequence?language=en_US):
```
SELECT (regexp_replace(the_string::text, '
u0000', '', 'g'))::json;
```

We need to sanitise all json input to postgresQL DB.

t83714:
### What this PR does

Fixes #3161 UTF-16 encoded CSV causes "unsupported Unicode escape sequence" error when saving a dataset draft

We only addressed this issue at frontend.

We will look for a more generic fix in ticket #3193

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -68,6 +68,7 @@
 - #3186 Fixed: avoid extra slash in inqury email dataset URLs
 - #3188 Publish openfaas chart & Move openfaas helm chart out of main repo
 - #3183 Make search input accept empty string as search query string
+- #3161 Fixed: UTF-16 encoded CSV causes "unsupported Unicode escape sequence" error when saving a dataset draft
 
 ## 0.0.59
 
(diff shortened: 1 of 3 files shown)


t83714:
tested locally

PR #3200, opened 2021-07-27
Fixes #3198 Indexer will by default auto reindex / trim obsolete publisher / format weekly

### LINKED ISSUE #3198: Add default indexer crawling job
# Add default indexer crawling job

At this moment, publisher & format indices in elastic search is created as part of dataset indexing process.

because of it, non-exist publisher or format records will not be automatically removed during the event processing (as unless all datasets are not linked to the publisher or format record anymore, we can't delete it)

At this moment, the trimming process can be triggered by a POST request to indexer's `reindex` endpoint.

To save the effort of triggering it from time to time, we should add a default job to trigger it:
- User can turn it off via helm config
- The job schedule can be config
- By default, run once a week on Saturday night.

t83714:
### What this PR does

Fixes #3198 Indexer will by default auto reindex / trim obsolete publisher / format weekly

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -74,6 +74,7 @@
 - Improve CSV visualisation tool date columns recognition logic
 - #3161 Fixed: UTF-16 encoded CSV causes "unsupported Unicode escape sequence" error when saving a dataset draft
 - Upgraded format-minion to v1.0.1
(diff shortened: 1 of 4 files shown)


t83714:
tested in local cluster

Issue #3256, opened 2021-10-28
Unified policy for secure both registry records & Elasticsearch datasets


t83714:
We currently have a [dataset policy](https://github.com/magda-io/magda/blob/3f9c013273176823566d8c862e8d2a2ca556feb2/deploy/helm/internal-charts/opa/policies/object/dataset/allow.rego) that secures dataset records only.

The policy was created for early trial of our access control model and only for purpose of securing dataset in elastic-search originally.

However, our main metadata store `registry` actually support a more generic `record` type which users can create other record types on top of it. e.g. `distributions`, `organisations` etc. 

This ticket is about creating one single "unified story" to:
- secure the generic record model including different record types
- secure two storage engines: registry & elastic search & different use cases
- the solution should be expansible so users can add more policy to add custom logic for custom record type

### Solution
- One policy entrypoint `object/record/allow.rego` to delegate auth decision to policy created specifically for a particular record type. 
  - e.g. when a record is a dataset record (test by checking if a record has [`dcat-dataset-strings` aspect](https://github.com/magda-io/magda/blob/master/magda-registry-aspects/dcat-dataset-strings.schema.json)), delegate the decision to `object/dataset/allow.rego`
- if it's an unknown record type, default access control logic applied.
  - i.e. Only users have matching permission with appropriate constraint can perform the operation.  
  - e.g. if a user has `object/record/update` permission with `owner` constraint, only the record owner/creator can update the record.
-  Existing policy `object/dataset/allow.rego` needs to be adjusted to match more generic registry API data structure.
  - Policy enforcement at Search API / Elasticsearch can use extra logic to translate the data field reference that matches registry data structure to data field references matches Elasticsearch data structure.
  - i.e. in policy we should access aspect data field with json patch `input.object.dataset.[aspect name].[aspect field name]`. 
  - e.g. reference `input.object.dataset.publishingState` in previous policy should change to `input.object.dataset. publishing.state` as `publishingState` field is defined as `state` field in [publishing aspect](https://github.com/magda-io/magda/blob/master/magda-registry-aspects/publishing.schema.json)
  - e.g. `input.object.dataset.accessControl.ownerId` reference should be `input.object.dataset["access-control"].ownerId` --- as `ownerId` field us defined in [access-control](https://github.com/magda-io/magda/blob/master/magda-registry-aspects/dataset-access-control.schema.json) aspect.
[...]

Issue #3250, opened 2021-10-18
Remove OPA AST parser in scala codebase & use decision endpoint instead


t83714:
#3253 offers a new opa decision endpoint thats response simple & concise response.

We should remove OPA AST parser in scala codebase and use this new decision endpoint instead.

Things to do:
- [x] Add generic registry `record` type policy entry point
- [x] Remove OPA parser for registry API
- [x] Rewrite decision enforcement logic for registry API
- [x] Remove OPA parser for search API
- [x] rewrite decision enforcement logic for search API

t83714:
done

Issue #3069, opened 2021-01-20
OPA based Access Control improvement


t83714:
### OPA based Access Control improvement

- Re-write OPA AST Parser (especially for scala code base) 
- make access control more generic and extensible (able to define/support new operations)

PR #3257, opened 2021-10-28
issue/3256: Unified policy for secure both registry records & Elasticsearch datasets

t83714:
### What this PR does

- #3256
- #3304
  - All registry APIs with all operations are secured (including web hook interface)
    - depends on resource type. available operations could be common ones like create, update, read, delete or (`ack` for web hook ) etc.
  - Role permission (e.g. admin role) are not hard coded anymore
  - Still need secure dataset searching interface (search API)
  - It should still work with existing internal communication code as we used to use a built-in admin account and admin users are by default have access to any resources.
- also #3250
- also #3303 rename aspect `dataset-access-control` to `access-control`
- remove old auth test cases add new auth test cases to all registry endpoints

They are both part of EPIC:
- #3069

### Checklist


t83714:
Have various unit tests to cover all endpoints

PR #3268, opened 2021-11-19
Fix sbt m1

t83714:
### What this PR does

- Fix scala sbt issue for apple m1 users
- Fix: correct db backup job default schedule to 0 15 * * 6
- All builder docker image now has `bash` as entrypoint (instead of node)

### Checklist


t83714:
tested in dev clustered & locally

Issue #3294, opened 2022-01-08
Indexer should auto fix non-topologically closed Polygons / MultiPolygons


t83714:
### Indexer should auto fix non-topologically closed Polygons

Polygons should be topologically closed. i.e. the last point should be a repeat of the first point.

However, the data we received might not be always valid.

e.g. 
```json
{"type": "Polygon", "coordinates": [[[113.58154999990352, -30.81596666648388], [115.19059913517697, -30.81596666648388], [115.19059913517697, -28.451649998896013], [113.58154999990352, -28.451649998896013]]]}
```

For this case, we should auto-fix polygon by copy the first point & insert into the coordinate list:

```json
{"type": "Polygon", "coordinates": [[[113.58154999990352, -30.81596666648388], [115.19059913517697, -30.81596666648388], [115.19059913517697, -28.451649998896013], [113.58154999990352, -28.451649998896013],[113.58154999990352, -30.81596666648388]]]}
```

t83714:
closed via https://github.com/magda-io/magda/pull/3297

Issue #3293, opened 2022-01-07
Upgrade elasticsearch to 6.8.22


t83714:
### Upgrade elasticsearch to 6.8.22

Upgrade to the latest v6 version elastic search for better support.

The upgrade will be part of the "next" milestone and released together with other v2 changes (e.g. adv access control)

t83714:
closed via https://github.com/magda-io/magda/pull/3297

PR #3297, opened 2022-01-14
Upgrade es 6.8

t83714:
### What this PR does

Fixes:
- #3293
- #3294  
- Upgrade logback to 1.2.10
- Crawler will now wait for 3 seconds (used to be 1 second) before perform trim action to avoid timeout issue
- Refactor Indexer code to remove the redundant index queue

### Checklist


t83714:
tested in dev cluster

PR #3300, opened 2022-01-18
use data61/elasticsearch:6.8.22 image in docker compose file

t83714:
### What this PR does

Update the docker image that is used for running test case (and local tests) to a muti-arch version image.

### Checklist


t83714:
all test cases passed

PR #3311, opened 2022-02-14
Some generic homepage visual adjustment

t83714:
### What this PR does

Some generic homepage visual adjustment.
- Make home tiles 2 columns layout (instead 3 tiles with various sizes)
- Keep original image size when uploading background images from the admin panel
- CSS fixes to make sure the background images original height is kept but not beyond the content length.

Those changes are introduced based on DGA feedback but are generic in nature.

### Checklist


t83714:
tested on test site

Issue #3315, opened 2022-02-28
Update integration test cases for endpoints secured with the new auth model


t83714:
### Update integration test cases for endpoints secured with the new auth model

We have all registry APIs secured with the new auth model. 

PR: https://github.com/magda-io/magda/pull/3257

But the integration test cases haven't been updated.

This ticket is about updating integration tests to match the new auth models

t83714:
closed via pr: https://github.com/magda-io/magda/pull/3344

PR #3344, opened 2022-04-14
Issue/3315 New Integration test facility & add integration tests for auth

t83714:
### What this PR does

Fixes
- #3315 
- Add Service Runner to make it easier to launch magda services for testing / local dev
- #3345
  - API added but still needs to expose via gateway & add auth to existing APIs (previously are internally only APIs)

### Checklist


t83714:
tested in pipelines & locally

PR #3357, opened 2022-05-17
Issue/3356 Add User API key related APIs and more

### LINKED ISSUE #3356: Add User API key related APIs and more
### Add user API key related APIs

At this moment, we don't have APIs for managing users' API keys (e.g. create/list API keys).

We only offer a [CMD utility script](https://github.com/magda-io/magda/blob/master/scripts/create-api-key.js) for creating users' API keys.

As we plan to build a new admin UI with API keys management function, we will need to expose user API key management functions via APIs.

#### Acceptance criteria
- Add the following APIs to auth APIs services
  - List all API keys of a user
  - create a new API key for the user
  - revoke an API of a user
- add a new column `expiry_time`.
  - If a time is set, the API key is only valid before the `expiry_time`
  - need to update API `/v0/private/users/apikey/:apiKeyId` to make sure it only accepts non-expired API keys 
- add new columns `last_successful_attempt_time` & `last_failed_attempt_time`
  - the API `/v0/private/users/apikey/:apiKeyId` will attempt to record the `last_successful_attempt_time` after successfully valid the API key and `last_failed_attempt_time` for failed attempts.
  - The update should be non-blocking updates. any errors will be printed to the server log only.
- Auth API should support a new config option: `failedApiKeyAuthBackoffSeconds`
  - any API key validation request will fail at once before the number of `failedApiKeyAuthBackOffSeconds` passed since the last fail attempt for the same API key.
- add a new column `enabled`
  - by default, it should be "True".
  - the API key is only valid when this field is true.
  - a user can choose to temporarily disable the API key by setting it to False

t83714:
### What this PR does

Fixes 

- #3356

### Checklist

-   [x] Manually tested but yet need full set of test cases

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,7 @@
 - Add `debug` switch to registry API, search API & Auth API
 - Related to #3331, remove any logic related to obsolete `authnReadPolicyId` record field
 - Related to #3331, remove obsolete policy files `object/registry/*`
+- Added APIs for managing API keys
 
 ## 1.3.0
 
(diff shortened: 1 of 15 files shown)


t83714:
tested in dev cluster

Issue #3363, opened 2022-06-14
upstream bitnami helm chart repo index truncated


t83714:
### upstream bitnami helm chart repo index truncated
See https://github.com/bitnami/charts/issues/10539
bitnami helm chart repo introduced a new index.yaml retention policy.
To access the older version helm chart, we need to:
- set repo to: `repository: https://raw.githubusercontent.com/bitnami/charts/pre-2022/bitnami`

This issue is only related to `next` (v2) branch

Issue #3362, opened 2022-06-14
Set "cluster-autoscaler.kubernetes.io/safe-to-evict" : true annotation to pods with local storage


t83714:
### Set "cluster-autoscaler.kubernetes.io/safe-to-evict" : true annotation to pods with local storage

Pods with local storage (e.g. emptyDir volume) can prevent cluster autoscaler from evicting nodes.
see [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node)

We should add `"cluster-autoscaler.kubernetes.io/safe-to-evict" : true` annotation to those pods with local storage (`emptyDir` volume).

We need to an extra release for v0.0.66, v1 & v2.

Issue #3269, opened 2021-11-20
Create `Group` concept as both a container that organises datasets and an access control toolkit for some use cases


t83714:
### Motivation

The `Group` concept can be used as:
- a container that organises datasets in meaningful ways. Use cases could be: 
  - `dataset categories`
  - `manually curated dataset catalogue
 - an access control toolkit for some use cases. e.g. : 
   - `dataset projects`: a group of people from cross-functional teams work on a project that involves access to a list of private datasets
  - Alternative way of controlling dataset visibility from dataset's side and generic & flexible implementation of [dataset access level feature](https://github.com/magda-io/magda/blob/1e101f1b58fd12f515e0efe68c558b1a77519395/magda-registry-aspects/publishing.schema.json#L13)

### Technical Proposal V2

> rename `group` to `access-group-group` and `group-details` aspect to `access-group-details` to avoid clash with TerriaMap group aspect definition.

- `access-group` (or `group` for short) is represented by registry records with `access-group-details` & [access-control aspect](https://github.com/magda-io/magda/blob/next/magda-registry-aspects/access-control.schema.json)
- From the `access-control aspect`, we can determine who has access to the `access-group` record
  - The creator will be granted full permission to the `access-group` record via `pre-Authorised Permission`
- `access-group-details` contains the following fields:
  - `name`: string; 
  - `description`: string.
  - `resourceUri`: One access group can only grant access to the same type of resources. The Resource URI specifies the type of resources that the access group manages. At this moment, only registry records `object/record` is supported.
  - `keywords`: string[].
  - `permissionId`: string. 
    - When a `access-group` is created, a permission record with [pre-authorised constraint](https://github.com/magda-io/magda/blob/next/docs/docs/architecture/Guide%20to%20Magda%20Internals.md#permissions--roles) will be created.
    - This permission specifies all operations that group user members are allowed to perform on datasets that belong to this `access-group`.
    - When a dataset is added to this group, the id of the permission should be added to the dataset's `access-control` aspect, `preAuthorisedPermissionIds` field. 
    - Group membership for datasets can also be determined by looking at datasets' `access-control` aspect, `preAuthorisedPermissionIds` field. i.e. only datasets who has the group's `permissionId` assigned to `access-control` aspect, `preAuthorisedPermissionIds` field, will be members of the group.
  - `roleId`: string.
    - When a group is created, a role record will be auto-created with the name in the pattern of `auto-role for group xxx`.
[...]

PR #3365, opened 2022-06-22
Issue/3269 Backend Support Access Group

t83714:
### What this PR does

> This PR is mainly for adding backend API support for the access group proposal [here](https://github.com/magda-io/magda/issues/3269). Without those extra APIs, the implementation of the access group is still possible (after all, the access group feature is built on top of the existing role, permission & registry record facility). Having said that, the extra access group related APIs will help to reduce the implementation complexity at the frontend.

Changes Summary:
- #3269
  - also add GET `/auth/roles/:roleId/users` & GET `/auth/roles/:roleId/users/count` APIs
- #3366
- Fixed: `set-scss-vars` script will use proper imagePullSecrets to generate the UI CSS update job
- Allow the `no cache` behaviour of `whoami` & `getUserById` api to be turned off
- refactor code related to create / update permission

Merged Changes from v1:
- #3362
- #3367
- #3369
- #3371
- #3363
- Upgrade openfaas helm chart to 5.5.5-magda.2 to be compatible with k8s 1.22

### Other Info

This PR doesn't include frontend UI --- although we have a closed source version access group UI built & leverage the new backend capability created by this PR.

### Checklist

### DIFF
--- a/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsService.scala
+++ b/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsService.scala
@@ -841,7 +841,6 @@ class RecordsService(
     * @apiParamExample {json} Request-Example
     * {
     *   "recordIds": ["dsd-sds-xsds-22", "sds-sdds-2334-dds-34", "sdds-3439-34334343"],
-    *   "aspectId": "access-control",
     *   "jsonPath": "$.preAuthorisedPermissionIds",
     *   "items": ["b133d777-6208-4aa1-8d0b-80bb49b7e5fc", "5d33cc4d-d914-468e-9f02-ae74484af716"]
     * }
--- a/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsService.scala
+++ b/magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsService.scala
@@ -824,6 +824,125 @@ class RecordsService(
   }
 
   /**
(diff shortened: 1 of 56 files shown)


magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsService.scala
```diff
@@ -571,6 +571,377 @@ class RecordsService(
...
+    *
+    * @apiParam (path) {string} aspectId the id of the aspect to be updated
+    * @apiParam (body) {string[]} recordIds a list of record IDs of records to be patched
+    * @apiParam (body) {string} jsonPath the jsonPath string that is used to locate the json array in the record aspect data
+    * @apiParam (body) {any[]} items a list of items to be removed from the located array. The type of the items can be either string or number.
+    *
+    * @apiParamExample {json} Request-Example
+    * {
+    *   "recordIds": ["dsd-sds-xsds-22", "sds-sdds-2334-dds-34", "sdds-3439-34334343"],
+    *   "aspectId": "access-control",
```
steve9164, line 844:
```suggestion
    *   "aspectId": "access-control",
```

Should `aspectId` be in the example request body?

t83714: Thanks it was mistake in API docs --- the aspectId is supplied via path segment rather than body

t83714:
tested in the dev cluster (v2.0.0-alpha.5 release) via catalog-manager deployment

Issue #3349, opened 2022-05-03
Secure content API with new auth model


t83714:
### Secure content API with new auth model

- Remove usage of `mustBeAdmin` middleware [here](https://github.com/magda-io/magda/blob/f8ada91fdcf1875002a02c03fec33ef929bc35aa/magda-content-api/src/createApiRouter.ts)
- Add back on fine-gained access control based on user's permission rather than roles

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3387

PR #3387, opened 2022-08-02
Issue/3349 secure content API with new auth model

t83714:
### What this PR does

Fixes 
- #3349

also:
- dropped support for legacy `csv/*` content items 
  - should use storage API instead
- add `skipAuth` & `debug` config to content API helm chart
- create `object/content` resource in SQL upgrade file
  - add `object/content/update` (content API has only one interface for both create & update), `object/content/read` & `object/content/delete` operations
  - assign `object/content/read` to anonymous user & authentication users

### Checklist


t83714:
tested in dev cluster

Issue #3343, opened 2022-04-04
Consolidate admin/settings UI functionalities


t83714:
### Consolidate admin/settings UI functionality

With the following admin/settings UI functionalities scatter over different places:
- Create Dataset / Metadata
- Admin Web UI for auth management #3306
- Access Group Management #3269 
  - This feature has been implemented as part of an internal project. Will not be available in the open-source codebase for now.
- User's own profile page
We probably want to figure out a way to re-organise those UI functionalities entry points into e.g. a single side menu system. 

> update: the above feature will be moved into a central "Settings" area (accessible from the header menu once logged in). 
> Legacy admin panel (for managing content e.g. headers, footers, logos, style & home page etc) will be moved to "Others" section of this "Settings" area as well

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3393

Issue #3306, opened 2022-01-27
Admin Web UI for auth management


t83714:
### Admin UI for auth management

We need an Admin Web UI for auth management, including:
- define / update `resources`
- define / update `operations`
- create / update / assign `permissions`
- create / update / assign `roles`
- create / update `user`
- create / update `API Keys`
- create / update `user crendantails` (only when internal auth plugin is enabled)

This web UI is not only useful for future users but would also help a lot during the development & testing. e.g. trials on different auth models

t83714:
A full functional UI has been done in a separate closed source project.
On the open source side, we will focus on #3343 for v2 release

PR #3393, opened 2022-08-10
Issue/3343 Consolidate admin/settings UI functionalities

t83714:
### What this PR does

Fixes 
- #3343 the following feature will be moved into a central "settings" area (accessible from header once logged in)
  - Create Dataset / Metadata
  - Admin Web UI for auth management #3306
  - Access Group Management #3269 (UI part is not available in open source code base. Backend APIs have full features )
  - User's own profile page (will be the landing page after the logged in)
  - The legacy admin panel (for managing content e.g. headers, footers, logos, style & home page etc) will be moved to "Others" section of this "Settings" area as well
- [fixed: publishingState filter is not processed probably in search API](https://github.com/magda-io/magda/pull/3393/commits/c71a609432cd897241683fdcf5615c98f95e90b7)
- Remove legacy code
- Fixed: dataset metadata editor incorrectly fetch records with cache
- #3395
- make dataset management page "draft" / "published" tab can be navigated via url path
- make sure file handling edging cases can be handled correctly. 
- fixed an issue of faas function invocation: metadata pre-fill feature should only call `data-url-processor` & `api-url-processor` type faas functions
- fixed indexer API request wrong permission
- Make source, publishing & "information-security" aspects available from distribution records as well
- remove 'object/dataset/published/unpublish' & 'object/dataset/published/unpublish' defined in schema version 1.3 for now, to be reintroduced in future as part of the publishing approval process
- Managing draft & published distribution separate like dataset records & update policies accordingly
- Allow changing dataset owner in the editing view
- make UserNameLabel use cache

### Checklist


t83714:
tested in dev cluster

PR #3407, opened 2022-09-14
Login workflow improvement

t83714:
### What this PR does

- Use global state instead in ValidateUser
- Periodically refresh login status (Interval configurable via web-server helm chart)
- Allow setting landingPage uri for anonymousUser and authenticatedUser
- Logout action will redirect user to LandingPage

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@
 
 ## v2.0.2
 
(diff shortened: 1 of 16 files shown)


t83714:
tested in dev cluster

PR #3418, opened 2022-10-31
Upgrade minion connector versions

t83714:
### What this PR does

- Upgrade default connectors & minions version (to v2.0.0)
- Fixed indexer error: distribution/byteSize field should be in `long` type to support possible larger numbers
- Upgraded indexer weekly reindexer trigger container to node14
- Distribution page UI minor improvements: display file size (when available) & adjust margins between information blocks

### Checklist


t83714:
tested in dev cluster

PR #3423, opened 2022-11-07
#3421 Improve dataset updating UI logic using `merge` option of the PUT record API

### LINKED ISSUE #3421: Dataset Metadata Editor: Better records updating logic
We have made some improvements to the records updating logic in ticket: 
- #3412

The previous includes some improvement to the records updating logic at frontend.

However, pure frontend effort won't be able to address all edging cases.

e.g. Because the current `PUT /api/v0/registry/records/{recordId}` only supports the `replace` behaviour, it requires API invoker always to include all record data (including `access-control-aspect`).

This might not always be feasible when you only want to update certain fields of the aspect without touching other fields.

Once this ticket:
- #3420
is done, we will be able to further improve the records updating logic to address all edging cases.

> Changes should be properly test to make sure issue #3412 will not break again

t83714:
### What this PR does

Fixes #3421 Improve dataset updating UI logic using `merge` option of the PUT record API

### Checklist

-   [ ] unit tests aren't applicable


t83714:
tested in dev cluster

PR #3460, opened 2023-05-05
Fixed an config field issue that could prevent inquiry form from functioning

t83714:
### What this PR does

Fixed an config field issue that could prevent inquiry form from functioning

CleanShot 2023-05-05 at 11 26 26@2x

### Checklist


t83714:
tested in test cluster

Issue #3478, opened 2023-08-13
New aspect "preview-map-settings" to provide dataset / distribution level controls over UI functionality


t83714:
### New aspect "preview-map-settings" to provide dataset / distribution level controls over UI functionality

Currently, we have config options available for both [web-server](https://github.com/magda-io/magda/tree/main/deploy/helm/internal-charts/web-server#values) & [magda-preview-map](https://github.com/magda-io/magda-preview-map#values). However, those config options only allow us to config the preview map global behaviour rather than the behaviour for each individual map.

This ticket proposes a new aspect, "preview-map-settings":
```json
{
  "$schema": "http://json-schema.org/schema#",
  "title": "Preview Map Settings",
  "description": "Provides extra preview map configuration options at dataset/distribution level.",
  "type": "object",
  "properties": {
    "enable": {
      "title": "Whether the preview map UI feature should be enabled for the attached dataset/distribution.",
      "type": "boolean",
      "default": true
    }
  }
}
```

We can manually attach this aspect to any dataset/distribution records to indicate whether the whole dataset (all distributions in the dataset) or the attached distribution record should not be used for preview map purposes in UI.

- The default value of the `enable` field is `true`.  For all datasets or distributions that don't have this aspect attached, we assume they are allowed to be previewed in the map by default.

At this stage, we will not provide any UI / tool to create this aspect. Users can create the aspect data for a particular dataset or distribution record via API when required.

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3479

Issue #3477, opened 2023-08-04
decision endpoint should make decision based on anonymous user's data when the user doesn't exist 


t83714:
### decision endpoint should make decision based on anonymous user's data when the user doesn't exist 

Currently, when the decision endpoint is asked for making make a decision for a non-exist user, it will throw an error:
```
 withAuthDecision middleware error: Error: Failed to retrieve auth decision for operation `object/content/read` from policy engine: Not Found User                                                                                 │
│ Failed to retrieve auth decision for operation `object/content/read` from policy engine: Not Found User  
```

For this rare case, it would be more appropriate to make decision endpoint make the decision assuming it's an anonymous user.

PR #3479, opened 2023-08-17
Issue/3478 and other v2.2.6 changes

t83714:
### What this PR does

- Upgrade magda-format-minion to v2.0.1
- #3478 Added the support of the new preview-map-settings aspect
- Move format `ESRI FEATURESERVER` under `csv-geo-au` in the default map preview format perference list
- Fixed: registry records editor UI doesn't encode long record ID correctly
- Provides HTTP header name information on user api key creation screen
- #3477: decision endpoint should make decision based on anonymous user's data when the user doesn't exist

### Checklist


t83714:
released & tested in dev cluster

Issue #3100, opened 2021-03-31
[connector-sdk] Allow runtime filter code to be supply to connectors via Helm config


t83714:
### Motivation

- We currently support filtering crawling result by harvest sources & organisations. However:
  - These are CKAN connector only features
  - There are situations that we need to filter out records by more complex logic rather than just harvest sources & organisations

One use case is to filter out VIC datasets from AURIN CSW endpoints:

https://openapi.aurin.org.au/public/csw?service=CSW&version=2.0.2&request=GetRecordById&elementsetname=full&outputschema=http%3A%2F%2Fwww.isotc211.org%2F2005%2Fgmd&typeNames=gmd%3AMD_Metadata&id=aurin%3Adatasource-VIC_Govt_DELWP-VIC_Govt_DELWP_datavic_ENERGY_ELECTRICITY_DISTRIBUTOR

There isn't any org info or harvest info available --- we probably need to filter by logic "any datasource id starts with `VIC_`"

### Proposed Solutions

Similar to recent introduced [Preset Aspect Feature](https://github.com/magda-io/magda/blob/57d5c8a98086fd9eebe9cd508cbc99ee3e10fec9/magda-typescript-common/src/JsonConnector.ts#L729) that applies to all our connectors via our connector SDK, we should introduce a feature that allow a flexible filter definition can be configured through connector helm config and executed by connector SDK core.

The filter definition will be JS code that return true or false based on pre-existing variables in the scope:
- ~record: magda record includes all its aspects that are about to create.~
- jsonData: the rawJson data retrieved by connector directly from remote API
- recordType: a string represent the type of the record. e.g. "Organization" | "Dataset" | "Distribution" ~| "Group"~

The js code supplied will be run in a scoped sandbox.

As the changes is done on conenctor-sdk, all connectors are only required to upgrade to the latest SDK to order to access this feature.

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3489

PR #3490, opened 2023-10-04
#3100: Conenctor SDK: Allow runtime filter code to be supply to connectors via Helm config

t83714:
### What this PR does

Fixes:
- #3100

### Checklist


t83714:
tested in dga test cluster

Issue #3511, opened 2024-03-26
Map Preview doesn't select the default feature layer correctly when it's supplied via query parameter


t83714:
### Map Preview doesn't select the default feature layer correctly when it's supplied via query parameter

e.g. for the distribution below the feature layer "summary-occ" should be auto-selected but doesn't:

![CleanShot 2024-03-26 at 14 39 45@2x](https://github.com/magda-io/magda/assets/674387/8a960984-a04f-4c71-a3f6-db94ea53e8f4)

#### Fixes

The `in` should be `of` in this line: https://github.com/magda-io/magda/blob/922f6a82fc21a17d90141a066cbbf0ab2d2ea39b/magda-web-client/src/Components/Common/DataPreviewMap.tsx#L239

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3512

PR #3512, opened 2024-03-26
#3511 fixed Map Preview doesn't select the default feature layer correctly when it's supplied via query parameter

t83714:
### What this PR does

Fixes:
- #3511 fixed Map Preview doesn't select the default feature layer correctly when it's supplied via query parameter
- Bump version to v3.0.1-alpha.0 on main branch

### Checklist


t83714:
tested on the client's dev server

Issue #3556, opened 2024-08-22
robots.txt should be served as content-type `text/plain`


t83714:
### robots.txt should be served as content-type `text/plain`

We currently serve a robots.txt with search engine bots instructions to:
- supply sitemap entry point
- avoid certain pages (e.g. search result page) being indexed

The serving response currently carries `content-type` header `text/html`, which might [see here](https://www.nextthing.org/archives/2007/03/12/robotstxt-adventure) prevent certain search bots from recognising this file.  

We will also add additional instructions to prevent URLs with `q` query parameter from being indexed --- although some existing instructions already cover this by URL path, the additional rule might help to secure this logic further.

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3561

PR #3561, opened 2024-08-31
Issue/3556  robots.txt should be served as content-type text/plain and other related improvements

t83714:
### What this PR does

Fixes:
- #3556
- Added additional rules to prevent any URL with `q` query parameter from being indexed
- Adjust sitemap URL for deployment where default UI is not served at the root path `/`
- Only attempt to popular search input content from `q` query parameters on the `search results` page
- Add robots `<meta>` (`noindex`) to all pages except pages that are allowed to be crawled & indexed by search engine bots (currently `/home` & `/datasets/*`)
- Use `uiBaseUri` for sitemap & crawler viewer URL generation so that the URL will work better when users choose to move the default UI to a none-root path (and use their own UI as landing/home UI)

### Checklist


t83714:
tested in dev site

Issue #3590, opened 2025-03-24
Make indexer working under Async mode


t83714:
### Make indexer working under Async mode

Currently, [the indexer](https://github.com/magda-io/magda/tree/main/magda-indexer) works under sync mode.

i.e. when the registry deliver the HTTP notification to the indexer, it expect the indexer to respond the processing via the same HTTP request.

As we introduce the vector search & embedding API to the processing, the indexer processing takes much longer than it was previously. 

It would be beneficial to update indexer logic to consume registry event notification asynchronously:
- When register a web hook (event listener) with the registry, we should set `async` = true
  - Here is one example minion working under async mode: https://github.com/magda-io/magda-minion-broken-link/blob/ecafafabb14aa2f75c9889bbe02b7d208203c91b/src/index.ts#L57
- Instead of trying to complete the indexing task within the same notification request
- respond deferred response (by setting `deferResponse` to true) see API doc: https://dev.magda.io/api/v0/apidocs/index.html#api-Registry_Webhooks-PostHttpSomeRemoteHostWebhookNotificationRecipientEndpoint
- indexer keep working on the processing
- once it's done, it should make another request to notify registry it's done and it's ready for new notification

t83714:
closed via PR: https://github.com/magda-io/magda/pull/3593

PR #3593, opened 2025-04-07
Issue/3590

JHZhang2736:
Implements the async indexer feature for issue #3590


t83714:
Excellent 🚀  looks good to me 👍 
I guess we probably just need to do a final in cluster test then we are good to merge 🎉 
Let's talk about in cluster test when we catch up today

t83714 (approved):
@JHZhang2736 Well done~ Looks great 🚀 
And seems your local test worked well~ 
image
I think it's ready to merge now 🎉

PR #3597, opened 2025-04-15
Bump OPA to v1.3.0, and use Regal for linting

anderseknert:
### What this PR does

As part of improving our linter [Regal](https://github.com/StyraInc/regal/), I tend to run it against Rego projects I can find in the wild, and this was one I hadn't seen before.

Since the version of OPA used was ancient, my first move was to have that bumped to the latest version, which entails also moving over to more modern syntax, keywords, etc.. where applicapble. With that out of the way, I addressed the few hundred issues reported by Regal. I've obviously made sure all the unit tests are passing, and I don't think there have been any changes in behavior. But if you have a more holistic way of testing that, that would of course be great.

I found at least one issue that Regal didn't correctly detect in this repo, so a small win! Anyway, I figured I'd submit back my fixes.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## v5.1.1
 
+- Improve: Bump OPA version to latest (v1.3.0) and use Regal for linting Rego
 - Fix: SQLConsole: reference error when join JSONL data source
 - Improve: SQLConsole JSON & JSONL data error handling
 
--- a/magda-opa/Dockerfile
+++ b/magda-opa/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:3
-COPY --from=openpolicyagent/opa:0.37.2-static /opa /opa
+COPY --from=openpolicyagent/opa:1.3.0-static /opa /opa
 COPY component/start.sh /bin/start.sh
 COPY component/policies /policies
 
--- a/magda-opa/docker-compose.yml
+++ b/magda-opa/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3"
 services:
   test-opa:
-    image: openpolicyagent/opa:0.37.2-static
+    image: openpolicyagent/opa:1.3.0-static
     ports:
       - 8181:8181
     volumes:
--- a/magda-opa/package.json
+++ b/magda-opa/package.json
@@ -3,12 +3,14 @@
   "version": "5.1.0",
   "description": "MAGDA's Open Policy Agent docker image with built-in policies.",
   "scripts": {
+    "check": "docker run -v $PWD/policies:/policies openpolicyagent/opa:1.3.0-static check --strict ./policies",
     "dev": "docker-compose up",
     "dev-stop": "docker-compose down",
     "docker-build-local": "create-docker-context-for-node-component --build --push --tag auto --local",
     "docker-build-prod": "create-docker-context-for-node-component --build --push --tag auto",
+    "lint": "run -v $PWD/policies:/policies ghcr.io/styrainc/regal:0.32.0 lint /policies",
     "retag-and-push": "retag-and-push",
-    "test": "docker run -v $PWD/policies:/policies openpolicyagent/opa:0.37.2-static test -v ./policies"
+    "test": "docker run -v $PWD/policies:/policies openpolicyagent/opa:1.3.0-static test -v ./policies"
   },
   "author": "",
   "license": "Apache-2.0",
--- a/magda-opa/policies/.regal/config.yaml
+++ b/magda-opa/policies/.regal/config.yaml
@@ -1,0 +1,24 @@
+project:
+  rego-version: 1
+
+rules:
+  idiomatic:
+    no-defined-entrypoint:
+      level: ignore
+  imports:
+    prefer-package-imports:
+      level: ignore
+  style:
+    external-reference:
+      level: ignore
+    file-length:
+      level: ignore
+    line-length:
+      level: ignore
+    prefer-snake-case:
+      level: ignore
+    rule-length:
+      level: ignore
+  testing:
+    test-outside-test-package:
+      level: ignore
--- a/magda-opa/policies/api/allow.rego
+++ b/magda-opa/policies/api/allow.rego
@@ -1,5 +1,7 @@
 package api
+
 import data.common.extractApiOperationUri
+import rego.v1
 
 # allow policy defines rules for generic gateway level enforcement based on HTTP request path & method
 # system admin might choose to use this access model for APIs that doesn't require making access control decision with high level data model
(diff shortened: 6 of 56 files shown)


t83714:
@anderseknert  Great work on the PR!
I created a [new branch](https://github.com/magda-io/magda/tree/test-opa) to run the PR through our test cases in CI.
It seems the newer version of OPA introduces a `--addr` parameter. Without explicitly setting it to `0.0.0.0:8181`, I wasn’t able to access it from the Docker container. I’ve pushed a new commit adding this parameter so we can test the PR further.

However, [the test cases failed with the following error](https://gitlab.com/magda-data/magda/-/jobs/9969136051):
```
Error: Invalid 3 terms rego expression, can't locate operator: internal.member_2 input.object.record.access-control.orgUnitId []
    at RegoExp.toOperatorOperandsArray (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:933:19)
    at RegoExp.evaluate (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:978:47)
    at Function.parseFromData (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:1110:13)
    at <anonymous> (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:380:46)
    at Array.map (<anonymous>)
    at Function.createExpressionsFromRuleBodyData (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:380:21)
    at Function.parseFromData (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:363:35)
    at <anonymous> (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:1648:47)
    at Array.forEach (<anonymous>)
    at <anonymous> (/builds/magda-data/magda/magda-typescript-common/src/OpaCompileResponseParser.ts:1647:23)
```

The error happens around here: https://github.com/magda-io/magda/blob/499a7d0f6787326eac86a3ad3830c9d1722e4777/magda-typescript-common/src/OpaCompileResponseParser.ts#L978

This code is part of our Rego AST parser, which translates the "partial auth decision" into an actual query.

I’m guessing the issue is caused by a syntax change in the newer version of OPA?

Do you have any recommendations on the best way to resolve this issue?

anderseknert:
Thanks @t83714 :) 

> It seems the newer version of OPA introduces a --addr parameter. Without explicitly setting it to 0.0.0.0:8181, I wasn’t able to access it from the Docker container.

That's correct. The default was changed with 1.0 to avoid situations where OPA would accidentally be exposed outside of where users intended. Setting it explicitly like you did for when needed is the way to go 👍 

Oh, I was not aware of that custom parsing! The `internal.member_2` and `internal.member_3` functions are the internal names used by OPA to represent the `in` operator: https://github.com/open-policy-agent/opa/blob/5a8932326fc7f40f123ed358b63d308d12da9805/v1/ast/builtins.go#L356

https://www.openpolicyagent.org/docs/latest/policy-language/#membership-and-iteration-in

That operator has been available for a long time, but before OPA 1.0 would require users to `import rego.v1` in their policies, which with 1.0 is no longer needed. There are a few more keywords that are common in 1.0, but most (like `contains` in partial rule heads) are not reflected in the AST, so should not have an impact here. https://www.openpolicyagent.org/docs/latest/v0-upgrade/#changes-to-rego-in-opa-v10

srenatus:
> This code is part of our Rego AST parser, which translates the "partial auth decision" into an actual query.

Yeah you might have left the fragment of what's translatable by going from some `x := foos[_]` to `some x in foos`. It's a harmless translation in terms of Rego (hence Regal advises it, it's more readable), but it'll leave the `internal.member_2` call in your PE queries.

So, it's a fork in the road: either enhance your PE translation to support the internal.member_2 and internal.member_3 builtins, or skip that regal rule, keep what you had been doing before.

t83714:
@anderseknert  & @srenatus  Thank you for your comments and suggestions - much appreciated.
I think the best option for us would probably be to upgrade our PE translation code to support the new builtins.
Before I dive into that, it would be great if you could shed some light on the following:
- Can I confirm that the only new things in AST between 0.xx & 1.3.x would be the `internal.member_2` and `internal.member_3` and any other new keywords / features introduced since v1 have no impact at AST level?
- a portion of the E AST I captured from test cases looks like the following (full result: [opa-ast.json](https://github.com/user-attachments/files/20160862/opa-ast.json)):  
```json
{
  "index": 5,
  "terms": [
    {
      "type": "ref",
      "value": [
        {
          "type": "var",
          "value": "internal"
        },
        {
          "type": "string",
          "value": "member_2"
        }
      ]
    },
    {
      "type": "ref",
      "value": [
        {
          "type": "var",
          "value": "input"
        },
        {
          "type": "string",
          "value": "object"
        },
        {
          "type": "string",
          "value": "record"
        },
        {
          "type": "string",
          "value": "access-control"
        },
        {
          "type": "string",
          "value": "orgUnitId"
        }
      ]
    },
    {
      "type": "array",
      "value": [
        {
          "type": "string",
          "value": "e87ecaf8-6ec2-4d98-a1b1-a4ef85c3540a"
        },
        {
          "type": "string",
          "value": "88e8d5e7-5eae-46cd-84bd-95936b67913b"
        },
        {
          "type": "string",
          "value": "3b959688-7f82-402e-b20e-b7cd05252780"
        },
        {
          "type": "string",
          "value": "f2ba4567-f4e3-4763-ab4f-1fe4815c2cf1"
        }
      ]
    }
  ]
}
```
> Here, the `internal.member_2` is one `ref` item of the `terms` array.

However, for the following rule:
```rego
deny if {
    some x in input.roles # iteration
    x == "denylisted-role"
}
```
I got the following AST:
```json
{
  "package": {
    "path": [
      {
        "type": "var",
        "value": "data"
      },
      {
        "type": "string",
        "value": "test"
      }
    ]
  },
  "rules": [
    {
      "body": [
        {
          "index": 0,
          "terms": {
            "symbols": [
              {
                "type": "call",
                "value": [
                  {
                    "type": "ref",
                    "value": [
                      {
                        "type": "var",
                        "value": "internal"
                      },
                      {
                        "type": "string",
                        "value": "member_2"
                      }
                    ]
                  },
                  {
                    "type": "var",
                    "value": "x"
                  },
                  {
                    "type": "ref",
                    "value": [
                      {
                        "type": "var",
                        "value": "input"
                      },
                      {
                        "type": "string",
                        "value": "roles"
                      }
                    ]
                  }
                ]
              }
            ]
          }
        },
        {
          "index": 1,
          "terms": [
            {
              "type": "ref",
              "value": [
                {
                  "type": "var",
                  "value": "equal"
                }
              ]
            },
            {
              "type": "var",
              "value": "x"
            },
            {
              "type": "string",
              "value": "denylisted-role"
            }
          ]
        }
      ],
      "head": {
        "name": "deny",
        "value": {
          "type": "boolean",
          "value": true
        },
        "ref": [
          {
            "type": "var",
            "value": "deny"
          }
        ]
      }
    }
  ]
}
```

> Here, the `internal.member_2` is one `call` item of the `symbols` array.

Could you give me some guidance on when `internal.member_2` is compiled as a `call` item (in the `symbols` array) rather than a `ref` item (in the `terms` array)?

If there's any documentation that could help me better understand AST generation in this area, I'd really appreciate it if you could share a link.

Issue #3596, opened 2025-04-13
Set Vector workload mode to "on_disk" by default to reduce costs with disk-based vector search


t83714:
## Description

Since OpenSearch 2.17 (we already on 2.17), the knn vector support `on_disk` vector workload mode in additional to the previous default `in_memory`. The new `on_disk` vector workload mode will [significantly reduce the costs of running vector search engine](https://opensearch.org/blog/Reduce-Cost-with-Disk-based-Vector-Search/) while offer acceptable performance for general use cases.

also see this re: estimate RAM requirement for `in_memory` mode: https://aws.amazon.com/blogs/big-data/choose-the-k-nn-algorithm-for-your-billion-scale-use-case-with-opensearch/ 

The document re: vector workload mode configuration can be found from: https://opensearch.org/docs/latest/field-types/supported-field-types/knn-memory-optimized/

More configuration of KNN vector fields: https://opensearch.org/docs/latest/field-types/supported-field-types/knn-methods-engines/

Therefore, we should:
- set vector workload mode of all our vector fields to `on_disk`
- allow it to be configurable via helm chart (people can stick to `memory` mode for higher performance for their needs).
-  Other parameters we should add & make them configurable via helm charts:
  - `compression_level`: default to 32x if not set

> keep in mind, when `on_disk` mode is set, `encoder` config cannot be set and only float vector types are supported.
> http://localhost:9200/datasets52/_mapping

## Technical Notes

- Currently, the vector fields are configured here: https://github.com/magda-io/magda/blob/c1d93c2f6cc8a84573a5d12842fd668f7f12884e/magda-scala-common/src/main/scala/au/csiro/data61/magda/search/elasticsearch/IndexDefinition.scala#L120
  - Most parameter are hardcoded and not configurable at this moment.
  - We should firstly make the following configurable (we still only support `faiss` engine & `Hnsw` algorithm. So they are not configurable):
    - `dimension`: default to 768
    - `spaceType`: default: l2
    - `efConstruction`: default: 100
    - `efSearch`: default to 100
    - `m`: default to 16
    - when `on_disk` mode: 
      - `encoder` field is not configurable. Should always be not set (default to `flat`). 
    - when `memory` mode: 
      - `encoder`: default to `sq` encoder
        - `type`: default to fp16
        - `clip`: default to false

[...]

PR #3598, opened 2025-04-23
Set Vector workload mode to "on_disk"

JHZhang2736:
### What this PR does

Set KNN  vector workload mod to "on_disk" as the default workload mode. ([issue #3596](https://github.com/magda-io/magda/issues/3596))

### Checklist

-   [ ] Update elastic4s fork to support on_disk vector workload mode and release snapshot version.
-   [ ] Make vector field parameters configurable via common.conf.
-   [ ] Allow it to be configurable via helm chart.


magda-int-test-ts/src/ServiceRunner.ts
```diff
@@ -86,6 +86,8 @@ export default class ServiceRunner {
     // in order to setup the indices in search engine.
     // however, indexer will auto exit if this field is set to false.
     public enableIndexer = false;
+    public hybridSearchVectorWorkloadMode: "on_disk" | "in_memory" =
```
t83714, line 89: Good job on injecting new config to make searchApi support hybrid search 💯 
As the ServiceRunner.ts is a common utility to run test cases I guess we might need to update searchApi config for other tests, is it possible to replace the specific `hybridSearchVectorWorkloadMode` config property here with a more generic config property e.g.:
```typescript
public searchApiConfig: string | null = null
```
Here, we only inject / set `-Dconfig.file=${confFilePath}` when `searchApiConfig` is a non-empty string.
This way, any test cases can pass in a new config content as string to get a customised search API for different purpose.

deploy/helm/internal-charts/indexer/values.yaml
```diff
@@ -45,12 +45,28 @@ autoReIndex:
...
   elasticSearch:
     serverUrl: http://opensearch:9200
     shards: 1
     replicas: 0
+    indices:
+      datasets:
+        hybridSearch:
+          # -- Configuration for hybrid vector search
+          vectorConfig:
+            mode: "on_disk"
```
t83714, line 59: Let's set to on_memory by default for now (till opensearch supports radial search or allow filter low score result easier)
Other default config might need to adjust according to this.
We probably can add instruction (or typical config) for on_disk to helm doc content (by adding comments) see helm docs here: https://github.com/norwoodj/helm-docs

magda-scala-common/src/main/resources/common.conf
```diff
@@ -97,7 +97,33 @@ elasticSearch {
                 # The config will be applied by the following logic:
                 # when `k` is specified, `minScore` & `maxDistance` will be ignored.
                 # Otherwise, when `maxDistance` is specified, `minScore` will be ignored.
+                # when mode = "on_disk", `k` must be specified for Radial search is not supported for indices which have quantization enabled
                 minScore = 0.5
+                vectorConfig {
```
t83714, line 102: not a big deal but any chance we can update `vectorConfig` with `knnVectorFieldConfig` --- it will make the purpose more clear

magda-scala-common/src/main/resources/common.conf
```diff
@@ -97,7 +97,33 @@ elasticSearch {
                 # The config will be applied by the following logic:
                 # when `k` is specified, `minScore` & `maxDistance` will be ignored.
                 # Otherwise, when `maxDistance` is specified, `minScore` will be ignored.
+                # when mode = "on_disk", `k` must be specified for Radial search is not supported for indices which have quantization enabled
```
t83714, line 100: It would be good to add those information (as user instruction) to helm doc (by adding comments in values.yaml)

magda-scala-common/src/main/resources/common.conf
JHZhang2736: Updated as suggested

magda-scala-common/src/main/resources/common.conf
JHZhang2736: Done. Added detailed instructions to helm doc.

t83714:
@JHZhang2736 last week, OpenSearch v3 was released: https://opensearch.org/blog/unveiling-opensearch-3-0/
I guess some of the hybrid search improvement might help our low score documents in result issue.
We probably can have a look to see if it's relevant to this PR

> Improve hybrid search result relevance with a lower bound for min-max normalization
This release addresses a key limitation in hybrid search relevance with the addition of a lower bound for min-max [normalization](https://docs.opensearch.org/docs/latest/search-plugins/search-pipelines/normalization-processor). Traditional min-max normalization can disproportionately elevate documents with extremely low scores when combining neural and keyword search results, leading to suboptimal rankings. A lower bound lets you establish a minimum threshold during normalization, preventing the over-amplification of negligible scores that previously could dominate search results and ensuring that the normalized values remain meaningful and proportionate to their actual relevance. When scores fall below the specified lower bound, they’re adjusted to maintain an appropriate ranking order, which is particularly beneficial for queries where one search method yields minimal relevance signals.

here is release notes: https://github.com/opensearch-project/opensearch-build/blob/main/release-notes/opensearch-release-notes-3.0.0.md

JHZhang2736:
@t83714  read the PR regarding the lower_bounds configuration and found that this feature was introduced to address the challenge of estimating the minimum score during the min-max normalization process:

> “In context of OpenSearch, finding the minimum score is based on assumption that may be not the most effective one. While handling search request, system retrieves limited amount of matching documents from each shard, this limit is defined by the query parameter size. Minimum score will be identified as minimum from all score from all collected documents. In case overall number of matching documents is much higher than number of retrieved documents, then the delta between real and retrieved minimum scores can be significant. This will negatively influence the final normalized score.”
https://github.com/opensearch-project/neural-search/issues/1189

From my understanding, lower_bounds allows users to manually specify minScore during the min-max normalization phase to improve the accuracy of score normalization. 


This could potentially help us improve the balance between different sub-query types (e.g. BM25 vs k-NN) in hybrid search scoring. However, I think lower_bounds only influences the normalized score distribution, but does not affect how many documents are retrieved. So I guess it cannot work as a filter and replace the behavior of min_score in radial search.

PR #3604, opened 2025-06-05
LLM Semantic Indexing Minion SDK

### LINKED ISSUE #3602: LLM Semantic Indexing Minion SDK
## Goal

Develop a reusable SDK to support building Magda minions for LLM-based vector indexing. The SDK should simplify the steps of subscribing to registry, detecting files or registry records to index, generating vector embeddings, and writing results to OpenSearch.

## Functionalities

1. **Auto Create OpenSearch Index**  
   Automatically create the index mapping with a predefined schema(https://github.com/magda-io/magda/issues/3536). 

2. **Built on Top of the Magda Minion Framework**  
   Inherits all core capabilities of a Magda minion:  
   - Subscribe to registry webhook
   - Listens to record changes.
   - Crawling and recrawling

3. **Configurable Listening Target**  
   Users can set `itemType` to choose what to index:

   - **`registryRecord`**: Index metadata fields from registry records  
     User should provide:  
     - `aspects`: which aspects to listen for  
     - `createEmbeddingText`: function to extract text from aspect object  

   - **`storageObject`**: Index content from files
     User should provide:  
     - `fileTypes`: which file types to process  
     - `createEmbeddingText`: function to extract text from file content
     - `autoDownloadFile`: whether to enable autodownloading. If enable, the SDK will automatically download the file content based on the record’s download URL before calling `createEmbeddingText`
    
4. **Configurable Chunking**  
   Users can specify `chunkSize` and `overlap`.
   Users can also provide self-designed chunking strategy. 

5. **Automatic Chunking, Embedding, and Writing to OpenSearch**  
   - This SDK will split the text returned from `extractText()` into chunks  
   - Each chunk is embedded and written to OpenSearch

  
### Indexing Metadata
```Typescript
semanticIndexer({
    itemType: "registryRecord",
    aspects: ["dcat-dataset-strings"],
    optionalAspects: [],
    chunkSizeLimit: 512,         // chunk size in tokens
    overlap: 64,                 // overlap in tokens
    createEmbeddingText: async (record: RegistryRecord, format: string | null, filePath: string | null, url: string | null) => {
        // Return a plain string to be chunked and embedded
    },
    chunkStrategy: async (text: string) => {
        // Optional: Provide custom chunking logic
    }
});
```
### Indexing File Content
```Typescript
semanticIndexer({
    itemType: "storageObject",
    fileTypes: ["pdf", "docx", "txt"],
    autoDownloadFile: true,
    chunkSizeLimit: 512,
    overlap: 64,
    createEmbeddingText: async (record: RegistryRecord, format: string | null, filePath: string | null, url: string | null) => {
        // Extract and return plain text from the file
    },
    chunkStrategy: async (text: string) => {
        // Optional: Provide custom chunking logic
    }
});
```

JHZhang2736:
## Summary

This PR implements the reusable SDK to support building Magda minions for LLM-based vector indexing. 

## Features Implemented

- [x] Auto Create OpenSearch Index with a predefined schema.
- [x] Built on top of Magda Minion Framework:
- [x] Configurable indexing target
- [x] Configurable Chunking
- [x] Automatic chunking, embedding, and indexing into OpenSearch.

Details are described in issue #3602.

This PR closes #3602.


magda-typescript-common/src/semantic-indexer/commonSemanticIndexerYargs.ts
```diff
@@ -0,0 +1,41 @@
+import yargs from "yargs";
+import { MinionArguments, commonYargs } from "@magda/minion-sdk";
```
t83714, line 2: @magda/minion-sdk might not be built yet at this stage as it depends on minion-framework and minion-framework is built on typescript-common.
It might work on your local as you probably already build @magda/minion-sdk previously (i.e. built result is already in @magda/minion-sdk/dist/* )

magda-typescript-common/src/BaseApiClient.ts
```diff
@@ -29,11 +29,9 @@ export default abstract class BaseApiClient {
     public readonly userId: string;
 
     constructor(options: BaseApiClientConfig) {
-        this.baseApiUrl = options.baseApiUrl
-            ? options.baseApiUrl
-            : "http://localhost:6103/v0";
+        this.baseApiUrl = options.baseApiUrl;
```
t83714, line 32: can we double check if any code rely on default value http://localhost:6103/v0"?

JHZhang2736: Only IndexerApiClient directly references BaseApiClient, I’ve set a default baseApiUrl in that class: [magda-typescript-common/src/IndexerApiClient.ts](https://github.com/magda-io/magda/pull/3604/files#diff-9adeb479ba85ff7e2b2cd7c87d2cab19c58cb63b69266976158ace892fe207da)
``` typescript
export default class IndexerApiClient extends BaseApiClient {
    constructor(options: BaseApiClientConfig) {
        options.baseApiUrl = options.baseApiUrl || "http://localhost:6103/v0";
        super(options);
    }
```

magda-semantic-indexer-framework/src/config.json
```diff
@@ -0,0 +1,30 @@
+{
```
t83714, line 1: I guess it's a default config file --- if so, could we rename it to default-config.json to make it more obvious?

magda-semantic-indexer-framework/src/indexSchema.ts
```diff
@@ -0,0 +1,113 @@
...
+export function createSemanticIndexerMapping() {
+    const knnVectorFieldConfig =
+        config.elasticSearch.indices.semanticIndex.knnVectorFieldConfig;
+
+    if (knnVectorFieldConfig.compressionLevel && knnVectorFieldConfig.encoder) {
+        throw new Error("compressionLevel and encoder cannot be used together");
+    }
+
+    return {
+        indexName: config.elasticSearch.indices.semanticIndex.indexName,
```
t83714, line 14: I guess the index mapping / schema will evolve down the line.
how about we use the indexName for purpose of index versioning (unless we have a better option)?
e.g. we assume the current version is 1.
so the generate the indexName should be `${config.elasticSearch.indices.semanticIndex.indexName-1}`
when release a new version that requires mapping changes, we can increase the version number in config so a new index with new mapping version will be created.

magda-semantic-indexer-framework/src/indexSchema.ts
```diff
@@ -0,0 +1,113 @@
...
+        settings: {
+            index: {
+                number_of_shards: 1,
+                number_of_replicas: 1
```
t83714, line 18: this number parameter likely need to be configurable as they need be set based on index size & availability needs.
shards = 1 & replicas = 0 could be a good default for a testing cluster

magda-semantic-indexer-framework/src/config.json
```diff
@@ -0,0 +1,30 @@
...
+    "default": {
+        "chunkSizeLimit": 100,
+        "overlap": 10
+    },
+    "elasticSearch": {
+        "serverUrl": "http://localhost:9200",
+        "indices": {
+            "semanticIndex": {
+                "indexName": "semantic-index",
+                "knnVectorFieldConfig": {
```
t83714, line 11: we should make fields in `knnVectorFieldConfig` configurable via either command lines, env variable or json config file interface as well (I guess, json config file might works better).
in future, those values will be passed through `global.searchEngine.semanticIndexer.xxx` either via command lines, env variable or json config file (mapping from k8s config map)

magda-semantic-indexer-framework/src/semanticIndexerOptions.ts
```diff
@@ -0,0 +1,51 @@
...
+    argv: SemanticIndexerArguments;
+    id: string;
+    itemType: ItemType;
+    aspects?: string[];
+    optionalAspects?: string[];
+    formatTypes?: string[];
+    createEmbeddingText: CreateEmbeddingText;
+    chunkSizeLimit?: number;
+    overlap?: number;
+    autoDownloadFile?: boolean;
```
t83714, line 15: since we already defined chunkStrategy & chunker, we might want to chunker to be configurable? i.e. by default, using RecursiveChunkStrategy chunker. But if the user supplies a new one, we just use the user supplied one?

t83714 (commented):
well done & good job 👍  a few notes for you to have a look

magda-minion-framework/tsconfig.json
```diff
@@ -4,7 +4,10 @@
         "outDir": "./dist",
         "strictNullChecks": false,
         "rootDir": "./src",
-        "tsBuildInfoFile": "./tsconfig.tsbuildinfo"
+        "tsBuildInfoFile": "./tsconfig.tsbuildinfo",
+        "composite": true,
+        "declaration": true,
```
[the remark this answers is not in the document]
JHZhang2736: Can we add `"declarationMap": true` to `../tsconfig-global.json`?  
It might improve development experience by mapping `.d.ts` files back to the original `.ts` source when debugging type definitions.

> "Generates a source map for `.d.ts` files which map back to the original `.ts` source file. This will allow editors such as VS Code to go to the original `.ts` file when using features like Go to Definition.  
> You should strongly consider turning this on if you’re using project references."  
>https://www.typescriptlang.org/tsconfig/#declarationMap:~:text=Generates%20a%20source,using%20project%20references.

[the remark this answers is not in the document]
t83714: At this moment, in our monorepo, we often import the source directly. 
e.g.:
```
import EmbeddingApiClient from "magda-typescript-common/src/EmbeddingApiClient.js";
import OpensearchApiClient from "magda-typescript-common/src/OpensearchApiClient.js";
```
> .js extension just a typescript solution in ESM imports to ensure compatibility with the runtime
and there is no .d.ts files are involved in this process.
You can try the "Go to Definition" now --- it will take you to the source without setting declarationMap.
declarationMap will be useful when we import like `import {EmbeddingApiClient} from "magda-typescript-common"`
This way it resolves to the actual built version and .d.ts files.
However, that option won't work now - before we make that approach work, we need to:
- tidy up the structure of magda-typescript-common and make sure all modules are only exported via designed public interface e.g. via index.ts or configured exports in packages
- package.json is properly set up

magda-semantic-indexer-framework/src/indexSchema.ts
JHZhang2736: Added configuration for shards and replicas.

magda-semantic-indexer-framework/src/config.json
JHZhang2736: This is now configurable via the json config file in the latest commit. let me know if I didn’t do it the right way.

magda-semantic-indexer-framework/src/semanticIndexerOptions.ts
JHZhang2736: Added support for user-defined chunking strategies.

magda-semantic-indexer-framework/src/onRecordFoundStorageObject.ts
```diff
@@ -0,0 +1,164 @@
...
+    const response = await fetch(url);
+    if (!response.ok) {
+        throw new SkipError(`${response.status} ${response.statusText}`);
+    }
+
+    const tempDir = tmpdir();
+    const tempFileName = `${uuidv4()}`;
+    const tempFilePath = join(tempDir, tempFileName);
+
+    const fileStream = await fs.open(tempFilePath, "w");
```
t83714, line 151: performance wise, use fs.createWriteStream to create a writable stream and pipe in the response.body (readable stream) would be better.
e.g. something like: response.body.pipe(fs.createWriteStream(filepath))

magda-semantic-indexer-framework/src/skipError.ts
```diff
@@ -0,0 +1 @@
+export class SkipError extends Error {}
```
t83714, line 1: Probably want to match the file name case with the class name i.e. the first letter of the file name should be upper case as well

t83714 (commented):
Good job~  added a few notes for you to have a look~

magda-semantic-indexer-framework/src/indexEmbeddingText.ts
```diff
@@ -0,0 +1,126 @@
...
+        );
+    }
+    if (EmbeddingText.subObjects) {
+        for (const sub of EmbeddingText.subObjects) {
+            await processSingleText(
+                options,
+                sub.text,
+                {
+                    ...meta,
+                    subObjectId: sub.subObjectId || uuidv4().slice(0, 8),
```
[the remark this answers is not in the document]
JHZhang2736: Thanks for the suggestion. I guess uniqueness is more important than saving space here. Will use uuidv4() without slicing.

JHZhang2736:
@t83714 I’ve updated the code for all the points above. let me know if I missed anything~

magda-semantic-indexer-framework/src/indexEmbeddingText.ts
```diff
@@ -0,0 +1,126 @@
...
+    opensearchApiClient: OpensearchApiClient,
+    bulkSize: number
+) {
+    const chunks = await chunker.chunk(text);
+    if (!chunks || chunks.length === 0) {
+        throw new SkipError("No chunks generated from text.");
+    }
+    let embeddings;
+    try {
+        embeddings = await embeddingApiClient.get(chunks.map((c) => c.text));
```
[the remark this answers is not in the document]
t83714: I think those two should be each individual indexer helm chart config rather than global settings as likely users might want to tune the config differently for different indexer. e.g. they might have a setup they want one more important indexer take more resources and work faster than others.
If we don't mind the troubles, we can do:
- make the config configurable at both global and at individual indexer helm charts
- unless values are set at indexer helm chart level, global values will be used.
This way can probably cover both cases (ability to set different config for different indexer helm charts and ability to set same config globally without having to repeat the same values for each indexer helm charts)
Those logic likely will be handled as part of helm charts. The SDK probably only need to make sure there is an interface that allows the two values to be configured.

t83714 (commented):
I guess you probably want to try released SDK by creating a "real" indexer - I created a repo here: https://github.com/magda-io/magda-pdf-semantic-indexer
so you can save your code and setup CI. If you plan to create a different type of index (rather than PDF), please let me know and I can create a different repo.

magda-semantic-indexer-framework/src/onRecordFoundStorageObject.ts
```diff
@@ -0,0 +1,241 @@
...
+                )
+                .on("finish", resolve);
+        });
+        return tempFilePath;
+    } catch (err) {
+        await deleteTempFile(tempFilePath);
+        throw err;
+    } finally {
+        if (writeStream) {
+            writeStream.end();
```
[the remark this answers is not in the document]
JHZhang2736: Thanks for the suggestion~ I’ve looked through some details of pipeline, and I agree it’s a better approach here. I’ve updated the code accordingly.

t83714:
superseded by PR: https://github.com/magda-io/magda/pull/3618

PR #3646, opened 2026-04-19
Issue/3608 alt 2

### LINKED ISSUE #3608: Attributes Sync minion for semantic indexer index
## Attributes Sync minion for semantic indexer index

The semantic indexer index mapping defined in ticket:

- #3536

chensuihui:
### What this PR does

Fixes #3608

This PR introduces an alternative authorization-aware semantic search strategy based on **access-scope caching**, implemented as a **conditional optimization path** (not a replacement for the existing two-phase approach).

#### Summary of changes

- Added a new Registry API endpoint: `GET /v0/records/accessibleIds`
  - Computes all record IDs accessible under the current `(tenant, authDecision)` context
  - Stores the result in Redis
  - Returns a Redis cache key instead of raw IDs
- Introduced Redis-backed cache keying/versioning model:
  - Data key: `records:{tenantId}:{authDecisionHash}:{eventId}`
  - Index key: `records:idx:{tenantId}:{authDecisionHash}`
- Added read-path optimization behavior:
  - Reuse cached key when present
  - Build and store snapshot on cache miss
- Added/updated tests for `accessibleIds` behavior and related auth flow integration.

### Why this approach

The existing two-phase query model remains correct, but can be inefficient for users with narrow access scopes (frequent fallback / repeated filtering).

This PR shifts authorization handling from post-retrieval filtering to pre-query scoping **when beneficial**, by reusing cached accessible ID sets for repeated authorization contexts.

### Design choices and trade-offs

- **Kept existing two-phase strategy as baseline**
  - This PR does not replace current logic globally.
  - The cache-based path is intended as a targeted optimization.
- **Versioned cache with `eventId`**
  - Ensures cache snapshots can be invalidated/refreshed as data changes.
- **Full refresh first (no incremental diff in this PR)**
  - Chosen for correctness and implementation simplicity.
  - Incremental `streamEventsSince` updates are intentionally deferred due to complexity around auth-membership changes.

### Operational notes

- Effectiveness is best for small/medium access scopes and repeated queries under stable auth contexts.
- For large accessible ID sets, performance can degrade (Redis size + OpenSearch `terms` filter cost), so this is designed to coexist with fallback behavior.

### Checklist


deploy/helm/internal-charts/registry-api/values.yaml
```diff
@@ -1,10 +1,10 @@
 global: {}
-image: 
+image:
```
t83714, line 2: Not a big issue - just a note: would always be good to minimise unnecessary changes to the files to reduce review workload 😄 
If we do want to adjust the format / whitespace largely, could use a seperate PR.

deploy/helm/internal-charts/registry-redis/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: A Helm chart for deploying Redis as a shared cache for registry-api and semantic-search-api
+type: application
+version: 5.6.0
+kubeVersion: ">= 1.21.0"
+annotations:
+  magdaModuleType: "core"
+dependencies:
+  - name: redis
+    version: "17.11.3"
+    repository: "https://charts.bitnami.com/bitnami"
```
t83714: bitnami has moved to OCI registry for helm charts since 2024 - see: https://github.com/bitnami/charts/issues/30110

deploy/helm/internal-charts/magda-redis/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v2
+name: registry-redis
+description: A Helm chart for deploying Redis as a shared cache for registry-api and semantic-search-api
+type: application
+version: 5.6.0
+kubeVersion: ">= 1.21.0"
+annotations:
+  magdaModuleType: "core"
+dependencies:
```
t83714, line 9: since we adds the external dependency, we should include a Chart.lock file for "registry-redis" chart

deploy/helm/internal-charts/registry-redis/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v2
+name: registry-redis
```
t83714: registry-redis name is a little too registry-specific if it is explicitly shared by semantic search

deploy/helm/internal-charts/registry-redis/templates/service.yaml
```diff
@@ -0,0 +1,21 @@
+# This service is used to make Redis accessible via a consistent DNS name
+# when deployed as part of the Magda cluster
+apiVersion: v1
```
t83714: why we need this manual service?
Bitnami Redis already creates services with chart-specific names/selectors. 
Better to use Bitnami's own service directly - if you want to control the name of the Bitnami's own service - I guess the  `fullnameOverride` config or similar should do.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsServiceRO.scala
```diff
@@ -1137,11 +1147,231 @@ class RecordsServiceRO(
...
+                      redisClient.set(dataKey, ids.toJson.compactPrint)
+
+                      // Index key stores all historical data keys for the same tenant+auth hash.
+                      // Future incremental flow can locate the newest cached key, call
+                      // streamEventsSince(latestEventId), patch only changed record IDs,
+                      // and write a new records:{tenant}:{authHash}:{newEventId} snapshot.
+                      redisClient.sAdd(indexKey, Seq(dataKey))
+
+                      dataKey
+                  }
```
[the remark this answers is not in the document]
t83714: @chensuihui yeah this comment is valid - we need to treat JWT token as secret and not to use raw JWT token value as key.
It seems the key (that includes the JWT token) could also be printed in logs in SemanticSearchService.ts

magda-semantic-search-api/src/index.ts
```diff
@@ -82,9 +105,51 @@ const embeddingApiClient = await retry(
...
+    5,
+    5,
+    (e, left) =>
+        console.error(
+            `Search API connection failed, remaining retries: ${left}, error:`,
+            e.message
+        )
+);
+
+const redisClient = new RedisClient({
```
t83714, line 140: it seems we don't use "REDIS_PASSWORD" at all when create client?

magda-typescript-common/src/registry/RegistryClient.ts
```diff
@@ -287,4 +287,63 @@ export default class RegistryClient {
...
+                console.log(
+                    formatServiceError(
+                        "Failed to filter records by access.",
+                        e,
+                        retriesLeft
+                    )
+                )
+        )
+            .then((result) => ({ records: result.body }))
+            .catch(toServerError("filterRecordsByAccess"));
```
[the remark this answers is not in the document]
t83714: @chensuihui  this is valid

deploy/helm/internal-charts/magda-redis/values.yaml
```diff
@@ -0,0 +1,20 @@
+# Redis Bitnami chart configuration
+redis:
+  auth:
+    # -- Enable Redis authentication
+    enabled: false
```
t83714, line 11: why we need "REDIS_PASSWORD" if we set `enabled: false` here?

t83714 (commented):
I guess benchmark & experiment design (against baseline) is more important for this PR than making it production-grade. Once the functionality looks good, you probably can add some scripts to help facilitate your benchmarking and experiments.

By the way, besides my comments, many of the copilot auto-review comments are worth looking into as well. Better have a look at all copilot auto-review comments closely.

magda-typescript-common/src/registry/RegistryClient.ts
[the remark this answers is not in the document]
chensuihui: image
I think this copilot comment is outdated. I have fixed this problem via [Merge branch 'issue/3608-alt-1' into issue/3608-alt-2](commit: e403c30505d5ff7760894b4b4b0fe8e569e9389f).

magda-semantic-search-api/src/api/createApiRouter.ts
```diff
@@ -53,16 +53,46 @@ export function createRoutes(
...
+            const rawParams = ((req.method === "GET"
                 ? req.query
                 : req.body) as unknown) as SearchParams;
+            const jwt = req.header("X-Magda-Session");
+            const tenantId = req.header("X-Magda-Tenant-Id");
+            const params: SearchParams = {
+                ...(rawParams as SearchParams),
+                jwt: jwt || undefined,
+                tenantId: tenantId === undefined ? 0 : Number(tenantId)
+            };
```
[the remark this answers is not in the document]
chensuihui: image
fixed. commit: e403c30505d5ff7760894b4b4b0fe8e569e9389f

magda-semantic-search-api/src/service/SemanticSearchService.ts
```diff
@@ -64,19 +203,54 @@ export class SemanticSearchService {
...
+        if (!Array.isArray(registryResponse)) {
+            throw new Error(
+                "Invalid filterByAccess response: expected string[]"
+            );
+        }
+        return new Set(
+            registryResponse.filter(
+                (id): id is string => typeof id === "string" && id.length > 0
+            )
+        );
```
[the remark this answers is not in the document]
chensuihui: I agree that transient registry failures should be handled explicitly, but I do not think converting a non-array response into an empty allow-list is the right behaviour here.

The access-filter endpoint uses an empty array to represent “no accessible records”. A non-array response is not a valid authorisation result; it indicates either an upstream registry failure or a broken response contract. Treating that as deny-all would hide infrastructure errors and make debugging difficult, because users would receive an apparently valid empty result instead of an error.

I will keep the validation here, but I can improve the error handling around the registry client call so registry failures are reported more explicitly rather than being mixed with normal authorisation denial.

magda-semantic-search-api/src/service/SemanticSearchService.ts
```diff
@@ -44,18 +51,150 @@ export class SemanticSearchService {
...
+        const dataKey = dataKeyResponse;
+
+        // Step 2: Fetch record ids from Redis using the data key
+        let accessibleRecordIds: string[] = [];
+        try {
+            const cachedJson = await this.redisClient.get(dataKey);
+            if (cachedJson) {
+                accessibleRecordIds = JSON.parse(cachedJson);
+            }
+        } catch (e) {
```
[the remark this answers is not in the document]
chensuihui: I agree that the Redis value should not be trusted as a valid string[] just because JSON.parse succeeds.

I have updated the code to add runtime validation for the cached value before using it in the OpenSearch terms filter. The parsed Redis value is now checked to ensure it is an array, and the array contents are filtered so that only non-empty strings are passed through. I also de-duplicate the IDs before calling executeVectorSearch.

If the cached JSON is malformed or the parsed value has an unexpected shape, searchAlt now falls back to the existing search path instead of building an invalid OpenSearch query.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsServiceRO.scala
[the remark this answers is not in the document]
chensuihui: Thanks for catching this. I previously used X-Magda-Session incorrectly as part of the Redis key, but that was not intended.

I have updated the cache key generation so it no longer incorporates the raw session/JWT value. The Redis key is now derived only from stable, non-sensitive context: tenant id, auth decision hash, and latest event id.

X-Magda-Session is still documented because it is required by the existing withAuthDecision flow to derive the current user's auth decision, but I have updated the API documentation to clarify that it is not used as part of the Redis key.

magda-registry-api/src/main/scala/au/csiro/data61/magda/registry/RecordsServiceRO.scala
```diff
@@ -1137,11 +1147,231 @@ class RecordsServiceRO(
...
+                    case None =>
+                      val ids = collectAccessibleRecordIds(tenantId, authDecision)
+                      redisClient.set(dataKey, ids.toJson.compactPrint)
+
+                      // Index key stores all historical data keys for the same tenant+auth hash.
+                      // Future incremental flow can locate the newest cached key, call
+                      // streamEventsSince(latestEventId), patch only changed record IDs,
+                      // and write a new records:{tenant}:{authHash}:{newEventId} snapshot.
+                      redisClient.sAdd(indexKey, Seq(dataKey))
+
```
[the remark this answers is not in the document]
chensuihui: I agree that these cache keys should not live indefinitely.

I have updated the implementation to use a configurable TTL for the accessible ID cache. The snapshot key is now written with setEx, and the index set is also assigned an expiry so historical snapshot references do not accumulate indefinitely in Redis. The TTL is configurable rather than hard-coded, so it can be tuned based on deployment and evaluation needs.

deploy/helm/internal-charts/registry-redis/templates/service.yaml
chensuihui: removed

deploy/helm/internal-charts/registry-redis/Chart.yaml
chensuihui: image
magda-redis as new name

magda-semantic-search-api/src/index.ts
chensuihui: remove REDIS_PASSWORD

deploy/helm/internal-charts/magda-redis/values.yaml
chensuihui: remove REDIS_PASSWORD

deploy/helm/internal-charts/magda-redis/Chart.yaml
chensuihui: added

PR #3710, opened 2026-07-14
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `magda-gateway/src/createHttpsRedirectionMiddleware.ts:L13`

The `createHttpsRedirectionMiddleware` function constructs the redirection URL using the `Host` header directly from the incoming request (`req.get('host')`). An attacker can supply a malicious `Host` header (e.g., `Host: evil.com`) to redirect the victim to an attacker-controlled domain, leading to phishing or token leakage via open redirect.

## Solution

Avoid trusting the `Host` header for constructing absolute redirect URLs. Use a server-side configured and trusted domain name, or validate the `Host` header against a strict allowlist before using it.

## Changes

- `magda-gateway/src/createHttpsRedirectionMiddleware.ts` (modified)
- `ci-scripts/npm-publish-release.js` (modified)

### DIFF
--- a/ci-scripts/npm-publish-release.js
+++ b/ci-scripts/npm-publish-release.js
@@ -6,7 +6,7 @@
 //
 // Treats "version already published" as a non-fatal outcome, since release
 // pipelines may be re-run against a version that was already published.
-const { execSync } = require("child_process");
+const { execFileSync } = require("child_process");
 const path = require("path");
 
 const pkg = require(path.join(process.cwd(), "package.json"));
@@ -14,7 +14,7 @@ const match = pkg.version.match(/-([a-zA-Z]+)/);
 const tag = match ? match[1] : "latest";
 
 try {
-    execSync(`npm publish --tag ${tag}`, {
+    execFileSync("npm", ["publish", "--tag", tag], {
         stdio: ["ignore", "inherit", "pipe"]
     });
 } catch (e) {
--- a/magda-gateway/src/createHttpsRedirectionMiddleware.ts
+++ b/magda-gateway/src/createHttpsRedirectionMiddleware.ts
@@ -1,7 +1,8 @@
 import { NextFunction, Response, Request, RequestHandler } from "express";
 
 export default function createHttpsRedirectionMiddleware(
-    enableHttpsRedirection: boolean
+    enableHttpsRedirection: boolean,
+    trustedDomain?: string
 ): RequestHandler {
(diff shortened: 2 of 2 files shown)


t83714 (changes requested):
Thanks again for the security-focused PR - appreciate you flagging these. Before we go further I wanted to share an assessment of the actual severity of each item, since I think both are rated higher in the description than they hold up to on closer inspection.

1. Host-header open redirect (createHttpsRedirectionMiddleware.ts) - realistically Low, not High

This is a genuine weakness class (trusting the Host header, CWE-601/CWE-644), but it doesn't work as a practical open redirect. The middleware only redirects the same request back to whoever sent it, using the Host that requester supplied. To weaponise an open redirect you need to hand a victim a link on the trusted origin that bounces them elsewhere - but a browser derives the Host header from the URL it's connecting to, so a remote attacker can't make a victim's browser reach the Magda host while sending Host: evil.com. Running curl -H "Host: evil.com" only redirects yourself. The one plausible vector is 301 cache-poisoning behind a cache that keys on path-without-Host, which is narrow.

On top of that, enableHttpsRedirection defaults to false, so most deployments never hit this path, and the gateway normally sits behind an ingress that handles HTTPS redirection anyway.

Net: I'd score this Low / defense-in-depth hardening rather than High. Still worth doing since it's cheap, but framed as hardening.

2. "Command injection" (ci-scripts/npm-publish-release.js) - not a real vulnerability

tag is derived via pkg.version.match(/-([a-zA-Z]+)/), so it can only ever contain ASCII letters - no shell metacharacters can survive that regex - and the input is our own repo's package.json, not attacker-controlled. There's no injection path here. The switch to execFileSync is fine as a defensive-style tidy-up, but it isn't fixing an exploitable issue, so I'd treat this as Informational and drop the "command injection" framing.

## Where that leaves us

Given the above, item 1 is a nice-to-have hardening and item 2 is essentially cosmetic - so no pressure either way. Are you still interested in taking these forward? If so, that's very welcome - just keep in mind the issues that are already in the current diff so we get them right this time: notably item 1 as written is a no-op (the new parameter is never passed at the call site (buildApp.ts), so behavior is unchanged), and it'll need to source a trusted host from existing config (i.e. config.externalUrl) plus a test to prove the spoofed-Host case is actually handled. Happy to point you in the right direction on those once you've decided.

Issue #3714, opened 2026-07-14
Investigate: full record-scoped version maintenance in the web client (event-gated reconcile)


t83714:
## Question

Should the web client move to the mgd CLI's fully **record-scoped** version model?

- Distribution record changes (record fields, aspects, file changes) → bump **only** the distribution's version.
- Dataset record changes (record fields, aspects — including `dataset-distributions`, so adding/removing a distribution) → bump **only** the dataset's version.
- A submission that only touched distributions would no longer bump the dataset version (today, every submission bumps the dataset once — the session model's semantic).

This was assessed during the design for #3713 and deliberately deferred: it is feasible, but pulls in product decisions and a rework of an untested, business-critical flow.

## What the assessment established (verified against the code)

**Feasible — the two hard prerequisites already exist:**

1. **Per-record change detection is free.** The registry only mints events for real diffs (`magda-registry-api RecordPersistence.scala`: `patchRecordById` creates a record event only when `name` changed; `putRecordAspectById` only when the aspect patch is non-empty; a no-op PUT returns event id `0` — "update record doesn't always generate events"). So "did this record actually change?" is answered by the `x-magda-event-id` the web client already reads.
2. **The reconcile algorithm is proven.** The CLI's pure `reconcileVersion` (eventId-gated tag-or-bump, `packages/mgd/src/versionAspect.ts`, PR #3709) ports directly; it was verified live against a minikube deployment.

**Core rework sketch (~2–3 days, mostly mechanical):**

- Strip the `version` aspect out of the dataset/distribution record builds (it currently rides *inside* the record PUTs — under event-gating it must move out, or version writes would self-trigger).
- After each record PUT: if its event id > 0 → reconcile that record's version (tag-or-bump) in a follow-up aspect PUT. Distribution descriptions from session state (file replacement → `Replaced superseded by a new distribution` + `internalDataFileUrl` + forceBump; otherwise `Distribution metadata updated`); dataset → `Version created on update submission`.
- Delete `createVersionForDatasetSubmission` and the `ConfirmMetadataModal` in-state bump; `dataset-draft` / `ckan-export` writes already happen outside the record PUT so they're naturally excluded from the bump signal.
- `dataset-distributions` rides inside the dataset record PUT, so add/remove-distribution → dataset bump falls out for free.

**The actual blockers are product decisions, not code:**

[...]

Issue #3687, opened 2026-07-10
mgd CLI: maintain the `version` aspect on high-level dataset/distribution/file commands


t83714:
## Context

Found during coding-agent testing of the `mgd` CLI (epic #3648, PR #3683). An agent observed that a dataset's `version.currentVersionNumber` stayed at `0` after edits and assumed versioning was "handled externally". Investigation shows a real gap: **MAGDA has two different notions of "version", and `mgd` only partially maintains the one it is responsible for.**

## Background — two version concepts

1. **Registry event history (system-managed).** Every create/modify/delete on a record or aspect emits an event carrying a JSON patch; this event stream is MAGDA's audit/event-sourcing history and lets you reconstruct any past state. It is fully automatic — no client action needed. (See the [Guide to Magda Internals](https://github.com/magda-io/magda/blob/main/docs/docs/architecture/Guide%20to%20Magda%20Internals.md).)
2. **The `version` aspect (application-managed).** A semantic version record on datasets/distributions: `{ currentVersionNumber, versions: [{ versionNumber, createTime, creatorId, description, title, internalDataFileUrl?, eventId? }] }`. The registry does **not** touch this — the **client** maintains it. The web client does so (`magda-web-client/src/api-clients/RegistryApis.ts` `getInitialVersionAspectData` / `tagRecordVersionEventId`, and `magda-web-client/src/Components/Dataset/Add/DatasetAddCommon.ts` `createVersionForDatasetSubmission`).

## Current `mgd` behaviour (the gap that was fixed)

- `dataset create` / `add-file` initialised the `version` aspect at `currentVersionNumber: 0` (correct — matches the web client's initial value) but left it **untagged**.
- `dist replace-file` appended a new version entry (correct).
- **But** `dataset update`, `dist update`, and removing a distribution did **not** touch the `version` aspect, and there was **no** `dist remove` command at all. Verified live against dev.magda.io: create → `dataset update` left `currentVersionNumber` at `0`.

---

### How the web client actually behaves (mapped)

- **Dataset version** is bumped **once per edit-session submission** (not per field change) and is **eventId-gated**. Each version item may carry an `eventId` — the registry's `x-magda-event-id` response header from the mutation it corresponds to. It works like a git tag pointing at a commit and enables the registry time-travel API.
  - On submission: if the current version item is **untagged**, it is merely tagged (no bump); if it is **already tagged with a different event id**, a new version (`max(versionNumber) + 1`) is appended and becomes current.
[...]

PR #3721, opened 2026-07-17
feat(web-client): bump distribution version aspect on metadata edits + harden version handling

### LINKED ISSUE #3713: Web client: align distribution version behaviour with mgd CLI & harden version-aspect handling
## Context

PR #3709 (issue #3687) made the mgd CLI maintain the application-managed `version` aspect with eventId-gated, record-scoped bumps. Re-verifying the web client against the divergences recorded there found one was **recorded incorrectly**, one should be resolved by changing the web client (this issue), and one is an intended model difference:

| # | Recorded divergence | Verified status |
| --- | --- | --- |
| D1 | CLI bumps the distribution version on a metadata edit; the web client bumps nothing | Real — **resolved by this issue**: the web client should bump too. |
| D2 | CLI `replace-file` bumps only the distribution; a web-client session that replaces a file also bumps the dataset | Real, but inherent to the web client's one-bump-per-submission session model. **Keep**; documented as intended. |
| D3 | "The web client leaves distribution versions untagged" | **Wrong.** `submitDatasetFromState` passes `tagVersion=true` → `updateDataset(…, tagDistributionVersion=true)` tags every distribution's current untagged version with its PUT/POST event id (`RegistryApis.ts` ~788). Both clients tag both record types. Docs corrected. |

### Latent robustness issues (now that two clients co-write the aspect)

1. `tagRecordVersionEventId` (`RegistryApis.ts:885`) looks up the current version by **array position** (`versions[currentVersionNumber]`) instead of `versions.find(v => v.versionNumber === currentVersionNumber)`. Correct on dense histories; on gappy ones (which the CLI's `max+1` recovery path can produce) it silently no-ops or reads the wrong item.
2. `Edit/Pages/EditFiles/ConfirmMetadataModal.tsx` (~line 207) bumps with `1 + currentVersionNumber` instead of `max(versionNumber) + 1` — can collide or regress on gappy histories. (The dataset-side `createVersionForDatasetSubmission` already does both correctly.)

### D1: bump the distribution version on metadata edits (at submission)

Compose with existing machinery — the web client already seeds dist `v0`, PUTs every distribution at submission, and tags the current-if-untagged version with the PUT's event id. Only the bump is missing:

1. **Dirty flag** — the two `editDistribution(distId)(updater)` implementations (`Add/Pages/AddFiles/index.tsx:59`, `Edit/Pages/EditFiles/index.tsx:61`) set a transient `metadataEditedDuringSession: true` on the edited Distribution state object (`DatasetLinkItem` receives `editDistribution` as a prop, so link/API distributions are covered).
2. **Strip before write** — `convertStateToDistributionRecords` strips the flag from the `dcat-distribution-strings` build, exactly like the existing `aspect.version = undefined` pattern.
3. **Bump at submission** — if a distribution is flagged **and** has an existing version aspect (pre-existing dist, not a new upload) **and** its current version is already tagged with an `eventId`, append a new untagged version: `versionNumber = max+1`, `description: "Distribution metadata updated"` (identical to the CLI), `creatorId = editingUserId`. The tagged-current guard kills two birds:
   - **No double bump**: a dist just bumped by a file replacement in the same session has an untagged current version → the metadata bump is subsumed.
   - **Graceful upgrade**: a legacy dist with an untagged current version skips the bump — it gets tagged by this submission and the *next* edit bumps (the CLI's eventId-gated upgrade path).
4. **Tagging rides the existing path** — the submission PUT + `tagRecordVersionEventId` tag the new version with the real event id. No new API calls.
5. **Flag lifecycle** — cleared after successful submission (no double-bump on resubmit); persists through draft saves inside `dataset-draft` JSON so resumed sessions still know.

New distributions keep today's behaviour: initial `v0`, tagged at submission.

### Hardening

Extract two pure helpers in `RegistryApis.ts` and use them at all three version-manipulation sites:

- `findCurrentVersion(data)` — lookup by `versionNumber`, never array index.
- `appendVersion(data, item)` — `max(versionNumber) + 1`, appends and advances `currentVersionNumber`.

Call sites: `tagRecordVersionEventId` (fixes positional indexing), `createVersionForDatasetSubmission` (shared arithmetic, no behaviour change), `ConfirmMetadataModal` (fixes `currentVersionNumber + 1`). Dense-history behaviour unchanged; gappy histories become correct instead of silently wrong.

### Documentation corrections

- Issue #3687's divergence list is updated (D3 removed, D1 → resolved by this issue, D2 → intended model difference).
- CLI code comments (`packages/mgd/src/versionAspect.ts` header, `commands/dist.ts` divergence notes) get the same correction — a small `packages/mgd` change folded into this PR.

## Testing

- **Unit:** the web client currently has zero test files (`craco test` exists but nothing runs). Put the two helpers in a small module and add the package's first jest test file: `findCurrentVersion` on dense/gappy data; `appendVersion` `max+1` on gappy data; the D1 bump/no-bump matrix (flagged+existing+tagged → bump; new dist → no bump; flagged+just-replaced → single bump; legacy untagged → tag-only). If the CRA runner proves bitrotted, fall back to manual verification and say so in the PR.
- **Manual UI checklist** against a local minikube deployment (same pattern as PR #3709's live smoke test):
  1. Edit an existing distribution's title → its version bumps to `Distribution metadata updated` and gets an `eventId` tag.
  2. Replace a file → exactly one distribution bump, tagged.
  3. Submit a session with no distribution edits → no distribution bumps; dataset bumps once.
  4. Cross-tool: edit a dataset with the CLI, then the UI, then the CLI — version numbers strictly increase, every version tagged.

## Out of scope

- Skipping the dataset bump for replace-only sessions (D2 stands — the one-bump-per-submission dataset semantic is inherent to the session model).
- Full record-scoped version maintenance in the web client (the CLI model: content PUT → event-gated reconcile per record) — feasible but requires product decisions (dataset `modified` semantics, publishing-aspect handling) and a rework of the untested submission flow; tracked separately in #3714.
- Registry-side (server-side) version maintenance.

## Related

- #3687 / PR #3709 (mgd CLI version maintenance), epic #3648.
- #3714 (full record-scoped alignment in the web client — future investigation, out of scope here).

t83714:
## What & why

Aligns the **web client's** `version`-aspect handling with the mgd CLI (PR #3709 / issue #3687), and hardens the shared arithmetic now that two clients co-write the aspect. Implements **#3713**.

Previously, editing a distribution's metadata in the dataset editor bumped **nothing** on the distribution — only the per-submission dataset version moved. Now a distribution edited during a session gets its own version bump at submission (`Distribution metadata updated`), tagged with the registry event id — matching what `mgd dist update` does on the API side.

## Changes

**New pure module** `magda-web-client/src/api-clients/recordVersionUtils.ts`
- Version types (`VersionItem`/`VersionAspectData`) and `getInitialVersionAspectData` **moved here** from `RegistryApis.ts` (re-exported there, so existing import sites are unchanged).
- `findCurrentVersion` (lookup by `versionNumber`, never array index), `appendVersion` (`max(versionNumber)+1`, non-mutating), and `reconcileDistributionVersionOnSubmit` (the bump/tag/no-op decision, eventId-gated).
- Ships with the package's **first jest test file** (11 tests covering the decision matrix + gappy-history arithmetic).

**Hardening** (correct behaviour on gappy histories a second writer can produce; no change on dense histories)
- `tagRecordVersionEventId` and `createVersionForDatasetSubmission` use the shared helpers instead of positional indexing / `currentVersionNumber+1`.
- `ConfirmMetadataModal`'s file-replacement bump uses `appendVersion` — fixes the `+1` arithmetic and a latent shared-object mutation.

**D1 behaviour** (the metadata-edit bump)
- `editDistribution` (both Add and Edit pages) sets a transient `metadataEditedDuringSession` flag.
- `convertStateToDistributionRecords` clones the aspect, strips the flag (never written to `dcat-distribution-strings`), and reconciles the version; `submitDatasetFromState` clears the flag after a successful submit (no double-bump on re-submit). The submission PUT + existing `tagRecordVersionEventId` tag the new version.

**Docs**
- `docs(mgd)`: corrected inaccurate comments in `packages/mgd`'s version code (the web client *does* tag distributions; #3713 makes it bump them too).
- New Web UI e2e test case + a reusable **JWT-injecting dev-proxy recipe** for running the local UI against a cluster without the auth-internal/HTTPS setup; surfaced the `e2e-test-cases/` folder from the cluster-deployment-test doc.

## Deliberate scope / divergences

- A submission that only replaces a file still bumps the parent dataset (the web client's one-bump-per-submission session model); documented as an intended difference from the CLI.
- Full record-scoped alignment (dist-only changes never touching the dataset version) was assessed and deferred to **#3714** (needs product decisions on `modified`-date and publishing semantics).

## Testing

- `yarn test` (jest) **11/11**; `tsc --noEmit` clean; mgd suite still green (148).
- **Live UI E2E against a minikube deployment**: metadata edit → version bump + eventId tag; re-submit without editing → no spurious bump; multi-version distribution → version-history list renders and each version link navigates (`?version=N`) to the time-travel-reconstructed state. Procedure captured in the new e2e test case.

Closes #3713.


t83714:
Full cluster E2E tested

Issue #3790, opened 2026-09-14
Re-evaluate dataset preview source preference after TerriaJS 8 upgrade


t83714:
## Context

`magda-web-client/src/Components/Common/DataPreviewMap.tsx` chooses the preferred distribution for the dataset-page Map Preview through `DEFAULT_DATA_SOURCE_PREFERENCE` (or deployment override `previewMapFormatPerference`).

The current default order is effectively:

```text
WMS
Esri MapServer
WFS
Esri FeatureServer
GeoJSON
csv-geo-au
KML
KMZ
```

That ordering was established while `magda-preview-map` used an older TerriaJS stack with weaker ArcGIS FeatureServer support.

The preview application is being upgraded to TerriaMap v0.4.8 / TerriaJS 8.13, which has substantially improved native ArcGIS MapServer/FeatureServer handling, including modern FeatureServer layer loading and TerriaJS-controlled request strategies.

Upgrade tracker: https://github.com/magda-io/magda-preview-map/issues/19

Final real-client validation / source-order evidence gathering: https://github.com/magda-io/magda-preview-map/issues/32

This is a **post-upgrade** follow-up. Do not change the production source order until #32 has produced evidence from representative real datasets.

## Goal

Re-evaluate `DEFAULT_DATA_SOURCE_PREFERENCE` after the TerriaJS 8 upgrade and choose the order that gives the most reliable and useful dataset-page preview.

The decision should be based on measured behavior rather than assuming either OGC or ArcGIS services are always preferable.

## Hypotheses to evaluate

Issue #32 will compare at least:

### A — current

```text
WMS -> Esri MapServer -> WFS -> Esri FeatureServer
```

### B — conservative TerriaJS 8 order

```text
WMS -> Esri MapServer -> Esri FeatureServer -> WFS
```

### C — ArcGIS-native-first

```text
Esri MapServer -> Esri FeatureServer -> WMS -> WFS
```

A likely outcome is that FeatureServer should move ahead of WFS, but **this issue should not prescribe that answer before reviewing the evidence**. The final recommendation may be A, B, C, or another refinement.

File formats after the service formats should also be reviewed if #32 reveals a reason to change them, but they are not the primary motivation for this issue.

## Evidence to consume from #32

For representative datasets, review:

- time to first useful render;
- completeness of data shown;
- styling fidelity;
- feature picking / feature-info behavior;
- network request count and approximate transfer volume;
- failure rate and service compatibility;
- large-dataset behavior;
[...]

Issue #3789, opened 2026-09-14
Use TerriaJS service discovery for compact WMS/WFS preview selection


t83714:
## Context

The current dataset preview UI in [`magda-web-client/src/Components/Common/DataPreviewMap.tsx`](https://github.com/magda-io/magda/blob/main/magda-web-client/src/Components/Common/DataPreviewMap.tsx) implements WMS/WFS service discovery itself:

- fetches WMS/WFS GetCapabilities;
- parses the XML in the web client;
- discovers available layers / feature types;
- chooses the initial member;
- renders the compact selector;
- sends `selectedWmsLayerName` / `selectedWfsFeatureTypeName` to `magda-preview-map`.

As part of the TerriaMap v0.4.8 / TerriaJS 8.13 upgrade in `magda-io/magda-preview-map`, the preview application now has native TerriaJS WMS/WFS group/service discovery available and already uses that functionality for preview-side fallback.

Upgrade tracker: https://github.com/magda-io/magda-preview-map/issues/19

Final real-client validation / evidence gathering: https://github.com/magda-io/magda-preview-map/issues/32

This issue is deliberately **post-upgrade**. Do not expand the v0.4.8 compatibility migration to implement it. Refine this issue after #32 records findings from the real Magda web-client test.

## Problem

WMS/WFS protocol knowledge currently exists in both:

1. the Magda web client; and
2. TerriaJS / `magda-preview-map`.

This duplicates capabilities parsing and service-member discovery, and means `DataPreviewMap.tsx` needs to understand WMS/WFS protocol details that TerriaJS already models natively.

At the same time, the current compact selector above the dataset-page preview is useful. We should **not** solve the duplication by exposing Terria's full explorer/workbench/catalog UI inside the small preview iframe.

## Direction to investigate

Prefer an architecture where:

```text
Magda web client
    |
    | distribution / preview request
    v
magda-preview-map / TerriaJS
    |
    +-- native GetCapabilities loading
    +-- native service-member discovery
    |
    | available members via constrained iframe protocol
    v
Magda compact dropdown
    |
    | selected member
    v
magda-preview-map / TerriaJS
```

The exact protocol is intentionally not prescribed yet. One possible design is a small `postMessage` extension where the preview reports available WMS/WFS members and the parent sends the selected member back, but #30/#32 findings on lifecycle/origin handling should inform the final design.

## Responsibilities to preserve

The likely ownership split should remain:

### Magda web client

- owns the dataset-page UX;
[...]

PR #3791, opened 2026-09-17
chore(magda-core): bump preview-map to 2.0.0-alpha.0

t83714:
Bumps the `magda-preview-map` dependency in `magda-core` from `1.1.3` to the TerriaJS 8 / Node 24 prerelease **`2.0.0-alpha.0`**, so Magda can integrate and validate the upgraded preview map.

## Changes
- `deploy/helm/magda-core/Chart.yaml` — `magda-preview-map` (aliased `preview-map`) `1.1.3` → `2.0.0-alpha.0`
- `deploy/helm/magda-core/Chart.lock` — regenerated (`helm dep up`)
- `deploy/helm/magda-core/README.md` — regenerated dependency table (helm-docs)

Scope is intentionally limited to the preview-map dependency. The `magda` umbrella lock is unchanged because `magda-core` is a `file://` dependency pinned at `6.2.1` and is repackaged (picking up the new preview-map) at build/release time. Unrelated pre-existing README drift in `local-deployment` (stale auth-chart versions) was left out of this PR.

## Verification
```
$ helm template t deploy/helm/magda-core | grep magda-preview-map:
image: "ghcr.io/magda-io/magda-preview-map:2.0.0-alpha.0"

$ helm lint deploy/helm/magda-core
1 chart(s) linted, 0 chart(s) failed
```

The `2.0.0-alpha.0` artifacts (multi-arch image — Node 24, `terriajs-server` 5.0.0 — and OCI Helm chart) are published and independently verified: magda-io/magda-preview-map#42.

The upgrade changes the preview-map runtime substantially (TerriaJS 6 → 8, React 18, Node 24, `terriajs-server` 5, new iframe security and basemap behaviour) while preserving the existing Magda iframe payload contract. Tracks magda-io/magda-preview-map#43.

## Out of scope
- No stable pin (this is a prerelease integration).
- Magda #3789 / #3790 (WMS/WFS discovery, source-preference follow-ups).


t83714:
⚠️ Blocker: `magda-preview-map@2.0.0-alpha.0` cannot be consumed behind the Magda gateway as-is.

When served under the gateway's `/preview-map/` path prefix, the v2 client fails to boot: the TerriaMap v0.4.8 build emits `<base href="/">`, so asset requests resolve to origin-root (`/build/TerriaMap.js` → 303 → `/error?404`) instead of `/preview-map/build/...`. v1.1.3 shipped no `<base>` tag, so its relative assets worked under the prefix. Verified end-to-end against a dev Magda deployment (client `booted: false`).

Fix is in magda-preview-map (build should not emit an absolute `<base href="/">`); tracked for a `2.0.0-alpha.1`. **This PR's dependency should be repinned from `2.0.0-alpha.0` to `2.0.0-alpha.1` once that is published.** Holding until then.

t83714:
✅ **Verified end-to-end through the real Magda gateway** (dev deployment, no CSP bypass).

Two fixes were required for the TerriaJS 8 preview-map to work behind the gateway, both now in this PR / its dependency:
1. **preview-map `2.0.0-alpha.1`** — removes the absolute `<base href="/">` that broke asset loading under the `/preview-map/` prefix (magda-io/magda-preview-map#49). This PR is repinned to alpha.1.
2. **magda-core gateway CSP** — `/preview-map/*` `script-src` now allows `'unsafe-eval' 'unsafe-inline' blob:` (TerriaJS 8 / Cesium requirement).

Tested via PR build `v7.0.0-pr.3791.2` deployed to a dev Magda: the client boots through the gateway and the iframe lifecycle completes (`ready → loading complete`). WMS (GA National Dynamic Land Cover) and Esri FeatureServer render; OSM fallback for legacy `Positron (Light)` works; a bogus distribution reaches the terminal error state with the spinner cleared. The only failing previews are upstream data/source issues (ALA `/reflect` is not a conformant OGC WMS; data.gov.au CSV returns 403 = known non-blocker #40), not preview-map regressions — identical to standalone results.

Ready to merge to `next`; a `v7.0.0-alpha.3` release should follow.

